PR #24, opened 2022-03-24
Footprint reduction 1: shrink struct pci_smbus_controller by reducing the impact of redundancy

debrouxl:
The pointers from pci_smbus_controller are highly redundant, so push them aside to a single struct.
I tested this on real hardware before the latest rebase.

### DIFF
--- a/system/smbus.c
+++ b/system/smbus.c
@@ -41,70 +41,125 @@ static void ich5_get_smb(void);
 static uint8_t ich5_process(void);
 static uint8_t ich5_read_spd_byte(uint8_t adr, uint16_t cmd);
 
-static const struct pci_smbus_controller smbcontrollers[] = {
-    {0x8086, 0x2413, "82801AA (ICH)",           ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x2423, "82801AB (ICH)",           ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x2443, "82801BA (ICH2)",          ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x2483, "82801CA (ICH3)",          ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x24C3, "82801DB (ICH4)",          ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x24D3, "82801E (ICH5)",           ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x25A4, "6300ESB",                 ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x266A, "82801F (ICH6)",           ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x269B, "6310ESB/6320ESB",         ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x27DA, "82801G (ICH7)",           ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x283E, "82801H (ICH8)",           ich5_get_smb, ich5_read_spd_byte},
-    {0x8086, 0x2930, "82801I (ICH9)",           ich5_get_smb, ich5_read_spd_byte},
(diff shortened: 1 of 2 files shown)


x86fr:
Missed the pull request before rebase/merge to main :-/

Do you prefer to merge it immediately or wait until the whole smbus.c code is finished to apply an overall code optimization?
I can't be sure at this time that the whole structure will not need a rework (as I already did with HAS_xxxx now replaced with another structure)

debrouxl:
Up to you :)
You can perform the rework after applying this optimization (well, provided it builds and works, but I tested that before the rebase), so that I don't have to rebase it later, or you can wait, maybe rework the structure, and I'll rebase it later. The former should create less work overall than the latter, but it's a pretty minor difference.

x86fr:
I don't think the sdr/ddr/ddr2fb parsing code will change anything on these structs, but the new code needed for smbus/i2c read/write on various other platforms like AMD CPUs might. :-/

debrouxl:
Indeed, the incomplete code for older RAM types living on the spd_decoding branch doesn't need to touch struct pci_smbus_controller, and the completed code certainly won't either.

Now that you made caps disappear, this PR leaves 2 bytes of padding on both x86 & x86_64: we could use those if the code performing SMBus/I2C R/W on other pieces of hardware happens to require extra information about hardware quirks or whatever. 3 bytes if we shrink the ops index to uint8_t, it's not like we're going to have a number of function pointer pairs even remotely close to 256.

x86fr:
I wonder if the whole struct is `pci_smbus_controller` struct is useful after all. 

- There is only one function for all `0x8086` vendor (Intel) and checking only if device is supported is enough
- The PCH name is not used actually (but might be in the future)
- I can easily get rid of the `ich5_get_smb` function by adding a initialized flag in the `read_spd` function. 

In the other hand, the struct is easy to understand (a bit less with the `SMBC_OPS_IDX_ICH5` but still easy) and easy to maintain...

Well, it's probably better to add AMD functions before trying to optimize (or delete) this struct

debrouxl:
As is, the alignment of the string pointer is 8 on x86_64, so there's no rush to remove the redundancy on 0x8086 values - there's going to be some padding anyway.

I fetched the code I used a couple decades ago on the TI-68k series to get rid of string pointers and the corresponding relocations, it was obviously nothing extraordinary:
```
asm("
.data
.xdef explorer_msgs
explorer_msgs:

      .word explorer_msgs_2  - explorer_msgs
      .word explorer_msgs_3  - explorer_msgs
      .word explorer_msgs_4  - explorer_msgs
...

    explorer_msgs_2:
      .asciz \"Address-Error\"

    explorer_msgs_3:
      .asciz \"Illegal-Instruction\"

    explorer_msgs_4:
      .asciz \"Divide-By-Zero\"
...
```
This could be used on other platforms and expanded beyond string arrays, but... under that form, the readability and maintainability of the corresponding code block will, uh, take a severe hit :)

The maintainability issue should be alleviated by building a macro containing assembly blocks using .section / .previous pairs, plus linker script magic to tie the pieces together, which would hopefully be used that way:
```
PCI_SMBUS_CONTROLLER_STRUCT(0x8086, 0x2413,    SMBC_OPS_IDX_ICH5, "82801AA (ICH)"),
```
That's what the Linux kernel, among others, does:
```
#define __get_user_asm(x, addr, err, itype, ltype)                      \
        asm volatile("\n"                                               \
                     "1:        mov"itype" %[umem],%[output]\n"         \
                     "2:\n"                                             \
                     ".section .fixup,\"ax\"\n"                         \
                     "3:        mov %[efault],%[errout]\n"              \
                     "  xorl %k[output],%k[output]\n"                   \
                     "  jmp 2b\n"                                       \
                     ".previous\n"                                      \
                     _ASM_EXTABLE_UA(1b, 3b)                            \
                     : [errout] "=r" (err),                             \
                       [output] ltype(x)                                \
                     : [umem] "m" (__m(addr)),                          \
                       [efault] "i" (-EFAULT), "0" (err))
```
so it ought to be possible.

x86fr:
Whole struct removed (PR #44)

PR #28, opened 2022-03-30
Add support for disabling the CPU cache

debrouxl:
It makes the test horrendously slow on recent computers which have lots of RAM.

The original version of this code was made before the cache / memory benchmark was restored, and this seems to run with cache enabled, judging by the displayed speeds; shouldn't I attempt to modify the benchmark, in `measure_memory_bandwidth` and possibly `display_init` ?

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -93,6 +93,7 @@ cpu_state_t     cpu_state[MAX_CPUS];
 
 bool            enable_temperature = false;
 bool            enable_trace       = false;
+bool            enable_cpucache    = true;
 
 bool            enable_sm          = true;
 bool            enable_bench       = true;
@@ -185,6 +186,8 @@ static void parse_option(const char *option, const char *params)
         }
     } else if (strncmp(option, "nobench", 8) == 0) {
         enable_bench = false;
+    } else if (strncmp(option, "nocpucache", 11) == 0) {
+        enable_cpucache = false;
     } else if (strncmp(option, "noehci", 7) == 0) {
         usb_init_options |= USB_IGNORE_EHCI;
     } else if (strncmp(option, "nopause", 8) == 0) {
--- a/app/config.c
+++ b/app/config.c
@@ -91,7 +91,7 @@ bool            smp_enabled        = true;
 
 bool            enable_temperature = true;
 bool            enable_trace       = false;
-bool            enable_cpucache    = true;
+bool            enable_cpu_cache   = true;
 
 bool            enable_sm          = true;
 bool            enable_bench       = true;
@@ -185,7 +185,7 @@ static void parse_option(const char *option, const char *params)
     } else if (strncmp(option, "nobench", 8) == 0) {
         enable_bench = false;
     } else if (strncmp(option, "nocpucache", 11) == 0) {
-        enable_cpucache = false;
+        enable_cpu_cache = false;
     } else if (strncmp(option, "noehci", 7) == 0) {
         usb_init_options |= USB_IGNORE_EHCI;
     } else if (strncmp(option, "nopause", 8) == 0) {
--- a/app/main.c
+++ b/app/main.c
@@ -469,8 +469,7 @@ void main(void)
     if (init_state < 2) {
         if (enable_cpu_cache) {
             cache_on();
-        }
-        else {
+        } else {
             cache_off();
         }
         if (my_cpu == 0) {
--- a/app/main.c
+++ b/app/main.c
@@ -467,7 +467,7 @@ void main(void)
         my_cpu = smp_my_cpu_num();
     }
     if (init_state < 2) {
-        if (enable_cpucache) {
+        if (enable_cpu_cache) {
             cache_on();
         }
         else {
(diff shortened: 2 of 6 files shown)


x86fr:
What's the goal to disable CPU caches?  

Back in the early ' 00 for the original Memtest86 and Memtest86+, that was a remnant of the era where L2 cache was in the motherboard and might fail (especially 486s). Since 25 years, all CPUs have embedded L2 cache. It's still used on test 0 to avoid modern IMC to optimize out those memory accesses, but I don't get the point about being able to disable it for anything other than avoiding that.

debrouxl:
I do have at least one MB for Pentium-class processors which has external L2 cache (and there have been newer Pentium MMX models which support external L2 cache), and memtest86+ is still supposed to support Pentium-class computers :)
Also, uncached mode probably triggers memory access patterns different from the standard mode (though it might not matter for finding bit flips), this feature was easy to develop and its maintenance impact going forward should be low because it's small.

However, it's a fact that in this day and age, this has no practical usefulness for a crushing majority of memtest86+ users, and I knew it when submitting this PR.

x86fr:
That deserve a test on a real Socket 7 motherboard with external L2 cache :)

IIRC, on a 486DX4-100 with 4MB of RAM and cache disabled, a single Memtest86+ pass took more than 24 hours. As this option is only needed for Socket 5/7 motherboard, it should not led to an unusable system. Such old platforms still in use today usually have RAM maxed out. Can you try on your old MB? Also nice to see if v6 is still able to run on prehistoric hardware :)

debrouxl:
Eh, that old MB isn't permanently housed into a computer case with a PSU :)
I'd have to fetch and connect the pieces together; also, maybe read the documentation about launching memtest86+ without GRUB, which is the only way I've ever used memtest86 / memtest86+, on higher-powered computers ^^

app/config.c
```diff
@@ -92,6 +92,7 @@ cpu_state_t     cpu_state[MAX_CPUS];
 
 bool            enable_temperature = false;
 bool            enable_trace       = false;
+bool            enable_cpucache    = true;
```
martinwhitaker, line 95: `enable_cpu_cache` would be a little easier to read.

app/main.c
```diff
@@ -466,7 +466,12 @@ void main(void)
         my_cpu = smp_my_cpu_num();
     }
     if (init_state < 2) {
-        cache_on();
+        if (enable_cpucache) {
+            cache_on();
+        }
+        else {
```
martinwhitaker, line 472: whitespace

martinwhitaker:
Do you want to merge this before the mass testing next week? I'm not sure how useful it is (I would have thought that if the cache wasn't working, the machine would fail to boot), but have no objections (apart from my two comments). It will need to be added to the README.

debrouxl:
I was in the process of rebasing my branch onto current main. On which line should "UC" be displayed, now that the aesthetic branch is merged ?

x86fr:
Ah! I'm also on this one to merge. Changing the UC display

debrouxl:
When you're done, I'll handle the rebase to get rid of the merge from main, and linearize the commit history, alright ? :)

x86fr:
Done. 

PS: As a small change, we can probably just Squash & merge. That will remove all unwanted noise in history.

debrouxl:
Yup, squashing is one of the ways to clean up, indeed.
What about modifying the benchmark (`measure_memory_bandwidth`, `display_init`, etc.), as I mentioned in my first comment ?

x86fr:
Well, after testing it on some motherboard, this feature will require much more code to be efficiently implemented (= to work as expected).

On modern CPU, just flipping the CD bit will disable L1->L3 cache. You can sometimes disable only L3 by messing with MTRRs and architecture-specific MSRs, but in all case, the internal cache is tested by the hardware at BIST since the 486 (I test that on my Universal Chip Analyzer as the BIST result is accessible via JTAG), and the CPU will not boot with a defective L1/L2/L3 internal cache.

So the only use (except to avoid the memory controller optimizing out the access), is to enable/disable the external L2 cache.

But the issue is that you will NOT disable an external L2 cache just by flipping the CD bit. On a 486/Pentium/K6, the CD bit will disable the internal L1 cache but the L2 cache will still work as before because it is handled by the memory controller, being completely transparent for the CPU. It's possible to disable the L2 cache on these old platforms, but you have to use a northbridge-specific register (usually a I/O access to some obscure port) that we don't have here.

x86fr:
It's probably time to close this PR. It could be reopened if anyone find a single real-world case where disabling cache this way is useful, but I doubt that will happen. Starting with the 486DX2s, the standard CPU BIST fully test all internal caches at POST and the external L2 (for 486 and P5s) can't be disabled without proprietary tricks.

debrouxl:
Alright, for the reasons you detailed.

For my knowledge wrt. write caching: can the usage of Page Attribute Table WC mode - when available and unencumbered by known processor errata on early x86 PAT-capable CPUs, and also coupled with memory fences when needed - prevent a memory tester from detecting some memory errors ?

PR #32, opened 2022-04-02
Add support for Serial/TTY

x86fr:


app/config.c
```diff
@@ -129,6 +158,14 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "high", 5) == 0) {
             power_save = POWER_SAVE_HIGH;
         }
+    } else if (strncmp(option, "console", 8) == 0) {
+        if (params != NULL) {
+            parse_serial_params(params);
+        }
+    } else if (strncmp(option, "nobench", 8) == 0) {
```
debrouxl, line 165: Do we want to keep alphabetical ordering ? In which case the previous location for the nopause and nobench flag ordering was better.

app/config.c
```diff
@@ -321,9 +360,17 @@ static void test_selection_menu(void)
         display_enabled(POP_R+12, i, test_list[i].enabled);
     }
 
+    bool tty_update = true;
     bool exit_menu = false;
     while (!exit_menu) {
         bool changed = false;
+
+        if (enable_tty && tty_update) {
```
debrouxl, line 368: This block is repeated multiple times across this file, so maybe there should be a macro for that ?

app/display.c
```diff
@@ -175,6 +184,10 @@ void display_start_run(void)
         next_spin_time = run_start_time + SPINNER_PERIOD * clks_per_msec;
     }
     display_spinner('-');
+
+    if(enable_tty){
```
debrouxl, line 188: Whitespace :)

app/error.c
```diff
@@ -365,5 +367,9 @@ void error_update(void)
                                    test_list[test_num].errors);
         }
         display_error_count(error_count);
+
+        if(enable_tty) {
```
debrouxl, line 371: Whitespace :)

lib/string.c
```diff
@@ -105,3 +120,29 @@ char *strstr(const char *haystack, const char *needle)
...
+        num /= 10;
+    }
+
+    // Last is string terminator
+    str[i] = '\0';
+
+    reverse(str);
+
+    return str;
+}
```
debrouxl, line 148: Newline at EOF.

system/keyboard.c
```diff
@@ -255,5 +258,13 @@ char get_key(void)
         }
     }
 
+    if (enable_tty) {
+        uint8_t c = tty_get_key();
+        if (c != 0xFF) {
+            if(c == 0x0D) c = '\n'; // Enter
```
debrouxl, line 264: Whitespace.

system/serial.c
```diff
@@ -0,0 +1,210 @@
...
+	if (!port->enable) {
+		return;
+	}
+	/* Now, do each character */
+	while (*p) {
+		serial_wait_for_xmit(port);
+
+		/* Send the character out. */
+		serial_write_reg(port, UART_TX, *p);
+		if(*p==10) {
```
debrouxl, line 76: Whitespace.

system/serial.h
```diff
@@ -0,0 +1,181 @@
...
+#define SERIAL_PORT_0x2F8   1
+#define SERIAL_PORT_0x3E8   2
+#define SERIAL_PORT_0x2E8   3
+
+static const uint16_t serial_base_ports[] = { 0x3f8, 0x2f8, 0x3e8, 0x2e8 };
+
+struct serial_port {
+	bool enable;
+	bool is_mmio;
+	unsigned int parity;
```
debrouxl, line 28: Could use narrower data types for these, but it wouldn't have much of an impact on the code size.

system/serial.c
```diff
@@ -0,0 +1,210 @@
...
+	if (!enable_tty) {
+	    return;
+	}
+
+	int uart_status, serial_div;
+	unsigned char lcr;
+
+    console_serial.enable       = true;
+    console_serial.is_mmio      = false;
+    console_serial.baudrate     = tty_params_baud;
```
debrouxl, line 114: The order of declaration and order of initialization do not match.

system/serial.c
```diff
@@ -0,0 +1,210 @@
...
+            p[28] = '\0';
+            tty_print(row,0,p);
+            tty_normal();
+            p[28] = '|';
+            tty_print(row, 28, p + 28);
+            continue;
+        }
+
+        // Replace degree symbol with '*' for tty to avoid a C437/VT100 translation table.
+        if(row == 7 && col_len > 77 && (shadow_buffer[7][73].value & 0xFF) == 0xF8) {
```
debrouxl, line 187: Whitespace.

system/serial.c
```diff
@@ -0,0 +1,210 @@
...
+            tty_normal();
+        }
+    }
+}
+
+char tty_get_key(void)
+{
+    int uart_status = serial_read_reg(&console_serial, UART_LSR);
+
+	if (uart_status & UART_LSR_DR) {
```
debrouxl, line 205: Whitespace.

system/serial.h
```diff
@@ -0,0 +1,181 @@
...
+#define TTY_UNDERLINE       "\x1b[4m"
+#define TTY_INVERSE         "\x1b[7m"
+
+/*
+ * Definitions for the Base UART Registers
+ */
+
+#define UART_REF_CLK 1843200
+
+#define UART_RX		0	/* In:  Receive buffer (DLAB=0) */
```
debrouxl, line 55: Whitespace.

system/serial.h
```diff
@@ -0,0 +1,181 @@
...
+#define UART_LCR_WLEN6  0x01	/* Wordlength: 6 bits */
+#define UART_LCR_WLEN7  0x02	/* Wordlength: 7 bits */
+#define UART_LCR_WLEN8  0x03	/* Wordlength: 8 bits */
+
+/*
+ * Definitions for the Line Status Register
+ */
+#define UART_LSR_TEMT	0x40	/* Transmitter empty */
+#define UART_LSR_THRE	0x20	/* Transmit-hold-register empty */
+#define UART_LSR_BI     0x10	/* Break interrupt indicator */
```
debrouxl, line 90: Whitespace here and in the following code blocks.

x86fr: PS: Most of these blank lines are intentional. They basically increase readability by separating code blocks (or isolate important line visually). Removing all blank lines add a lot of obfuscation to the code.

debrouxl: The blank lines are fine indeed, but there's inconsistent indentation of the right side of #defines :)

system/keyboard.c
x86fr: I'm doing a pass to solve all tabs->spaces, but I can't see the issue here.

debrouxl: Most other places use `if (` :)

debrouxl:
Hmm... I seem to have failed to send a comment about rebasing the branch first, instead of using the Rebase and merge Github option, in order to move conflict resolution from a merge commit to the commits where the changes belong.

In my experience at work, conflict resolution as part of a merge commit proves to be a pain when history rewrites (usually to be avoided) are needed for some reason: they break the neat automatic history linearization `git rebase --interactive` can usually perform with these conflict-resoluting merge commits.

system/keyboard.c
x86fr: ah yes, I was looking for superfluous spaces :) 

I always had trouble switching between coding convention. Ie: many devs I work with consider anything other than a 2-spaces indent as heretic (but they also loves to write obscure one liner...) .

x86fr:
For this specific PR, I plan to squash it as it's a single feature.

debrouxl:
Oh yeah, squashing is a pretty fine - and simpler, besides - way to handle this PR :)

app/config.c
martinwhitaker: Alternatively, change all occurrences of `tty_update = true` to `tty_update = enable_tty`. then you can simply use `if (tty_update)`.

app/config.c
```diff
@@ -321,9 +360,17 @@ static void test_selection_menu(void)
...
+            tty_popup_redraw();
+            tty_update = false;
```
martinwhitaker, line 370: This is redundant, because you immediately set tty_update back to true (same in other places).

app/display.c
```diff
@@ -69,14 +73,16 @@ void display_init(void)
 
     clear_screen();
 
-    set_foreground_colour(RED);
-    set_background_colour(WHITE);
+    set_foreground_colour(BLACK);
+    set_background_colour(GREEN);
```
martinwhitaker, line 77: This is unrelated to serial/TTY support, so belongs in a separate commit.

app/config.c
```diff
@@ -44,27 +45,6 @@
...
-#define POP_REGION  POP_R, POP_C, POP_LAST_R, POP_LAST_C
-
-#define POP_LM      (POP_C + 3)     // Left margin
-#define POP_LI      (POP_C + 5)     // List indent
-
-#define SEL_W       32
-#define SEL_H       2
-
-#define SEL_AREA    (SEL_W * SEL_H)
-
```
martinwhitaker, line 67: You could avoid making all these constants public by using `tty_send_region(POP_REGION)` instead of `tty_popup_redraw`. I think that would be just as readable.

system/screen.h
```diff
@@ -19,6 +19,26 @@
 #define SCREEN_WIDTH    80
 #define SCREEN_HEIGHT   25
 
+
+typedef struct {
+    uint8_t     r;
+    uint8_t     g;
+    uint8_t     b;
+} __attribute__((packed)) rgb_value_t;
```
martinwhitaker, line 27: I don't think you need to make this type public.

app/config.c
x86fr: yes! the redraw block used to be at the end of the function :-/

app/display.c
x86fr: Correct. That belong to the aesthetic branch (and also the red + some line below). Anyway, I'll try to push that quickly to main because the actual color confuses some betatesters (I still get bug reports for passmark's memtest86 :-/). Removing here.

system/keyboard.c
martinwhitaker: I'm quite used to switching. For my own code I use Camel case and an indentation of 2. I also help maintain Icarus Verilog - the only code base I know of that uses an indentation of 6.

app/display.c
martinwhitaker: I find black on green hard to read which is why I changed the colour scheme.

app/config.c
x86fr: I moved them for the popup redraw in read.c, but I will revert that abd only redraw the row/col region passed to `read_value()` instead in that loop

app/config.c
x86fr: Yes!

app/display.c
x86fr: I also have the same concern about the RED/WHITE color scheme :-/ 

Anyway, I don't have any fetish for the green color. It was useful for a very long time because the original memtest86 was red and memtest86+ was green, so anyone could tell at a glance (even from a distance) if a system was running memtest86 or memtest86+. 

But now, passmark changed the title for their memtest86 from red to green so that didn't work anymore.

I'll do some tries next week and ask for feedbacks

Issue #25, opened 2022-03-25
Dealing with Secure Boot


x86fr:
Secure Boot-enabled PCs (now mandatory since Windows 11) will be a major issue for Memtest86+. 

As a GPL software, we can't have our binary signed by Microsoft for a fee like passmark's closed-source memtest. 
(Well, every GPLv3 software are clearly blacklisted, but it seems there is a (tiny) possibility to have a GPLv2 software signed by Microsoft.)

shim is another option but the process looks harsh. 

We will probably need the support of a major distro (Debian?) to be able to chain-load Memtest86+ from a signed GRUB. 

I'm also thinking about some hack to bypass Secure Boot in last resort... Doable but very dirty.

Any idea?

gordboy:
What about just making the header big enough for the users to sign their own binaries with their own MOK, for now. And look at other options later.

I was horrified when I went thru all the palaver of creating and enrolling keys only to find that the efi binary was unsuitable for signing ... :)

x86fr:
@vathpela is working on this on PR #34

corvus1:
Secure Boot is not mandatory with win11. UEFI booting is, if you want to upgrade from win10. For a fresh install, nothing is mandatory, you can install on a core2duo if you like pain and suffering. :-) 

But point is, the actual requirement is the PC being booted in UEFI mode at the time you're upgrading. After you've done that, you can boot in CSM for memtesting purposes if you want to.

AADJND5:
Sorry if it's off topic. Why in BIOS the program is loaded on the whole screen, but not on the whole in UEFI, black stripes on top and bottom, like a wide screen.

AADJND5:
Can this be fixed ?

corvus1:
I have a very old laptop that does that in text mode. There is a setting in BIOS to stretch the image to the full size of the screen, but I wouldn't say it makes it look any better. 

Why, though? Does it interfere with the function?

AADJND5:
No, it doesn't interfere with work. It just looks more comfortable. Can I add automatic loading of the screen resolution ? It's just a wish. Thank you for your work and your program.

martinwhitaker:
> Sorry if it's off topic. 
Please don't hijack issues. Start a new thread in the discussions section.

corvus1:
@martinwhitaker  Sorry. :-)

x86fr:
Secure Boot is mandatory for any SI (system integrator) that install Windows 11 on their PCs. 
[...]

PR #34, opened 2022-04-05
Add support for UEFI signing, SBAT signature revocations, and NX mitigations.

vathpela:
(This is not yet ready)

In order to facilitate adoption by major linux distributions, memtest86+ EFI builds are going to need to support Authenticode signatures and the SBAT revocation method, and will eventually need to support being loaded with NX (W^X) vulnerability mitigations enabled.

This patch series aims to add all of those, though right now it doesn't work correctly on some UEFI firmware builds, and I'm still debugging, so there are some extra patches here that won't be necessary once I'm ready to propose it upstream.

### DIFF
--- a/boot/header.S
+++ b/boot/header.S
@@ -163,7 +163,20 @@ extra_header_fields:
 	.long	0				# SizeOfHeapCommit
 #endif
 	.long	0				# LoaderFlags
-	.long	0				# NumberOfRvaAndSizes
+	.long	IMAGE_DIRECTORY_ENTRY_DEBUG	# NumberOfRvaAndSizes
+
+	.long	0				# DataDirectory.Export.VirtualAddress
+	.long	0				# DataDirectory.Export.Size
+	.long	0				# DataDirectory.Import.VirtualAddress
+	.long	0				# DataDirectory.Import.Size
+	.long	0				# DataDirectory.Resource.VirtualAddress
+	.long	0				# DataDirectory.Resource.Size
+	.long	0				# DataDirectory.Exception.VirtualAddress
+	.long	0				# DataDirectory.Exception.Size
+	.long	0				# DataDirectory.Certs.VirtualAddress
+	.long	0				# DataDirectory.Certs.Size
+	.long	0				# DataDirectory.BaseReloc.VirtualAddress
+	.long	0				# DataDirectory.BaseReloc.Size
 
 	# Section table
 section_table:
--- a/boot/header.S
+++ b/boot/header.S
@@ -90,7 +90,7 @@ coff_header:
 #else
 	.word	IMAGE_FILE_MACHINE_I386		# Machine (i386)
 #endif
-	.word	2				# NumberOfSections
+	.word	3				# NumberOfSections
 	.long	0 				# TimeDateStamp
 	.long	0				# PointerToSymbolTable
 	.long	0				# NumberOfSymbols
@@ -118,14 +118,14 @@ optional_header:
 	.byte	0x14				# MinorLinkerVersion
 
 	.long	_text_size			# SizeOfCode
-	.long	0				# SizeOfInitializedData
+	.long	_sbat_size			# SizeOfInitializedData
 	.long	0				# SizeOfUninitializedData
 
 	.long	BASE_OF_CODE + 0x1e0		# AddressOfEntryPoint
 
 	.long	BASE_OF_CODE			# BaseOfCode
 #ifndef __x86_64__
-	.long	0				# BaseOfData
+	.long	_sbat_start			# BaseOfData
 #endif
 
 extra_header_fields:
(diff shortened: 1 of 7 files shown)


debrouxl:
Whoa, thanks for this draft contribution :)

It shall help with #25 when merged.

x86fr:
Thanks Peter, this is greatly appreciated 😍

martinwhitaker (commented):
Adding boot/peimage.h will also require updating the top level LICENSE file. Currently all files in the project are GPL 2 only.

martinwhitaker:
We don't need to load at 0x200000, but we do need to load at an address < 3GB (preferably < 2GB). I don't know any way to achieve that other than specifying a fixed address. To keep the program memory footprint low, we set up the paging tables to only use the first 4GB of virtual memory space. We use the space between 2GB and 3GB when testing memory, and we use the space between 3GB and 3.5GB to provide permanent access to memory mapped hardware and the frame buffer. So if we allow the EFI loader to load the program anywhere in physical memory, we would have to be prepared to to relocate it.

vathpela:
> Adding boot/peimage.h will also require updating the top level LICENSE file. Currently all files in the project are GPL 2 only.

I'm pretty sure I can find a GPLv2 version of it, let me look around a bit.

vathpela:
> We don't need to load at 0x200000, but we do need to load at an address < 3GB (preferably < 2GB). I don't know any way to achieve that other than specifying a fixed address. To keep the program memory footprint low, we set up the paging tables to only use the first 4GB of virtual memory space. We use the space between 2GB and 3GB when testing memory, and we use the space between 3GB and 3.5GB to provide permanent access to memory mapped hardware and the frame buffer. So if we allow the EFI loader to load the program anywhere in physical memory, we would have to be prepared to to relocate it.

Is relocating it an option you'd be okay with? (I am able to contribute relocation code...)

martinwhitaker:
Relocating is okay, and is already supported - the program regularly relocates itself whilst the tests are running to allow the memory it occupies to be tested. So it should just be a matter of scanning the E820 memory map to find a suitable location, copying the program code and data sections, and contriving to return from `efi_setup()` to the relocated address. The relocation fixups will be done by the startup code.

I just wasn't going to bother doing that unless it was found to be needed.

x86fr:
@vathpela Is there something we can do to help?

01e3:
Hi there.

Just checking what is the plan here?

The PR is currently marked as draft, how much work is left here?

Do we expect this can be merged early during 6.10 development?

vathpela:
So, I haven't had the time to debug whatever issue I'm hitting with this on QEMU, but it works on all of my real hardware, so I think that's probably not a good reason to hold this back, especially since there's little point in actually running memtest in QEMU.

x86fr:
Awesome! We will carefully review this PR then I'll put this in betatesters' hands as soon as possible.

martinwhitaker:
This looks OK to me. I did use QEMU extensively for developing and debugging the USB drivers, but I'm not seeing any issues with it.

x86fr:
Betatesters found an issue with this PR on Asus 700-series motherboard. We're investigating the issue with their BIOS team. It seems linked to a specific ME.

martinwhitaker:
I do note that now _bss_size is not included in SizeOfImage, I can't see that there is any guarantee that the bss section does not collide with memory reserved by the EFI BIOS (which is why I included it).

x86fr:
Info point: We found an issue (indirectly related) on ASUS motherboard BIOS based on 700-series Intel chipset that prevent testing of secureboot related code right now. The bug is serious as it lead to a firmware corruption with boot failure. It's only present on 0602 BIOS update. A beta fix has been sent by ASUS and successfully tested. Betatesters are now waiting for the final new BIOS to resume validation on SB.

vathpela:
> Info point: We found an issue (indirectly related) on ASUS motherboard BIOS based on 700-series Intel chipset that prevent testing of secureboot related code right now. The bug is serious as it lead to a firmware corruption with boot failure. It's only present on 0602 BIOS update. A beta fix has been sent by ASUS and successfully tested. Betatesters are now waiting for the final new BIOS to resume validation on SB.

Am I right in understanding that we believe this is a bios bug?

x86fr:
Yes, this has been confirmed and solved by Asus. Unfortunately, the betatester team found in the meanwhile another (unrelated) [critical bug](https://twitter.com/d0cTB/status/1597948487928516609) on Asus motherboards that broke the validation process on the production line. They're now waiting for a fix to resume testing this PR.

x86fr:
Got the the first preliminary report from betatesters. This PR has been tested on 50+ platforms with Secure Boot Enabled with great success! Thanks @vathpela! The only fails they got were on Intel NUCs (waiting for more details). I tested it myself on some motherboards and laptops without issues so far. It's time to merge!

tenzap:
@vathpela , could you please have a look at https://github.com/memtest86plus/memtest86plus/issues/516 ?
There is I think a typo in your sbat, but even when I fix this, I have another "relocation" issue https://github.com/memtest86plus/memtest86plus/issues/516#issuecomment-2929751736 (Case2) that I don't know how to address.

PR #41, opened 2022-04-14
Various aesthetic changes, more information displayed (and code for detection)

x86fr:
It's time to rebase the branch before the plugfest next week.
I tried to create the better mix possible based on all feedbacks. The biggest changes are on line 8/9. 
Others changes are on spacing, alignment, title color, temperature (max & current), update timer and build number addition.
Code has been added for timing detection (with support for XMP3.0/2.0/1.0 and EPP)

Line 8 now displays CPU Topology (Socket/Core/Thread) and Threading Status (SMP).
Line 9 now displays the most accurate specs as possible :
  1. The actual RAM frequency / Timings from the IMC (will be added later)
  2. If not available, the max RAM frequency/timings possible from SPD (implemented for DDR1->5)
  3. If not available, the Core actually being used for testing (as last resort fallback)

There is some TODO remaining but nothing critical:
- Hybrid CPU (ADL+) topology must be detected with e-core/p-core distinction, but that require a full SMP enumeration
- On DDR1, non integer timings (especially CAS) are not supported (will be rounded to nearest int)
- Last minute: paging mode (X64/PAE) is erased after being displayed and will be re-added soon.

The only information missing in some case is the actual CPU used. It's now only displayed by default if RAM frequency/timing is not detected. It's probably not a big deal, but it can be re-added at the end of line 6 if really needed.

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -21,6 +21,7 @@
 
 #include "config.h"
 #include "error.h"
+#include "githash.h"
 
 #include "tests.h"
 
@@ -76,13 +77,9 @@ void display_init(void)
     set_foreground_colour(BLACK);
     set_background_colour(WHITE);
     clear_screen_region(0, 0, 0, 27);
-#if TESTWORD_WIDTH > 32
-    prints(0, 0, "  Memtest86+ v6.00pre (64b)");
-#else
-    prints(0, 0, "  Memtest86+ v6.00pre (32b)");
-#endif
+    prints(0, 0, "     Memtest86+ v6.00b1");
     set_foreground_colour(RED);
-    printc(0, 11, '+');
+    printc(0, 14, '+');
     set_foreground_colour(WHITE);
     set_background_colour(BLUE);
     prints(0,28,                             "| ");
@@ -118,6 +115,13 @@ void display_init(void)
     set_background_colour(WHITE);
     clear_screen_region(ROW_FOOTER, 0, ROW_FOOTER, SCREEN_WIDTH - 1);
     prints(ROW_FOOTER, 0, " <ESC> exit  <F1> configuration  <Space> scroll lock");
+    prints(ROW_FOOTER, 64, "6.00.");
+    prints(ROW_FOOTER, 69, GIT_HASH);
+#if TESTWORD_WIDTH > 32
+    prints(ROW_FOOTER, 76, ".x64");
+#else
+    prints(ROW_FOOTER, 76, ".x32");
+#endif
     set_foreground_colour(WHITE);
     set_background_colour(BLUE);
 
--- a/app/display.c
+++ b/app/display.c
@@ -95,7 +95,7 @@ void display_init(void)
     prints(4, 0, "L3 Cache:   N/A             | Testing: ");
     prints(5, 0, "Memory  :   N/A             | Pattern: ");
     prints(6, 0, "--------------------------------------------------------------------------------");
-    prints(7, 0, "CPU:                     SMP: N/A         | Time:           Status: Init. ");
+    prints(7, 0, "CPU:                      SMP: N/A        | Time:           Status: Init. ");
     prints(8, 0, "Using:                                    | Pass:           Errors: ");
     prints(9, 0, "--------------------------------------------------------------------------------");
 
@@ -137,6 +137,11 @@ void display_init(void)
     if (clks_per_msec) {
         display_cpu_clk((int)(clks_per_msec / 1000));
     }
+    if (cpuid_info.flags.lm) {
+        display_cpu_addr_mode("X64");
+    } else if (cpuid_info.flags.pae) {
+        display_cpu_addr_mode("PAE");
+    }
     if (l1_cache) {
         display_l1_cache_size(l1_cache);
     }
@@ -172,9 +177,9 @@ void display_cpu_topology(void)
     int num_cpu_sockets = 1;
 
     if(smp_enabled) {
-        printf(7,30, "%uT (%s)", num_enabled_cpus, cpu_mode_str[cpu_mode]);
+        display_threading(num_enabled_cpus, cpu_mode_str[cpu_mode]);
     } else {
-        prints(7,30, "Disabled");
+        display_threading_disabled();
     }
 
     // If topology failed, assume topology according to APIC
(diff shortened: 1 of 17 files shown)


app/display.c
```diff
@@ -126,9 +138,9 @@ void display_init(void)
         display_cpu_clk((int)(clks_per_msec / 1000));
     }
     if (cpuid_info.flags.lm) {
-        display_cpu_addr_mode("(x64)");
+        display_cpu_addr_mode("X64");
```
martinwhitaker, line 141: Would it be better to use "LM " here?

app/display.c
```diff
@@ -159,10 +171,80 @@ void display_init(void)
     scroll_message_row = ROW_SCROLL_T;
 }
 
+void display_cpu_topology(void)
+{
+    extern int num_enabled_cpus;
+    int num_cpu_sockets = 1;
+
+    if(smp_enabled) {
```
martinwhitaker, line 179: Would prefer space between "if" and "(" (throughout this function). The rationale is this makes "if" statements visually distinct from function calls.

app/display.c
```diff
@@ -159,10 +171,80 @@ void display_init(void)
...
+        display_spec_mode("IMC: ");
+        display_spec(ram.freq, ram.type, ram.tCL, ram.tRCD, ram.tRP, ram.tRAS);
+        display_mode = 2;
+    } else if (ram.freq > 0 && ram.tCL > 1) {
+        // If not available, grab max memory specs from SPD
+        display_spec_mode("RAM: ");
+        display_spec(ram.freq, ram.type, ram.tCL, ram.tRCD, ram.tRP, ram.tRAS);
+        display_mode = 1;
+    } else {
+        // If nothing avilable, fallback to "Using Core" Display
```
martinwhitaker, line 245: Typo: avilable -> available.

app/display.c
```diff
@@ -63,6 +66,10 @@ static bool timed_update_done = false;  // update cycle status
 
 int scroll_message_row;
 
+int display_mode = 0; // RAM Info from: 0 = N/A - 1 = SPD - 2 = IMC
```
martinwhitaker, line 69: Better to use an enum for this.

app/display.h
```diff
@@ -36,57 +37,76 @@
...
-    printf(1, 10, "%i MHz", freq)
+    printf(1, 10, "%iMHz", freq)
 
 #define display_cpu_addr_mode(str) \
-    prints(1, 20, str)
+    prints(5, 76, str)
 
 #define display_l1_cache_size(size) \
-    printf(2, 10, "%kB", (uintptr_t)(size))
+    printf(2, 9, "%6kB", (uintptr_t)(size));
```
martinwhitaker, line 46: Unnecessary ";" (also in following definitions).

app/display.h
```diff
@@ -36,57 +37,76 @@
...
 #define display_l3_cache_size(size) \
-    printf(4, 10, "%kB", (uintptr_t)(size))
+    printf(4, 9, "%6kB", (uintptr_t)(size));
+
+#define display_memory_size(size) \
+    printf(5, 9, "%6kB", (uintptr_t)(size));
 
 #define display_l1_cache_speed(size) \
-    printf(2, 19, "%kB/s", (uintptr_t)(size))
+    printf(2, 18, "%S6kB/s", (uintptr_t)(size))
```
martinwhitaker, line 58: The parameter name should be "speed" not "size" (also in following definitions).

app/display.h
```diff
@@ -36,57 +37,76 @@
...
+
+#define display_threading_disabled() \
+    prints(7,31, "Disabled");
+
+#define display_cpu_topo_hybrid(nb) \
+    printf(7, 5, "%u Threads (Hybrid)", nb);
 
-#define display_available_cpus(count) \
-    printi(7, 10, count, 4, false, false)
+#define display_cpu_topo_multi_socket(nbs, nbc, nbt) \
```
martinwhitaker, line 81: An example of over-abbreviation. Would be easier to read with "(num_sockets, num_cores, num_threads)"

lib/print.h
```diff
@@ -52,18 +52,21 @@ int printx(int row, int col, uintptr_t value, int length, bool pad, bool left);
 /**
  * Prints a K<unit> value on screen starting at location (row,col) in a field of
  * at least length characters, optionally padding the number with leading zeros,
- * and optionally left-justifying instead of right-justifying in the field. The
- * value is shown to 3 significant figures in the nearest K/M/G/T units. Returns
- * the next column after the formatted value.
+ * optionally left-justifying instead of right-justifying in the field, and
+ * optionnaly adding a space between number and unit. The value is shown to
```
martinwhitaker, line 56: typo: "optionaly" -> "optionally"

system/cpuid.c
```diff
@@ -1,12 +1,14 @@
...
+// Copyright (C) 2004-2022 Sam Demeulemeester.
 //
 // Derived from memtest86+ cpuid.h
-// (original contained no copyright statement)
+
 
 #include <stdint.h>
 
 #include "cpuid.h"
+#include "display.h"
```
martinwhitaker, line 11: Why are you including display.h? The system library shouldn't have any dependencies on the application.

system/cpuid.h
```diff
@@ -133,6 +133,14 @@ typedef union {
     };
 } cpuid_custom_features;
 
+typedef struct {
+    int     core_count;
+    int     thread_count;
+    int     is_hybrid;
+    int     ecore_count;
+    int     pcore_count;
+} topology_t;
```
martinwhitaker, line 142: Better to name type "cpuid_topology_t".

martinwhitaker:
The first time the "<Enter> Single step" message is displayed in the footer, it only partially overwrites the build info.

app/display.c
x86fr: Fixed (spaces added)

app/display.c
x86fr: Fixed (enum added)

system/cpuid.c
x86fr: Remnant of a debug function. Removed.

app/display.c
x86fr: LM is more accurate but probably harder to understand the meaning. Do we need to display the paging mode on 64-bit binary? It's obvisouly LM (well, until we add support for PML5 in a far future...).

martinwhitaker: No, there's currently no need to display it on the 64-bit binary.

x86fr:
Ready to merge! :)

Issue #43, opened 2022-04-15
USB keyboard detection hangs on ECS K7S5A


debrouxl:
When I plug the usual HP KU-0316 USB keyboard, into one of the USB ports of my ECS K7S5A equipped with Athlon XP & DDR, the USB keyboard detection hangs: the "Press any key to continue..." string is not displayed within any reasonable amount of time.

This MB model is equipped with the following OHCI USB controllers, which are detected as such by memtest86+ 32b9ce38e7df78c4c569d26ded463760d1709927 + my SMBIOS parsing improvement ( #39 ):
```
# lspci
00:02.2 USB controller: Silicon Integrated Systems [SiS] USB 1.1 Controller (rev 07)
00:02.3 USB controller: Silicon Integrated Systems [SiS] USB 1.1 Controller (rev 07)

# lspci -n
00:02.2 0c03: 1039:7001 (rev 07)
00:02.3 0c03: 1039:7001 (rev 07)
```

```
# hwinfo --all
17: PCI 02.2: 0c03 USB Controller (OHCI)
  [Created at pci.386]
  SysFS ID: /devices/pci0000:00/0000:00:02.2
  SysFS BusID: 0000:00:02.2
  Hardware Class: usb controller
  Model: "Silicon Integrated USB 1.1 Controller"
  Vendor: pci 0x1039 "Silicon Integrated Systems Corp."
  Device: pci 0x7001 "USB 1.1 Controller"
  SubVendor: pci 0x1019 "Elitegroup Computer Systems"
  SubDevice: pci 0x0a14 "K7S5A motherboard"
  Revision: 0x07
  Driver: "ohci-pci"
  Driver Modules: "ohci_pci"
  Memory Range: 0xceffd000-0xceffdfff (rw,non-prefetchable)
  IRQ: 5 (no events)
  Module Alias: "pci:v00001039d00007001sv00001019sd00000A14bc0Csc03i10"
  Driver Info #0:
    Driver Status: ohci-hcd is active
    Driver Activation Cmd: "modprobe ohci-hcd"
  Driver Info #1:
    Driver Status: ohci_pci is active
    Driver Activation Cmd: "modprobe ohci_pci"
  Config Status: cfg=new, avail=yes, need=no, active=unknown

22: PCI 02.3: 0c03 USB Controller (OHCI)
  [Created at pci.386]
  SysFS ID: /devices/pci0000:00/0000:00:02.3
  SysFS BusID: 0000:00:02.3
[...]

PR #44, opened 2022-04-15
Add SPD decoding for the remaining (common) RAM types

debrouxl:
TODO:
- [x] (Sam) Add the DRDRAM formulas for capacity, freq, timings
- [x] (Sam) Add support for older SMBus controller models from Intel, e.g. 8086:2413 (ICH) on the Intel VC820, so that the code for SDRAM and DRDRAM parsing can be tested. Sam ? See https://github.com/anphsw/memtest86/blob/master/spd.c .
- [x] Split the commit in four parts: fixes on existing code, improvements on existing code, SDRAM decoding, DDR2 FB-DIMM decoding
- [x] (Lionel) Add the DDR2 FB-DIMM formula for capacity
- [x] (Lionel) Add the SDRAM formulas for capacity, freq, timings
- [ ] Add the DDR2 FB-DIMM formulas for freq, timings
- [ ] Test the code :)

This ought to occur before the feature freeze and beta test fest on older hardware.

### DIFF
--- a/system/smbus.c
+++ b/system/smbus.c
@@ -22,6 +22,7 @@ ram_info ram = { 0, 0, 0, 0, 0, "N/A"};
 int smbdev, smbfun;
 unsigned short smbusbase = 0;
 uint32_t smbus_id;
+static uint16_t extra_initial_sleep_for_smb_transaction = 0;
 
 static int8_t spd_page = -1;
 static int8_t last_adr = -1;
@@ -37,117 +38,22 @@ static spd_info parse_spd_ddr4  (uint8_t slot_idx);
 static spd_info parse_spd_ddr5  (uint8_t slot_idx);
 static void print_spdi(spd_info spdi, uint8_t lidx);
 
-static int find_smb_controller(void);
+static bool setup_smb_controller(void);
+static bool find_smb_controller(uint16_t vid, uint16_t did);
+
 static uint8_t get_spd(uint8_t slot_idx, uint16_t spd_adr);
 
-static void nv_mcp_get_smb(void);
-static void amd_sb_get_smb(void);
-static void fch_zen_get_smb(void);
-static void piix4_get_smb(void);
-static void ich5_get_smb(void);
+static bool nv_mcp_get_smb(void);
+static bool amd_sb_get_smb(void);
+static bool fch_zen_get_smb(void);
+static bool piix4_get_smb(void);
+static bool ich5_get_smb(void);
 static uint8_t ich5_process(void);
 static uint8_t ich5_read_spd_byte(uint8_t adr, uint16_t cmd);
 static uint8_t nf_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
 
-// ----------------------------------------------------------
-// WARNING: Be careful when adding a controller ID!
-// Incorrect SMB accesses (ie: on bank switch) can brick your
-// motherboard or your memory module.
-//                           ----
-// No Pull Request including a new SMBUS Controller will be
-// accepted without a proof (screenshot) that it has been
-// tested successfully on a real motherboard.
-// ----------------------------------------------------------
-
-static const struct pci_smbus_controller smbcontrollers[] = {
-    {0x8086, 0x2413, ich5_get_smb},    // 82801AA (ICH)
-    {0x8086, 0x2423, ich5_get_smb},    // 82801AB (ICH)
-    {0x8086, 0x2443, ich5_get_smb},    // 82801BA (ICH2)
-    {0x8086, 0x2483, ich5_get_smb},    // 82801CA (ICH3)
-    {0x8086, 0x24C3, ich5_get_smb},    // 82801DB (ICH4)
-    {0x8086, 0x24D3, ich5_get_smb},    // 82801E (ICH5)
-    {0x8086, 0x25A4, ich5_get_smb},    // 6300ESB
-    {0x8086, 0x266A, ich5_get_smb},    // 82801F (ICH6)
-    {0x8086, 0x269B, ich5_get_smb},    // 6310ESB/6320ESB
-    {0x8086, 0x27DA, ich5_get_smb},    // 82801G (ICH7)
-    {0x8086, 0x283E, ich5_get_smb},    // 82801H (ICH8)
-    {0x8086, 0x2930, ich5_get_smb},    // 82801I (ICH9)
-    {0x8086, 0x5032, ich5_get_smb},    // EP80579 (Tolapai)
-    {0x8086, 0x3A30, ich5_get_smb},    // ICH10
-    {0x8086, 0x3A60, ich5_get_smb},    // ICH10
-    {0x8086, 0x3B30, ich5_get_smb},    // 5/3400 Series (PCH)
-    {0x8086, 0x1C22, ich5_get_smb},    // 6 Series (PCH)
-    {0x8086, 0x1D22, ich5_get_smb},    // Patsburg (PCH)
-    {0x8086, 0x1D70, ich5_get_smb},    // Patsburg (PCH) IDF
-    {0x8086, 0x1D71, ich5_get_smb},    // Patsburg (PCH) IDF
-    {0x8086, 0x1D72, ich5_get_smb},    // Patsburg (PCH) IDF
-    {0x8086, 0x2330, ich5_get_smb},    // DH89xxCC (PCH)
-    {0x8086, 0x1E22, ich5_get_smb},    // Panther Point (PCH)
-    {0x8086, 0x8C22, ich5_get_smb},    // Lynx Point (PCH)
-    {0x8086, 0x9C22, ich5_get_smb},    // Lynx Point-LP (PCH)
-    {0x8086, 0x1F3C, ich5_get_smb},    // Avoton (SOC)
-    {0x8086, 0x8D22, ich5_get_smb},    // Wellsburg (PCH)
-    {0x8086, 0x8D7D, ich5_get_smb},    // Wellsburg (PCH) MS
-    {0x8086, 0x8D7E, ich5_get_smb},    // Wellsburg (PCH) MS
-    {0x8086, 0x8D7F, ich5_get_smb},    // Wellsburg (PCH) MS
-    {0x8086, 0x23B0, ich5_get_smb},    // Coleto Creek (PCH)
-    {0x8086, 0x8CA2, ich5_get_smb},    // Wildcat Point (PCH)
-    {0x8086, 0x9CA2, ich5_get_smb},    // Wildcat Point-LP (PCH)
-    {0x8086, 0x0F12, ich5_get_smb},    // BayTrail (SOC)
-    {0x8086, 0x2292, ich5_get_smb},    // Braswell (SOC)
-    {0x8086, 0xA123, ich5_get_smb},    // Sunrise Point-H (PCH)
-    {0x8086, 0x9D23, ich5_get_smb},    // Sunrise Point-LP (PCH)
-    {0x8086, 0x19DF, ich5_get_smb},    // Denverton  (SOC)
-    {0x8086, 0x1BC9, ich5_get_smb},    // Emmitsburg (PCH)
-    {0x8086, 0xA1A3, ich5_get_smb},    // Lewisburg (PCH)
-    {0x8086, 0xA223, ich5_get_smb},    // Lewisburg Super (PCH)
-    {0x8086, 0xA2A3, ich5_get_smb},    // Kaby Lake (PCH-H)
-    {0x8086, 0x31D4, ich5_get_smb},    // Gemini Lake (SOC)
-    {0x8086, 0xA323, ich5_get_smb},    // Cannon Lake-H (PCH)
-    {0x8086, 0x9DA3, ich5_get_smb},    // Cannon Lake-LP (PCH)
-    {0x8086, 0x18DF, ich5_get_smb},    // Cedar Fork (PCH)
-    {0x8086, 0x34A3, ich5_get_smb},    // Ice Lake-LP (PCH)
-    {0x8086, 0x38A3, ich5_get_smb},    // Ice Lake-N (PCH)
-    {0x8086, 0x02A3, ich5_get_smb},    // Comet Lake (PCH)
-    {0x8086, 0x06A3, ich5_get_smb},    // Comet Lake-H (PCH)
-    {0x8086, 0x4B23, ich5_get_smb},    // Elkhart Lake (PCH)
-    {0x8086, 0xA0A3, ich5_get_smb},    // Tiger Lake-LP (PCH)
-    {0x8086, 0x43A3, ich5_get_smb},    // Tiger Lake-H (PCH)
-    {0x8086, 0x4DA3, ich5_get_smb},    // Jasper Lake (SOC)
-    {0x8086, 0xA3A3, ich5_get_smb},    // Comet Lake-V (PCH)
-    {0x8086, 0x7AA3, ich5_get_smb},    // Alder Lake-S (PCH)
-    {0x8086, 0x51A3, ich5_get_smb},    // Alder Lake-P (PCH)
-    {0x8086, 0x54A3, ich5_get_smb},    // Alder Lake-M (PCH)
-    {0x8086, 0x7A23, ich5_get_smb},    // Raptor Lake-S (PCH)
-
-    // ATI SMBUS
-    {0x1002, 0x4385, amd_sb_get_smb},  // ATI SB600+ (Now AMD)
-
-    // AMD SMBUS
-    {0x1022, 0x780B, amd_sb_get_smb},  // AMD FCH (Pre-Zen)
-    {0x1022, 0x790B, fch_zen_get_smb}, // AMD FCH (Zen)
-
-    // nVidia SMBUS
-    // {0x10DE, 0x01B4, nv_mcp_get_smb},  // nForce
-    {0x10DE, 0x0064, nv_mcp_get_smb},     // nForce 2
-    // {0x10DE, 0x0084, nv_mcp_get_smb},  // nForce 2 Mobile
-    // {0x10DE, 0x00E4, nv_mcp_get_smb},  // nForce 3
-    // {0x10DE, 0x0052, nv_mcp_get_smb},  // nForce 4
-    // {0x10DE, 0x0264, nv_mcp_get_smb},  // nForce 410/430 MCP
-    // {0x10DE, 0x0446, nv_mcp_get_smb},  // nForce 520
-    // {0x10DE, 0x0542, nv_mcp_get_smb},  // nForce 560
-    // {0x10DE, 0x07D8, nv_mcp_get_smb},  // nForce 630i
-    {0x10DE, 0x03EB, nv_mcp_get_smb},  // nForce 630a
-    {0x10DE, 0x0752, nv_mcp_get_smb},  // nForce 720a
-    // {0x10DE, 0x0AA2, nv_mcp_get_smb},  // nForce 730i
-    // {0x10DE, 0x0368, nv_mcp_get_smb},  // nForce 790i Ultra
-
-    {0, 0, NULL}
-};
-
-void print_smbus_startup_info(void) {
-
-    int8_t index;
+void print_smbus_startup_info(void)
+{
     uint8_t spdidx = 0, spd_line_idx = 0;
 
     spd_info curspd;
@@ -158,15 +64,7 @@ void print_smbus_startup_info(void) {
         quirk.process();
     }
 
-    index = find_smb_controller();
-
-    if (index == -1) {
-        return;
-    }
-
-    smbcontrollers[index].get_adr();
-
-    if (smbusbase == 0) {
+    if (!setup_smb_controller() || smbusbase == 0) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


x86fr:
The scope of this PR is probably too wide as I don't want to wait for the whole list being complete to merge. 
The addition of timings parse (now mandatory on all spd parsing functions) make it a bit more complex to add than before and I never add any code that hadn't been tested (so I must dig my basement to find a FB-DIMM motherboard :D)

Code factorization is nice but it's also great to have easily readable code and factorization may break it (especially where the code is very linear). Not sure it's worth the hundreds bytes saved here.

debrouxl:
As you wish, we'll add the memory types as separate commits :)
I have readily accessible computers (that is, permanently wired to a PSU and inside a case) for all three memory types, and it's easy to feed SPD data dumps from real sticks, obtained from Windows or Linux, to the decoder.
(At some point, we'll probably want unit tests for the SPD decoder, but that's another matter)

* both of my computers equipped with DDR2 FB-DIMM (HP xw6600 workstation, Dell PowerEdge 2950) use Xeon [EX]54xx processors, so the PCI IDs of their SMBus controllers probably already belong to the list. The formulas for computing memory size and frequency still need to be plugged into the code, though, and I don't think I can do that today, especially since I need to test other things, e.g. for #43 .
* if testing DRDRAM on our VC820s *in a safe and reliable manner* is as simple as adding a
```
{0x8086, 0x2413, "82801AA (ICH)",           ich5_get_smb, ich5_read_spd_byte},
```
entry at the beginning of the table, I can do it as soon as I read the confirmation. But the formulas for memory size and speed still need to be implemented.
* SDRAM SPD decoding is going to be harder to test on my side. My main SDRAM test machine, the ECS K7S5A (DDR + SDRAM), is new enough to have a SMBus controller, but it's a SiS model which would require a whole new driver. I have a PIIX4 MB, which hopefully has a SMBus controller, but PIIX4 support hasn't yet been reimplemented in the memtest86+ 6.00 code base.

x86fr:
Can you test your VC820 with the last commit? RDRAM detection works on my Abit TH7 and I can't remember where is my VC820.

image

I think I will now switch to modern AMD platform that really deserve some attention.

debrouxl:
* on the VC820, a build of e0d4566 displays:
```
...
RAM: 355 MHz (RDRAM-711) CAS 8-8-8-20
...
 - Slot 0 : 128 MB RDRAM-711 - GTE 4R128CEE6C-745
 - Slot 1 : 128 MB RDRAM-711 - GTE 4R128CEE6C-745
```
so it works for me as well  :)

* `decode-dimms` can gather information from all four memory sticks of the dual PII Xeon computer through `/sys/bus/i2c/drivers/eeprom/0-005*/eeprom`, but the SPD access functionality is provided by a PCI device which does not specifically appear as SMBus controller:
```
00:00.0 Host bridge: Intel Corporation 440GX - 82443GX Host bridge
00:01.0 PCI bridge: Intel Corporation 440GX - 82443GX AGP bridge
00:07.0 ISA bridge: Intel Corporation 82371AB/EB/MB PIIX4 ISA (rev 02)
00:07.1 IDE interface: Intel Corporation 82371AB/EB/MB PIIX4 IDE (rev 01)
00:07.2 USB controller: Intel Corporation 82371AB/EB/MB PIIX4 USB (rev 01)
00:07.3 Bridge: Intel Corporation 82371AB/EB/MB PIIX4 ACPI (rev 02)

00:00.0 0600: 8086:71a0
00:01.0 0604: 8086:71a1
00:07.0 0601: 8086:7110 (rev 02)
00:07.1 0101: 8086:7111 (rev 01)
00:07.2 0c03: 8086:7112 (rev 01)
00:07.3 0680: 8086:7113 (rev 02)
```

debrouxl:
`decode-dimms` uses the following code to compute the size of SDRAM sticks in MBs:
```
	$ii = ($bytes->[3] & 0x0f) + ($bytes->[4] & 0x0f) - 17;
	if (($bytes->[5] <= 8) && ($bytes->[17] <= 8)) {
		 $k = $bytes->[5] * $bytes->[17];
	}

	if ($ii > 0 && $ii <= 12 && $k > 0) {
		printl("Size", ((1 << $ii) * $k) . " MB");
	} else {
		printl("Size", "INVALID: " . $bytes->[3] . "," . $bytes->[4] . "," .
			       $bytes->[5] . "," . $bytes->[17]);
	}
```
Easy enough, after a bit of hardening: the sum of the 4 LSB of bytes 3 and 4 might be <= 17 on improper devices.
However, JEDEC 21-C mentions asymmetric banking in bytes 3 and 4, as well as up to 14 banks in byte 5. Did some manufacturers implement these in real-world devices ?

debrouxl:
I've reimplemented the above, in the meantime. This leaves only the frequency formulas to be implemented and tested - preferably with suitable hardware and/or drivers, but now is probably a good time to implement a CLI client for exercising the SPD detection code...

No way to obtain a SPD dump of the FB-DIMMs on the PowerEdge 2950 either, sadly. I'll have to find FB-DIMM SPD data dumps elsewhere.

x86fr:
I can probably provide them from CPU-Z dumps (but I'm not at home right now).

Anyway, don't forget to add timings (CAS-RCD-RP-RAS) as they're now mandatory.

x86fr:
Here are some SPD from FB-DDR2

Seems there is timings table, so not that easy (timings must be reported at tCycleMin).

```
DIMM #				1
	SMBus address		0x0
	Memory type		FB-DDR2
	Module Manufacturer(ID)	SK Hynix (AD00000000000000000000000000)
	Size			2048 MBytes
	Max bandwidth		PC2-5300 (333 MHz)
	Part number		HYMP525F72CP4N3-Y5
	Serial number		49450208
	Manufacturing date	Week 15/Year 08
	Number of banks		4
	EPP			no
	XMP			no
	AMP			no
JEDEC timings table		CL-tRCD-tRP-tRAS-tRC @ frequency
	JEDEC #1		4.0-4-4-12-16 @ 266 MHz
	JEDEC #2		5.0-5-5-15-20 @ 333 MHz

DIMM #				1
SPD registers	
		00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F 
	00	92 11 09 12 48 23 07 10 A1 01 04 0C 20 24 3C 42 
	10	3C 42 40 3C 1E 3C 00 B4 F0 A4 01 1E 1E 03 07 01 
	20	C2 57 90 80 37 24 2E 52 2C 3C 00 00 00 00 00 00 
	30	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
	40	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 22 
	50	00 02 00 10 58 42 38 30 6A 84 6E AF 8B 00 00 00 
	60	00 00 00 8A 00 80 20 00 44 04 80 48 53 B1 43 65 
	70	4C 00 10 80 89 80 AD 01 08 15 49 45 02 08 56 7E 
	80	48 59 4D 50 35 32 35 46 37 32 43 50 34 4E 33 2D 
	90	59 35 00 00 80 AD 01 09 00 00 00 00 00 00 00 00 
	A0	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
	B0	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
	C0	53 55 50 45 52 4D 49 43 52 4F 00 00 00 00 00 00 
	D0	58 37 44 42 38 00 00 00 00 00 00 00 00 00 00 00 
	E0	01 03 01 02 03 09 00 00 00 00 00 00 00 01 00 00 
	F0	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 

DIMM #				2
	SMBus address		0x1
	Memory type		FB-DDR2
	Module Manufacturer(ID)	Samsung (CE00000000000000000000000000)
	Size			2048 MBytes
	Max bandwidth		PC2-5300 (333 MHz)
	Part number		M395T5750EZ4-CE66 
	Serial number		051A2302
	Manufacturing date	Week 47/Year 07
	Number of banks		4
	EPP			no
	XMP			no
	AMP			no
JEDEC timings table		CL-tRCD-tRP-tRAS-tRC @ frequency
	JEDEC #1		3.0-3-3-9-12 @ 200 MHz
	JEDEC #2		4.0-4-4-12-16 @ 266 MHz
	JEDEC #3		5.0-5-5-15-20 @ 333 MHz

DIMM #				2
SPD registers	
		00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F 
	00	92 11 09 12 48 24 07 10 52 01 04 0C 20 33 3C 42 
	10	3C 32 40 3C 1E 3C 00 B4 F0 A4 01 1E 1E 03 07 01 
	20	C2 51 3E 14 15 1F 16 23 0D 15 00 00 00 00 00 00 
	30	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
	40	00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 22 
	50	00 02 00 10 36 34 32 2A 62 77 61 9F 84 00 00 00 
	60	00 00 1F CA 00 00 E2 62 20 80 9C 00 00 F0 70 60 
	70	60 60 60 7F B3 80 CE 01 07 47 05 1A 23 02 95 8E 
	80	4D 33 39 35 54 35 37 35 30 45 5A 34 2D 43 45 36 
	90	36 20 00 00 80 CE 00 00 00 59 47 42 35 34 30 35 
	A0	01 00 01 01 00 00 00 00 00 00 00 00 00 00 00 08 
	B0	00 00 00 00 00 00 00 00 00 00 00 00 00 00 24 0E 
	C0	11 01 43 5A 43 38 30 32 33 38 4E 46 00 00 00 00 
	D0	00 00 0E 11 07 62 02 00 00 05 00 00 00 00 FF FF 
	E0	FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
	F0	FF FF FF FF FF FF 00 00 00 00 00 00 00 00 00 00 
```

debrouxl:
The SDRAM decoder now seems to be in a decent shape, in that when the interactive testing code (now found on the fuzzing branch) is used to feed it SPD dumps from my real memory sticks and display the result, the output has no deviations from the expected content which are immediately obvious to me:
```
 - Slot 0 : 131072KB SDRAM-133 ECC - InfineonHYS72V16301GR-7.5 (W01'02)
tCL: 1  tRCD: 2 tRP: 2  tRAS: 6 tRC: 0
 - Slot 0 : 131072KB SDRAM-133 ECC - InfineonHYS72V16301GR-7.5 (W15'02)
tCL: 1  tRCD: 2 tRP: 2  tRAS: 6 tRC: 0
 - Slot 0 : 262144KB SDRAM-133 ECC - InfineonHYS72V32301GR-7.5 (W42'00)
tCL: 1  tRCD: 2 tRP: 2  tRAS: 6 tRC: 0
```
(I had to comment out the `&& spdi.fab_year > 1` test in `print_spdi()`, BTW)

DDR2 FB-DIMM decoding is quite a bit more annoying :'(

debrouxl:
I have pushed the unfinished DDR2 FB-DIMM changes aside for now, so that the subset of changes which is ready for merging can be merged.

x86fr:
Planned for merge on Friday

debrouxl:
IMO, for this branch, rebase and merge should be used instead of squash and merge: the three commits have different purposes.

In the CPU-Z data, could you find several computer models with both DDR2 FB-DIMM and a SMBus controller model currently supported by memtest86+ ?

x86fr:
Did you tested the SDRAM code on a real platform? I finally took some time to boot an Asus TUSL2 (the most modern SDRAM platform I have with an ICH2) and the smbus is not detected at all. Memtest86+ works as expected, but as if the code was just not here. 

Before trying to debug, I wonder if it's just this motherboard or if this is a general problem. Unfortunately, my P3C2000 don't want to boot and the others boards I have with SDRAM are AMD or PIIX4-based.

debrouxl:
Sadly, nope, I haven't been able to test it on a real SDRAM platform here. My 3 SDRAM MBs whose SMBus controllers are enabled and supported by Linux (I've successfully gathered SPD data dumps from these MBs) but not yet by memtest86+:
* a HP workstation with dual PII Xeon and PIIX4 chipset, broken DMI information so it requires booting with Linux `acpi=off` parameter to take advantage of SMP under Linux;
* Biostar M7VKA with Athlon and VIA chipset, no dedicated SMBus controller on the PCI bus but said SMBus controller + RAM are supported by Linux `i2c_viapro` + `eeprom` modules;
* ECS K7S5A with Athlon XP and SiS chipset, SMBus controller 1039:0016 on the PCI bus.

There's also Dell Optiplex G1 with PIII and PIIX4 chipset, but Linux `dmesg` contains "piix4_smbus 0000:00:07.3: SMBus Host Controller not enabled!", so this one is unusable. I'd have to try whether `acpi=off` helps SMBus operation on that computer.

x86fr:
Merge delayed a bit (now planned for Beta 2). 

There's something broken that need investigation. Fortunately, the TUSL2-C has a SMBUS header, so this shouldn't be too hard to debug. I'll hook up my logic analyzer and check what's wrong on the bus.

x86fr:
I've found the issue. On the TUSL2-C, there is 2 multiplexed smbus, one with various monitoring controllers (temperature, ADC, etc.), and the second with RAM EEPROMs. By default, the multiplexer is located inside the ASUS's ASB100 ASIC and requires proprietary code to switch between both buses. Worst, the code is specific to the TUSL2-C even if some other ASUS boards like the P4G8X uses the same ASB100.

It seems that a vast majority of ICH0/1/2-based motherboards require proprietary tricks to enable SPD access (if possible). Adding such undocumented, obscure and specific tricks is clearly not on top of the list, but I'll add the TUSL2-C as an example on the code.

debrouxl:
> Adding the TUSL2 trick and the DDR2-FB require more code than expected, so I think it's better to merge this quickly to get a more clean main (as there is many small but important bug fixes in this PR).

Alright.

I'll handle the merge, in order to restore the commit message for 01e3's commit, and modify your commit to add back the unsupported SMBus controllers as comments (it took me enough time to build up the list that there's no point in having someone unwittingly redo that work) :)

x86fr:
Ok. Try to keep only smbus controllers device_ids that have a chance to be implemented one day (ie: that can communicate with SPDs and not dedicated to non-x86 platforms).

debrouxl:
Sure.
I had a look at the Linux PCI IDs list to build the list of x86-related SMBus devices I added previously. This time, in order to check for SPD communication presence, I used the Linux drivers in `drivers/i2c/busses/i2c-*`, selected files which contain "SMBus", filtered out platforms which clearly aren't x86, then filtered on I2C_CLASS_SPD... Well, it looks like I missed a bunch of old I2C controllers with SPD capability :)

The relevant Linux I2C drivers are:
```
i2c-ali1535.c
i2c-ali1563.c
i2c-ali15x3.c
i2c-amd756.c + i2c-amd756-s4882.c
i2c-amd8111.c
i2c-i801.c
i2c-isch.c (based on the `drivers/mfd/lpc_sch.c` platform driver, which defines SMBus ranges for several PCI IDs)
i2c-nforce2.c + i2c-nforce2-s4985.c
i2c-piix4.c
i2c-sis5595.c
i2c-sis630.c
i2c-sis96x.c
i2c-via.c
i2c-viapro.c
```
i2c-ismt.c only advertises I2C_CLASS_HWMON for multiple Atom SMBus controller models spanning 6 PCI IDs, which looks odd for hardware of that generation.

I added back a number of PCI IDs from my list, removed several ones, and added several ones found in the Linux drivers. TBC later.

debrouxl:
As for the rewrite changing a struct into a bunch of nested switches... I looked at the resulting code before merging the PR, and sadly, it proves to be poorly optimized size-wise, with cases for `did`s ranging from 16 bytes to 6 bytes:
```
    50b8:       66 3d a3 34             cmp    $0x34a3,%ax
    50bc:       0f 84 e5 01 00 00       je     52a7 <print_smbus_startup_info+0x28e>
    50c2:       0f 87 1c 01 00 00       ja     51e4 <print_smbus_startup_info+0x1cb>

    50fd:       66 3d a3 02             cmp    $0x2a3,%ax
    5101:       e9 9f 01 00 00          jmp    52a5 <print_smbus_startup_info+0x28c>

    5225:       66 3d a3 43             cmp    $0x43a3,%ax
    5229:       74 7c                   je     52a7 <print_smbus_startup_info+0x28e>
```
The much smaller switch on the `vid` adds to that.

The name string in my optimized version of `struct pci_smbus_controller` was by far the main contributor to the size of this struct; without it, my optimized version of `struct pci_smbus_controller` reduces to 6 bytes, which leaves room for storing flags. The array of `struct pci_smbus_controller_ops` adds to the size of the array of optimized `struct pci_smbus_controller`, though the latter dwarfs the former.

IOW, your rewrite with nested switches is certainly a size pessimization over a 6-byte, or even a hypothetical 8-byte version containing extra expanded flags or whatever else isn't envisioned right now, of `struct pci_smbus_controller`, plus the 2 -> 3 instances of `struct pci_smbus_controller_ops` when PIIX4 is reimplemented (and unless someone contributes support for the older SMBus controllers, the count might stay at that value for quite a while), plus code to traverse these structs. I think that we can, and should, build further on the simplification (removal of function arguments) brought by your rewrite :)

x86fr:
Agree, but while it's desirable to keep code as optimized as possible, we're not really limited by the code size anymore. The most important is to keep the code readable and easy to understand for potential contributors, even at the expense of a couple hundreds of bytes (or even few kBs). 

Anyway, can you merge/reabse this asap to main?

01e3:
If it helps, please don't worry too much about the commit message, maybe next time.... ;)

However, I would like to ask about the direction we are going here - having memtest being able to recognize the chipset and state if I2C has been detected and is supported, seems like an useful feature to me. Similar like printing the CPU name...

However, I noticed all the chipset strings ended up being removed in favor of a long 'case / switch' section. Perhaps it reduces the code size, but is this the path we are taking here? I think we can benefit from a similar optimization by making the struct two-level which should avoid all these 0x8086...

Anyway, this way or another, can we at least show the PCI ID of the SMBus somewhere on the screen if it is detected? In the future, if SPD is not available (like in my HP z444 case) we can at least know if this is WAI (as the chipset is not supported) or there is some other issues?

debrouxl:
I disagree with nested switches being more readable and maintainable than an array of structs... which is the way the Linux drivers for PCI devices do it, BTW :)
```
drivers/i2c/busses/i2c-amd8111.c:
static const struct pci_device_id amd8111_ids[] = {
	{ PCI_DEVICE(PCI_VENDOR_ID_AMD, PCI_DEVICE_ID_AMD_8111_SMBUS2) },
	{ 0, }
};

drivers/i2c/busses/i2c-viapro.c (reformatted):
static const struct pci_device_id vt596_ids[] = {
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_82C596_3), .driver_data = SMBBA1 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_82C596B_3), .driver_data = SMBBA1 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_82C686_4), .driver_data = SMBBA1 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8233_0), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8233A), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8235), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8237), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8237A), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8237S), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8231_4), .driver_data = SMBBA1 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_8251), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_CX700), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_VX800), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_VX855), .driver_data = SMBBA3 },
	{ PCI_DEVICE(PCI_VENDOR_ID_VIA, PCI_DEVICE_ID_VIA_VX900), .driver_data = SMBBA3 },
	{ 0, }
};

drivers/mfd/lpc_sch.c:
static const struct pci_device_id lpc_sch_ids[] = {
	{ PCI_VDEVICE(INTEL, PCI_DEVICE_ID_INTEL_SCH_LPC), LPC_SCH },
	{ PCI_VDEVICE(INTEL, PCI_DEVICE_ID_INTEL_ITC_LPC), LPC_ITC },
	{ PCI_VDEVICE(INTEL, PCI_DEVICE_ID_INTEL_CENTERTON_ILB), LPC_CENTERTON },
	{ PCI_VDEVICE(INTEL, PCI_DEVICE_ID_INTEL_QUARK_X1000_ILB), LPC_QUARK_X1000 },
	{ 0, }
};
```
Due to the quirks of some devices, PIIX4 support is going to make the nested switches code more complex - and so is support for bus MUXes, necessary on some computers equipped with i801 or PIIX4, as we've seen before.
Simplicity is good, but it's hard not to make things overly simple and therefore limited as a result.

I've already restored the commit message from #74 , no problem about that :)
Yes, showing the PCI IDs for detected but unsupported SMBus controllers, not all of which have 0x0c05 PCI type, is a good idea.

x86fr:
> However, I would like to ask about the direction we are going here - having memtest being able to recognize the chipset and state if I2C has been detected and is supported, seems like an useful feature to me. Similar like printing the CPU name...

If I2C has been detected and supported, `Memory SPD Information` will be printed.

> However, I noticed all the chipset strings ended up being removed in favor of a long 'case / switch' section. Perhaps it reduces the code size, but is this the path we are taking here? I think we can benefit from a similar optimization by making the struct two-level which should avoid all these 0x8086...

These strings were there since 15+ years and unused since then. Nobody ever asked to have the name of the SMBUS controller displayed in a memory testing tool. Memtest86+ is not CPU-Z not dmi-decode. 

> Anyway, this way or another, can we at least show the PCI ID of the SMBus somewhere on the screen if it is detected? In the future, if SPD is not available (like in my HP z444 case) we can at least know if this is WAI (as the chipset is not supported) or there is some other issues?

Really, I don't see the point to display the PCI ID of the SMBus Controller on the screen. Even adding this as a debug mode with a specific flag is not really interesting. Finding the SMBus Controller on the PCI Bus is trivial.

debrouxl:
> The code to handle quirks and a first sample code with the TUSL2-C is already wrote, but I can't push it because the branch diverged since my last commit due to the rewrite.

Could you push that on a new branch ? I'm interested :)

debrouxl:
> Really, I don't see the point to display the PCI ID of the SMBus Controller on the screen.

One point is to provide feedback to users and ultimately to maintainers and contributors: distinguishing between "no SMBus controller found" and "SMBus controller found but not supported yet, either use one of the dedicated hardware inventory tools for your favorite OS or raise your hand and it may increase the likelihood of seeing it implemented in memtest86+ by someone".

> Feel free to remove the nested switch array and revert to the function pointer table.

Nah, there's a better way to eliminate the section of poor generated code :)
See my latest commit, untested because I don't have access to my computers tonight: I kept the control flow, i.e. the main switch + most nested switches, and only changed the sparse switch on Intel `did`s to an array + linear table lookup. 

Together, the three remaining commits on the `spd_decoding` branch - which probably ought to be squashed after I add some more PCI IDs - shrink .text + .data + .bss by over 6.5 KB on x86_64.

x86fr:
> One point is to provide feedback to users and ultimately to maintainers and contributors: distinguishing between "no SMBus controller found" and "SMBus controller found but not supported yet, either use one of the dedicated hardware inventory tools for your favorite OS or raise your hand and it may increase the likelihood of seeing it implemented in memtest86+ by someone".

We already have all these information. Finding the SMBus controller is never the issue. We have a list of all supported and unsupported controller, so if the SPD Information are not displayed, just with the name of the motherboard, I can tell if the issue is the SMBus controller not supported (if it's not in the list) or if there is an issue reading it (if it's in the list).

> Nah, there's a better way to eliminate the section of poor generated code :) See my latest commit, untested because I don't have access to my computers tonight: I kept the control flow, i.e. the main switch + most nested switches, and only changed the sparse switch on Intel `did`s to an array + linear table lookup.
> 
> Together, the three remaining commits on the `spd_decoding` branch - which probably ought to be squashed after I add some more PCI IDs - shrink .text + .data + .bss by over 6.5 KB on x86_64.

Now that we know that the nested switches are pointless, I still don't understand why we absolutely need to change the original, very easy to understand and finally quite efficient, struct 

Doing more than 1/ removing the string and 2/ the unused function pointer from this struct, seems useless to me . 

Ie:
```
static const struct pci_smbus_controller smbcontrollers[] = {
    {0x8086, 0x2413, "82801AA (ICH)",           ich5_get_smb, ich5_read_spd_byte},
```
to 
```
static const struct pci_smbus_controller smbcontrollers[] = {
    {0x8086, 0x2413, ich5_get_smb},   // 82801AA (ICH)
```

debrouxl:
That still leaves a highly redundant structure, 16 bytes + relocation per entry due to alignment, or 12 bytes + relocation per entry with `__attribute__((packed))`. No match for 2-byte entries (and fewer of them, too).

debrouxl:
TL;DR: the form to which this PR has converged thanks to the lengthy discussion is an improvement over the current code on the main branch, *and not just for efficiency reasons*... so I'm raising the matter one more time ;)

Sam, with the combination of your (partially merged) work and the remainder of my work, we're onto something.
The current state of a switch on PCI vendor IDs + a table of PCI device IDs where it matters - i.e. exclusively for Intel SMBus controllers at the moment, and probably for much longer unless someone contributes nForce2 and viapro SMBus controller support - is:
* both a readability and a size efficiency improvement over my earlier suggestion of an array of structs which still contains redundancy and an indirection through another array of structs;
* a size efficiency improvement over exclusive usage of nested switches, without raising complexity and therefore impairing maintainability in any remotely serious way. As Intel likes to produce new PCI IDs at every generation, the number of Intel SMBus controllers should keep increasing, so the difference between a table without function pointers and a table with function pointers will keep increasing;
* both an efficiency and a maintainability improvement over any variant of an array of structs containing a function pointer.

Indeed, pushing the function calls from a table into the code brings the ability to have different prototypes for the `*_get_smb()` function. I'm not making a purely hypothetical case just to try to annoy you or whatever: for instance, the Linux PIIX4 I2C driver shows that the ServerWorks CSB5 needs a quirk, namely extra delay in the PIIX4 transaction - and my IBM xSeries 345 has that device model:
```
00:0f.0 Host bridge: Broadcom CSB5 South Bridge (rev 93)
00:0f.0 0600: 1166:0201 (rev 93)
``` 

Clearly, creating a MB quirk entry + a function for handling the affected PCI VID+DID pairs which would set a global variable from smbus.c + making `piix4_get_smb()` and `ich5_process()` read this global variable to act accordingly is technically _possible_. However, handling this quirk entirely from inside smbus.c is easier to follow and maintain, besides being a tiny bit smaller. In this case, it would be a matter of storing the quirk information from the switch before calling `piix4_get_smb()`; in more complex cases, passing a new argument to the sole relevant `*_get_smb()` function - without impacting all other `*_get_smb()` functions which don't need such a thing, as would be the case with the existing array of structs containing a function pointer - and storing the quirk information inside the function. In both cases, the quirk information will be used by `ich5_process()` later.

x86fr:
I think the main problem here is that we're trying to work simultaneously on the same part of the code and probably we've also put the cart before the horse: the code is evolving quickly and some new controllers/quirks may or may not break these tweaks. Working on the optimization a bit later, when the code will be almost stable would probably have been more efficient. 

Anyway, it's really time to move on. The switch/pointer function looks readable enough to me, so let's use it.

I'll do some small comments right now, then this could be merged as soon as conflicts are solved.

system/smbus.c
```diff
@@ -1189,35 +1119,220 @@ static spd_info parse_spd_sdram(uint8_t smb_idx, uint8_t slot_idx)
...
+}
+
+
+static bool find_smb_controller(uint16_t vid, uint16_t did) {
+
+    switch(vid)
+    {
+        case VID_INTEL:
+        {
+            if (find_in_did_array(did, intel_ich5_dids, sizeof(intel_ich5_dids) / sizeof(intel_ich5_dids[0]))) {
```
x86fr, line 1221: Would avoid passing the length of the array by doing something like that:

```
--- a/system/smbus.c
+++ b/system/smbus.c
@@ -1200,11 +1200,16 @@ static const uint16_t intel_ich5_dids[] =
     0x7AA3, // Alder Lake-S (PCH)
     0x51A3, // Alder Lake-P (PCH)
     0x54A3, // Alder Lake-M (PCH)
+    0xFFFF
 };

-static bool find_in_did_array(uint16_t did, const uint16_t * ids, unsigned int size) {
-    for (unsigned int i = 0; i < size; i++) {
-        if (*ids++ == did) {
+static bool find_in_did_array(uint16_t did, const uint16_t *ids)
+{
+    uint16_t curdid = 0;
+
+    while (curdid != 0xFFFF) {
+        curdid = *ids++;
+        if (curdid == did) {
             return true;
         }
     }
@@ -1212,13 +1217,14 @@ static bool find_in_did_array(uint16_t did, const uint16_t * ids, unsigned int s
 }


-static bool find_smb_controller(uint16_t vid, uint16_t did) {
+static bool find_smb_controller(uint16_t vid, uint16_t did)
+{

     switch(vid)
     {
         case VID_INTEL:
         {
-            if (find_in_did_array(did, intel_ich5_dids, sizeof(intel_ich5_dids) / sizeof(intel_ich5_dids[0]))) {
+            if (find_in_did_array(did, intel_ich5_dids)) {
                 return ich5_get_smb_base();
             }
```

system/smbus.c
```diff
@@ -1189,35 +1119,220 @@ static spd_info parse_spd_sdram(uint8_t smb_idx, uint8_t slot_idx)
...
+*/
+                case 0x790B: // AMD FCH (Zen 2/3)
+                    return fch_zen_get_smb_base();
+                default:
+                return false;
+            }
+            break;
 
-    return -1;
+        case VID_HYGON:
```
x86fr, line 1256: Hygon's CPU are strictly identical to AMD's CPU (at least from a SMBus PoV) as they're only rebranded EPYC with some features fused-down, so this can be handled on the same case.

system/smbus.c
```diff
@@ -1189,35 +1119,220 @@ static spd_info parse_spd_sdram(uint8_t smb_idx, uint8_t slot_idx)
...
+            {
+                case 0x790B: // AMD FCH (Zen 2/3)
+                    return fch_zen_get_smb_base();
+                default:
+                    return false;
+            }
+            break;
+
+/*
+      case VID_ATI:
```
x86fr, line 1267: Having a list of DID ready for future implementation is nice, but maybe moving them all as a condensed list at the end of the function is better than a lot of commented pseudo-code (suggesting that there is only a function to code to add support, which might be false).

system/smbus.c
debrouxl: Indeed :)

system/smbus.c
debrouxl: OK, provided the corresponding controller type / driver information is kept.
For VID_VIA: the Linux i2c-via driver handles `0x3040`, whereas the i2c-viapro driver handles
```
0x3050
0x3051
0x3057
0x3074
0x3147
0x3149
0x3177
0x3287
0x3337
0x3372
0x8235
0x8324
0x8353
0x8409
0x8410
```
(list of DIDs sorted by ascending order)

x86fr: I just realized that I can parse the CPU-Z database and extract all SMBUS controller IDs present, that should be almost 100% of the IDs used in real-world.  I can also get the % for each of them, so we will know on which ones we must focus.

x86fr: Results are posted in #86

debrouxl:
Now that 6.00b2 is released, I'll handle the rebase, alright ?
I plan on depending on the switch + array of PCI DIDs to add Serverworks CSB5 support, as detailed above.

x86fr:
Sounds good.

debrouxl:
Rebased; however, I need to test it :)

debrouxl:
On the xSeries 345 with ServerWorks CSB5, the SPD gathering works:
P1100849
DDR-285 is close enough to the DDR-266 the server is using, but inexact.

I need to test on a wider range of computers and chipsets, but not right now :)

debrouxl:
I've wired up the standard PIIX4 device ID, tested on the aforementioned dual PII Xeon 350 (QS).
P1100850

system/smbus.c
x86fr: Don't forget this one ;)

debrouxl: Indeed, I forgot about that one :)
However, I can now see that the binary happens to be smaller (simpler code) without the comparison against 0xFFFF... so let's keep this classic array size construct ?

x86fr:
Ready to merge?

debrouxl:
For good measure, I wanted to test this version of the commit on SB800 (especially, since I've modified the PIIX4 path, though the change is a no-op on everything but ServerWorks CSB5), MCP61 and ICH10 first, but not before tonight or even tomorrow evening. If you don't think that it's useful, or can beat me to it, go ahead :)

I too had the urge to fuse AMD and Hygon, but I resisted it because of the older, unimplemented AMD chipsets, and hoped that the compiler would optimize the code.

x86fr:
Alright. Doing some quick tests then squash & merge

debrouxl:
Boo, you removed the commented Linux driver names, which still serve informational purposes :)

x86fr:
The dev who will add support can feel free to add any source he wants. Personally, i found the I2C drivers from free/openBSD much more easier to read than the drivers from linux.

x86fr:
True, but it's not really an issue because copy-pasting code is almost impossible. All we need here is the registers description and the way to access it ... and the real source of these information is often a datasheet under NDA.

debrouxl:
Testing on my RS904A shows that 1002:4385 rev 3d support is broken in memtest86+ after this, whereas `decode-dimms` can still access the SPD data, even under Linux 5.18. I'll have to test which false branch is taken, but maybe not before the week-end.

BTW: after a couple quiet years, there's recently been a set of changes in the Linux PIIX4 driver by Terry Bowman @ AMD, for improving AMD platform support. https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/drivers/i2c/busses/i2c-piix4.c .

debrouxl:
Fixed by 63a07258fd7488c1cdc48bf84d52361a65246e0f for RS904A, and certainly others.

PR #46, opened 2022-04-17
Added a new usb_debug flag, slighly changed startup timing strategy

x86fr:
Hi Martin, can you please check if these small changes are OK for you?

Here are the changes:

- `pause_at_start` is now unrelated to USB detection and only used for the initial configuration
- `pause_at_start` has been reduced from 5 to 3 seconds
- A new flag called `pause_usb_debug` (disabled by default) will force an infinite delay after USB detection (until keypress, as previously)
- In default mode (`pause_usb_debug` = false), if an USB keyboard is found, the USB detection routine will immediately continue to the initial `pause_at_start`. If no USB keyboard is detected, a 10 seconds timer will begin before continue. That will allow users with a keyboard issue to take a screenshot while allowing the test to continue without waiting for an impossible keypress.

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -96,6 +96,7 @@ bool            enable_sm          = true;
 bool            enable_bench       = true;
 
 bool            pause_at_start     = true;
+bool            pause_usb_debug    = false;
 
 power_save_t    power_save         = POWER_SAVE_HIGH;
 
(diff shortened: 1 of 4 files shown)


martinwhitaker:
I was going to add USB_DEBUG to usb_init_options_t rather than create another flag.
I think we should only delay when no USB keyboard is detected if keyboard=usb is specified on the command line (i.e. when legacy keyboards aren't enabled).
I can do the changes if you want.

x86fr:
Sounds good. :) Post here when it's done or just merge this PR.

martinwhitaker:
Done.

Issue #49, opened 2022-04-21
Crash on start / Insufficient free space in range - Cannot relocate program.


danielstr007:
![This is an image](https://user-images.githubusercontent.com/102787488/164422395-f32b34ea-78cd-4c21-b9cf-8ab80317b7b8.png)

Current release crash directly after start

With older release from 07.04. everything running fine:
![image2022-4-12_14-12-5](https://user-images.githubusercontent.com/102787488/164422890-7579a228-784e-4315-a487-c2809897d49f.png)

Any ideas?

debrouxl:
Nice computer you have there :)

What's the size of the respective memtest86+ binaries from the previous version and this one ? I'd venture that the binary's size increased slightly, and .text + .data + .rodata + .bss do no longer fit in the 100-665c7 range.

danielstr007:
Hey,
thanks yes, this is one of our larger systems 

I could only quick check the complied bin files, size increased - so this looks like the problem?

Current:
-rwxr-xr-x 1 145792 Apr 21 08:51 memtest-6beta/memtest86plus-main/build64/memtest.bin
-rw-r--r-- 1 144256 Apr 21 08:51 memtest-6beta/memtest86plus-main/build64/memtest_shared.bin
OLD:
-rwxr-xr-x 1 133216 Apr  7 10:54 memtest-6beta_old/memtest86plus-main/build64/memtest.bin
-rw-r--r-- 1 131680 Apr  7 10:54 memtest-6beta_old/memtest86plus-main/build64/memtest_shared.bin

x86fr:
can you try to disable some CPU/Core? The root of the issue probably comes from the size of the AP structures and stack (394 kB total!), more than the size of the binary by itself.

danielstr007:
> can you try to disable some CPU/Core? The root of the issue probably comes from the size of the AP structures and stack (394 kB total!), more than the size of the binary by itself.

Will try to disable all cores to 1 tomorrow, but in the older release it worked? 
So, it looks like there was a code change?

x86fr:
Yes, but I think you were already extremely close to the limit. The tiny 12KB added code took it over the line.
Disabling only half of the cores (or even 1/4) should be enough.

Anyway, this is an issue we have to solve to claim support for 256 threads or even more.

danielstr007:
[...]

PR #54, opened 2022-04-24
Rework memory mapping to allow for larger program size

martinwhitaker:
Previously memtest86+ alternated between loading the program above 1MB & testing below 640K and loading the program below 640K & testing above 1MB. As the code size increases (and if we want to support more than 256 cores) it becomes increasingly unlikely the program will fit in the free memory below 640K (as already seen in issue #49). So this pull request reworks the memory management to set the low/high boundary at 4MB (which can easily be increased if needed in the future).

I have tested this on a couple of machines, but it would be good to get some wider testing before merging.

We should still merge Lionel's work to reduce the memory footprint - that is also beneficial.

### DIFF
--- a/app/main.c
+++ b/app/main.c
@@ -79,7 +79,6 @@ static bool             rerun_test = false;
 
 static bool             dummy_run  = false;
 
-static int              window_num   = 0;
 static uintptr_t        window_start = 0;
 static uintptr_t        window_end   = 0;
 
@@ -109,6 +108,8 @@ int         vm_map_size = 0;
 
 int         pass_num = 0;
 int         test_num = 0;
+
+int         window_num = 0;
 
 bool        restart = false;
 bool        bail    = false;
--- a/app/main.c
+++ b/app/main.c
@@ -58,6 +58,10 @@
 #define TEST_INTERRUPT      0
 #endif
 
+#define LOW_LOAD_LIMIT      SIZE_C(4,MB)  // must be a multiple of the page size
+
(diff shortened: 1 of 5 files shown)


x86fr:
Looks fine to me. The only "drawback" seems to be the minimum RAM available now fixed at ~8 MB but all P5 came with at least 16MB so it's really not an issue.

I will ask the betatester team to test this PR specifically next week, then merge on ~Friday for the first beta 1 binary release.

About optimization on smbus.c, I'll do that soon (I'm working on the new website and the windows installer right now) but I plan to remove a lot of unused strings, redundant calls/pointers and rework the whole code. The initial code was built in 2005 with a structure able to accommodate complex hardware evolution, but that never happened (modern controller still use the same SPD access method than 20 years ago).

debrouxl:
So far, this patch does what it's supposed to on 4 different computers: MS-6712, MS-6398E, K7S5A, (HP xw6600 workstation): the low memory is now 0-4 MB instead of 0-1024 KB. Tests on P5QC and VC820 to follow, but they're not expected to fail.

debrouxl:
Not directly related to this PR, but since we're mentioning this here: BTW, Sam, does the optimization you have in mind leave the door open to third-party contribution of the implementation of old(er) AMD, NVIDIA, SIS, VIA SMBus controllers, which are supported by Linux and/or some DOS/Windows hardware information tools ?

x86fr:
No issue to date. Will merge tonight.

debrouxl:
It looks like that upon the squash-merge, Github produced crappy information for both the author name and the committer name. Boo, bad Github. Oh well.

PR #62, opened 2022-05-09
Add AP Enumeration to distinguish P-Core from E-Core on Hybrid CPUs

x86fr:
(WIP)

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -1,11 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0
 // Copyright (C) 2020-2022 Martin Whitaker.
+// Copyright (C) 2004-2022 Sam Demeulemeester.
 //
 // Derived from memtest86+ config.c:
 //
-// MemTest86+ V5.00 Specific code (GPL V2.0)
-// By Samuel DEMEULEMEESTER, sdemeule@memtest.org
-// http://www.x86-secret.com - http://www.memtest.org
 // ----------------------------------------------------
 // config.c - MemTest-86  Version 3.4
 //
@@ -19,6 +17,7 @@
 #include "bootparams.h"
 
 #include "cpuinfo.h"
+#include "cpuid.h"
 #include "hwctrl.h"
 #include "keyboard.h"
 #include "memsize.h"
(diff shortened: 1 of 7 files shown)


x86fr:
@martinwhitaker> I have a chicken & egg situation here. I'm trying to add an option to disable E-Cores. To distinguish E-Cores from P-Cores, I must run cpuid on all cores. To avoid enumerating APs too early, I use the `main()` loop. That works and I'm able to populate a `hybrid_core_type[]` array with the type of each core. I then re-call `display_topology()` after the last AP started to update the topology on screen. Seems OK. But I'm now too far on the code to disable these cores (they're already initialized and running). And I need them running to run cpuid. So I see only three options, both will change significantly the flow : 

1.  Enumerating APs much earlier (just after `smp_init`) and change the early init loop on main()
2.  Stop the APs we don't want  (= E-Cores) after they have been started (if possible).
3.  Keep the APs in running state but skip the ones we don't want before starting each tests.

What's your thought?

app/config.c
```diff
@@ -87,6 +86,9 @@ error_mode_t    error_mode = ERROR_MODE_NONE;
 
 cpu_state_t     cpu_state[MAX_CPUS];
 
+core_type_t     hybrid_core_type[MAX_CPUS];
```
debrouxl, line 89: This - and cpu_state - ought to be arrays of uint8_t, so as not to waste 3 * MAX_CPU bytes. Last time I checked the C standard, the C++11+ typed enums hadn't been imported to C, though :'(

martinwhitaker: Last time I checked, GCC optimised the storage size to be a single byte per element.

debrouxl: So it does, indeed. TIL.
I'm not fully happy with reliance on such kinds of special optimizing compiler behaviour, but as long as it works, after all...

martinwhitaker:
@x86fr, I think it is quite important to keep the ability to disable individual cores before trying to start them. So I prefer options 2 or 3.

x86fr:
Ok. If E-Core are excluded (initial setup only), I will hlt them and skip them.

x86fr:
That's the easiest way I found to remove the E-Cores from the test list. Tests are now significantly faster on Hybrid CPUs (Memory accesses from P-Core are much faster than from E-Core).

I tried to hlt these cores with

```
    if (hybrid_core_type[my_cpu] == CORE_ECORE && exclude_ecores) {
        cpu_state[my_cpu] = CPU_STATE_DISABLED;
        num_enabled_cpus--;
        num_available_cpus--;
         __asm__ volatile ("hlt");
    }
```

But the main loop freeze somewhere later, so something is missing.

x86fr:
Option 3 implemented and working. Option 2 would be better, but requires more work. 
I'm asking the beta team to do some test then merge

PR #76, opened 2022-05-19
Save space by inlining functions and getting rid of library functions which become unused

debrouxl:
Together, these three commits save over 500 bytes on the .text size in a x86_64 build, and over 400 bytes on a x86 build:
* passing arguments and branching to a subroutine has a cost and can worsen register allocation;
* the `rep stos*` and `rep movs*` x86 instructions make for small inline `memcpy()` / `memset()`. Every occurrence of `memset()` switched to `__builtin_memset()` either reduced the size, or did not increase it.


martinwhitaker:
I would prefer to investigate whether we can remove the `-fno-builtin` option and let the compiler do this automatically.

debrouxl:
Well... On my side, with GCC 11,
```
index d04a2b0..1c9cfcf 100644
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -9,7 +9,7 @@ else
   GIT_AVAILABLE = true
 endif
 
-CFLAGS = -std=c11 -Wall -Wextra -Wshadow -m32 -march=i586 -fpic -fno-builtin \
+CFLAGS = -std=c11 -Wall -Wextra -Wshadow -m32 -march=i586 -fpic \
          -ffreestanding -fomit-frame-pointer -fno-stack-protector
 
 INC_DIRS = -I../boot -I../system -I../lib -I../tests -I../app -Iapp
index 564329c..3e615a8 100644
--- a/build64/Makefile
+++ b/build64/Makefile
@@ -10,7 +10,7 @@ else
 endif
 
 CFLAGS = -std=c11 -Wall -Wextra -Wshadow -m64 -march=x86-64 -mno-mmx -mno-sse -mno-sse2 \
-         -fpic -fno-builtin -ffreestanding -fomit-frame-pointer -fno-stack-protector
+         -fpic -ffreestanding -fomit-frame-pointer -fno-stack-protector
 
 INC_DIRS = -I../boot -I../system -I../lib -I../tests -I../app -Iapp
 
```
yields zero size change on all sections :)

```
$ size memtest_shared
 109984   29832  292480  432296   698a8 memtest_shared
```

whereas with this pull request:
```
$ size memtest_shared
 109437   29832  292480  431749   69685 memtest_shared
```

x86fr:
> yields zero size change on all sections :)

Isn't it precisely because we don't use builtin functions?

martinwhitaker:
The `-ffreestanding` option also disables automatic use of the built-in functions. And removing that is problematic because I have used a non-standard `printf` prototype.

Still, I think it would be better to have
````
#define memcpy(d, s, n) __builtin_memcpy((d), (s), (n))
#define memset(s, c, n) __builtin_memset((s), (c), (n))
````
in `string.h` rather than change every call.

I also prefer using `#if 0` to disable sections of code rather than comment markers.

**Comment by debrouxl** [comment]
Both changes done.

File: lib/string.c [hunk 1]
```diff
@@ -43,6 +43,7 @@ int memcmp(const void *s1, const void *s2, size_t n)
     return 0;
 }
 
+#if 0
 void *memcpy(void *dest, const void *src, size_t n)
```
x86fr, line 47: If we don't use our own function anymore, it's probably better to remove it completely.

system/ehci.c
```diff
@@ -337,12 +335,12 @@ static void build_ehci_qtd(ehci_qtd_t *this_qtd, const ehci_qtd_t *final_qtd, ui
     this_qtd->control       = EHCI_QTD_CPAGE(0) | EHCI_QTD_CERR(3) | control;
     this_qtd->data_length   = dt | (length & 0x7fff);
     this_qtd->buffer_ptr[0] = (uintptr_t)buffer;
+
+    memset((char *)this_qtd + offsetof(ehci_qtd_t, buffer_ptr[1]), 0, sizeof(ehci_qtd_t) - offsetof(ehci_qtd_t, buffer_ptr[1]));
```
x86fr, line 339: I don't understand that change in ehci.c, but if it's OK for @martinwhitaker, that's fine.

debrouxl: The previous fields were already initialized one at a time and in the right order, after the call to `memset()` located at the beginning of the function. This `memset()` was therefore effectively doing superfluous work.

lib/string.c
debrouxl: Fine with me :)
I'll do it later.

system/ehci.c
martinwhitaker: Yes, but the original code is much easier to read and to see is correct.
Does this actually reduce the code size?

debrouxl: Sure, in fact ISTR that the out of line full memset -> reduced __builtin_memset change in that function even makes for the largest individual size optimization in that `memset()`-related commit :)

martinwhitaker: I just tested in a 64-bit build, and the changes in ehci.c (separate from the changes in string.{c,h} reduce the program size by 16 bytes, although most of that is likely due to alignment - disassembly shows that `build_ehci_qtd` is 2 bytes smaller and `build_ehci_qhd` is the same size. I'd rather have the more readable code.

debrouxl: Here on a 64-bit build, with GCC 11 (Debian sid amd64):
* before the first commit containing the memset -> __builtin_memset change: `0000000000002488 l     F .text  0000000000000070 build_ehci_qtd`
* after the first commit, i.e memset as. __builtin_memset + memset(x + offsetof, 0, size - offsetof) at the end of the function: `0000000000002482 l     F .text  000000000000004a build_ehci_qtd`
* memset as. __builtin_memset + memset(x, 0, size) at the beginning of the function: `0000000000002482 l     F .text  000000000000004f build_ehci_qtd`
I'll take the third version nevertheless.

debrouxl:
Patches refreshed.
```
$ file memtest_shared
memtest_shared: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped

$ size memtest_shared
 107563   28872  292672  429107   68c33 memtest_shared (original)
 107332   28872  292672  428876   68b4c memtest_shared (patch 1)
 107265   28872  292672  428809   68b09 memtest_shared (patch 2)
 107032   28872  292672  428576   68a20 memtest_shared (patch 3)
```

x86fr:
I think we should really keep the code readable and easy to understand, even at the expense of a few "lost" kilobytes in total code size. While it's desirable to keep the code reasonably optimized and as small as possible, it's not a problem to increase the code size by even a large margin if needed. The size of the commercial Memtest is well over 1 MB and I don't think anybody ever complained.

In the long term, the less obfuscate the code is, the better. Having an highly skilled developer like Martin willing to spend time on an open-source project memory testing tool is awesome, but the vast majority of the devs willing to help are less skilled (including myself). Hopefully, Martin's code is still understandable while being quite complex on some part. But I think it took 15 years to find someone (Martin) able to understand some part of the bootloader written by Eric Biederman. 

If Martin is not available in the future and the code complexity increase, I (or someone else) will probably need 10x the time to understand and solve an issue (and probably fix it with a quick and dirty patch). A few hundred bytes saved aren't worth it.

PS: Of course, as long as optimizations don't impact readability, they are welcome.

01e3:
Regarding printf / printc / vprintf / etc - I wonder if it would make sense to simply rename them, given they have different syntax than the official ones.

Some ideas: printfxy, xyprintf (or versions with underscores).

As a bonus, it would also later help to add unit tests to the code.

If such a change would be desired / acceptable, I'm happy to contribute a patch.

martinwhitaker:
> While it's desirable to keep the code reasonably optimized and as small as possible, it's not a problem to increase the code size by even a large margin if needed. The size of the commercial Memtest is well over 1 MB and I don't think anybody ever complained.

Code size is important to us because we relocate the code twice per test (to allow us to test the memory the code occupies). The commercial MemTest86 has been rewritten as a UEFI application, so I suspect it no longer does that.

x86fr:
> Code size is important to us because we relocate the code twice per test (to allow us to test the memory the code occupies). The commercial MemTest86 has been rewritten as a UEFI application, so I suspect it no longer does that.

I still don't get it. Relocating a few hundreds kB with memory bandwidth in the GB/s range only need a few hundreds µsec to complete (some milliseconds at worst on very old architecture). In a total pass, the overall time needed for all relocations on a modern platform is probably less than 0.1s. So does it really matters at all?

martinwhitaker:
> I still don't get it. Relocating a few hundreds kB with memory bandwidth in the GB/s range only need a few hundreds µsec to complete (some milliseconds at worst on very old architecture). In a total pass, the overall time needed for all relocations on a modern platform is probably less than 0.1s. So does it really matters at all?

You're probably right. My first computer was a homebrew Z80 system with 1KB ROM and 256 bytes of RAM. Old habits die hard!

PR #77, opened 2022-05-19
Add a way to handle hardware quirks at init or later in the code.

x86fr:
Add a way to handle hardware quirks at init or later in the code.  
Add ASUS TUSL2-C ASB100 Mux as (working) example


debrouxl:
I like it :)
If it were just me, I'd use `quirk_t quirks[1]`, so that no refactor is necessary *if* a MB proves to require multiple quirks, but that's probably a big "if", so carry on ;)

PR #85, opened 2022-05-22
Optimize screen & serial output handling.

01e3:
- Enable VGA/FB to output box drawing characters while maintaining VT100
   character set for serial. Shorten and simplify the screen setup code.

 - Track the background color to decide if the serial output needs to be
   inverted. Remove no longer needed logic for known areas of the screen that
   need to be inverted. As a bonus - popup menu can now be also inverted on
   serial.

 - Reduce the amount of data sent to serial by using CR+LF when possible
   instead of always relying on absolute positioning. Add tty_print()
   for positioning the cursor, remove no longer needed tty_print().

 - Remove no longer needed "LF -> LF+CR" logic from serial_echo_print().

```
Before (gcc-11.3.0-x86_64):
      text       data        bss      total filename
       929        357         64       1350 system/serial.o
      3517       1356         54       4927 app/display.o

After (gcc-11.3.0-x86_64):
      text       data        bss      total filename
       907        336         64       1307 system/serial.o
      3442       1242         54       4738 app/display.o
```
memtest-serial

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -80,42 +80,43 @@ void display_init(void)
 
     clear_screen();
 
+    /* The commented horizontal lines provide visual cue for where and how
+     * they will appear on the screen. They are drawn down below using
+     * Extended ASCII characters.
+     */
+
     set_foreground_colour(BLACK);
     set_background_colour(WHITE);
     clear_screen_region(0, 0, 0, 27);
-    prints(0, 0, "     Memtest86+ v6.00b1");
+    prints(0, 5,      "Memtest86+ v6.00b1");
     set_foreground_colour(RED);
     printc(0, 14, '+');
     set_foreground_colour(WHITE);
     set_background_colour(BLUE);
-    prints(0,28,                             "| ");
-    prints(1, 0, "CLK/Temp:   N/A             | Pass   % ");
-    prints(2, 0, "L1 Cache:   N/A             | Test   % ");
-    prints(3, 0, "L2 Cache:   N/A             | Test #   ");
-    prints(4, 0, "L3 Cache:   N/A             | Testing: ");
-    prints(5, 0, "Memory  :   N/A             | Pattern: ");
-    prints(6, 0, "--------------------------------------------------------------------------------");
-    prints(7, 0, "CPU:                      SMP: N/A        | Time:           Status: Init. ");
-    prints(8, 0, "Using:                                    | Pass:           Errors: ");
-    prints(9, 0, "--------------------------------------------------------------------------------");
+    prints(1, 0, "CLK/Temp:   N/A             | Pass   %");
+    prints(2, 0, "L1 Cache:   N/A             | Test   %");
+    prints(3, 0, "L2 Cache:   N/A             | Test #");
+    prints(4, 0, "L3 Cache:   N/A             | Testing:");
+    prints(5, 0, "Memory  :   N/A             | Pattern:");
+//  prints(6, 0, "--------------------------------------------------------------------------------");
+    prints(7, 0, "CPU:                      SMP: N/A        | Time:           Status: Init.");
+    prints(8, 0, "Using:                                    | Pass:           Errors:");
+//  prints(9, 0, "--------------------------------------------------------------------------------");
 
-    // Redraw lines using box drawing characters.
-    // Disable if TTY is enabled to avoid VT100 char replacements
-    if (!enable_tty) {
-        for (int i = 0;i < 80; i++) {
-            print_char(6, i, 0xc4);
-            print_char(9, i, 0xc4);
-        }
-        for (int i = 0; i < 6; i++) {
-            print_char(i, 28, 0xb3);
-        }
-        for (int i = 7; i < 10; i++) {
-            print_char(i, 42, 0xb3);
-        }
-        print_char(6, 28, 0xc1);
-        print_char(6, 42, 0xc2);
-        print_char(9, 42, 0xc1);
+    for (int i = 0;i < 80; i++) {
+        print_char(6, i, 0xc4);
+        print_char(9, i, 0xc4);
     }
+    for (int i = 0; i < 6; i++) {
+        print_char(i, 28, 0xb3);
+    }
+    for (int i = 7; i < 10; i++) {
+        print_char(i, 42, 0xb3);
+    }
+
+    print_char(6, 28, 0xc1);
+    print_char(6, 42, 0xc2);
+    print_char(9, 42, 0xc1);
 
     set_foreground_colour(BLUE);
     set_background_colour(WHITE);
--- a/system/serial.c
+++ b/system/serial.c
@@ -67,33 +67,24 @@ void serial_echo_print(const char *p)
     if (!port->enable) {
         return;
     }
+
     /* Now, do each character */
     while (*p) {
-        serial_wait_for_xmit(port);
-
         /* Send the character out. */
-        serial_write_reg(port, UART_TX, *p);
-        if (*p==10) {
-            serial_wait_for_xmit(port);
-            serial_write_reg(port, UART_TX, 13);
-        }
-        p++;
+        serial_wait_for_xmit(port);
+        serial_write_reg(port, UART_TX, *p++);
     }
 }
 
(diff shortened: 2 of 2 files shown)


x86fr:
Looks good. Many betatesters uses the TTY function. I just send them a build. 
If goes well, this PR is planned for merge before Beta 2 (next Friday)

app/display.c
```diff
@@ -83,39 +83,33 @@ void display_init(void)
     set_foreground_colour(BLACK);
     set_background_colour(WHITE);
     clear_screen_region(0, 0, 0, 27);
-    prints(0, 0, "     Memtest86+ v6.00b1");
+    prints(0, 5, "Memtest86+ v6.00b1");
```
martinwhitaker, line 86: Indent the string by 5 spaces so that we don't lose the visual cue for where this will appear on the screen.

app/display.c
```diff
@@ -83,39 +83,33 @@ void display_init(void)
...
     printc(0, 14, '+');
     set_foreground_colour(WHITE);
     set_background_colour(BLUE);
-    prints(0,28,                             "| ");
-    prints(1, 0, "CLK/Temp:   N/A             | Pass   % ");
-    prints(2, 0, "L1 Cache:   N/A             | Test   % ");
-    prints(3, 0, "L2 Cache:   N/A             | Test #   ");
-    prints(4, 0, "L3 Cache:   N/A             | Testing: ");
-    prints(5, 0, "Memory  :   N/A             | Pattern: ");
-    prints(6, 0, "--------------------------------------------------------------------------------");
```
martinwhitaker, line 97: Comment the line rather than removing it, again because it documents the screen layout.

app/display.c
```diff
@@ -83,39 +83,33 @@ void display_init(void)
...
-    prints(7, 0, "CPU:                      SMP: N/A        | Time:           Status: Init. ");
-    prints(8, 0, "Using:                                    | Pass:           Errors: ");
-    prints(9, 0, "--------------------------------------------------------------------------------");
```
martinwhitaker, line 100: And again.

app/display.c
01e3: Good idea, done.

app/display.c
01e3: Done.

app/display.c
01e3: Done

x86fr:
Many betatesters reported screen corruption with this PR. It gave me a headache to understand what's wrong because your patch is OK and everything is fine on my test platforms when I test it. I even hooked up a logic analyzer without seeing any issue.

Finally, I dug through my emails and found a decade old discussion I had with a betatester when I switched from a TTY code that looked like yours (using \r\n) to the actual code (with x/y sent on each line).

The issue come from the fact that standard VGA used by Memtest86+ is 80x25 and almost all VT100 terminals (PuTTY, KiTTY, many IPMI and serial-over-terminal consoles) are configured by default in 80x24. Sometimes there is no option to display 80x25 (or users are too lazy to configure it).

On the current code, when trying to display at line 25 on a VT100 80x24terminal , it just overwrite the last line (24) and works fine. 
By using \r\n, your patch makes the screen scroll, resulting to lines being shifted by 1. The subsequent refreshes are written in the wrong place, resulting to a display that looks like that :

```
CLK/Temp: 2079MHz           | Pass   %
L1 Cache:   64KB  11.9 GB/s | Test  1%
L2 Cache:  512KB  3.85 GB/s | Test 87% ##################################
L3 Cache:   N/A             | Test #3  [Moving inversions, 1s & 0s]
Memory  : 1023MB   713 MB/s | Patting: 4MB - 1023MB [1019MB of 1023MB]     [PAE]
----------------------------------ern: 0xffffffff
CPU: 1 Cores 1 Threads    SMP: Disabled   | Time:           Status: Testing  -
CPU: 1 Cores 1 Threads    SMP: Disabled   | Time:  0:00:18  Status: Testing  -
Using: All Cores                          | Pass:  0        Errors: 0













                          ASUSTeK Computer INC. A7N8X-LA
 <ESC> Exit  <F1> Configuration  <Space> Scroll Lock            6.00.4a12dda.x3
```

I tried to bypass the display of line 23 (usually blank) or to send the coordinate when displaying line 24&25, but it stills fail on some IPMI consoles (and maybe elsewhere, but there is not many testers at work these days). 

I'm not sure if there is a way to solve this without always relying on absolute positioning (or adding some nasty tricks for the last lines).

01e3:
Heh, thanks for very detailed testing!

Indeed, I have not tested 80x24.

We can switch to absolute positioning, the penalty is not that significant, but I wonder what we would like to do in this case given we have 80x25 VGA that needs to map to the terminal that may be 80x24?

I will start with checking how the current code handles this.

01e3:
The current code overrides line 24 (platform info) with line 25 - menu & version.

I can do the same, or we can think about some other options. Implementing something is easy once we have a solid PRD. ;)

What I should probably ask ask - what is the minimal terminal size we would like to support? 80x24?

x86fr:
80x24 minimum seems reasonable. The current code overwrite line 24 with line 25, but ONLY when the terminal is set to 80x24. When it's set to 80x25, no lines are lost.

I already tried to send absolute coordinates for line 24&25 on this PR, but it still fail sometimes :-/

01e3:
Something like fixes 80x24 for me - I'm using putty:
```

-        if (start_col || row == start_row) {
+        if (start_col || row == start_row || row >= 24) {

```

If this is what we want, I can update the comment to explain the trap here of 80x24 and submit an updated patch.

However, "it still fail sometimes" is concerning and suggests different problem. Could you give me some hints how I can reproduce this?

x86fr:
Everyone is OOO today :-/ Will ask them next week.

x86fr:
OK, the issue is much harder to reproduce but still common (5-10% of all computers tested on 2 specific warehouse).  They all use Digi Portserver TS or Connectport (RS232-Ethernet Gateway). We were finally able to reproduce the issue by forcing baudrate to 300 bps and unplug/replug the RJ45 when it starts to write on screen. For some reasons, when an Ethernet packet is lost, the RS232 Gateway send a CR-LF after some millisecond (Like in "Hey! Something goes wrong, let's start a new line!"). Without absolute positioning on each line, it sometimes scramble the screen. 

The patch at https://github.com/memtest86plus/memtest86plus/pull/85#issuecomment-1139649829 solves the issue, so, even if it's not a Memtest86+ issue, I think it's better to add it.

01e3:
Thank you so much for all the details! I did try reproducing this problem myself with Serial Console over IPMI / DRAC, but yes, I did not try nor have Digi Portserver to test it against. :/

I will update the patch and add a detailed comment regarding this problem.

Also, as a secondary defense mechanism for potential screen corruption due to "spurious CR-LF" ans scroll, I think we may always want to position the cursor at the [1,1] (VT100) at the end of the end of tty_send_region() by sending "<ESC>[H". WDYT?

01e3:
BTW: are we talking about something like https://www.ebay.com/itm/254185779131?

x86fr:
They're mainly use Portserver TS8 and TS16 (https://ftp1.digi.com/support/documentation/prd_ts_portserverts816.pdf).

About sending `<ESC>[H` at the end of `tty_send_region()`, it's a good idea, but let's wait if it's really needed first.

Can you push the change at https://github.com/memtest86plus/memtest86plus/pull/85#issuecomment-1139649829 so we can merge this PR?

x86fr:
No issues reported since with https://github.com/memtest86plus/memtest86plus/pull/85#issuecomment-1139649829 patch.

Someone to push the patch to the PR?

01e3:
I apologize for the delay and thank you so much for more testing! I have been recently traveling internationally, I should be able to take care of this within ~2 days, assuming United won't cancel my flight again this time. :/

01e3:
Once again sorry for the delay and thanks for merging. International (intercontinental) travel with kids is not easy. As a bonus, I can now test memtest on a new set of [ancient] HW. ;)

Seems like we are all done here?

x86fr:
Please post screenshots of working systems here : #20 ;)

PR #88, opened 2022-05-24
Enable support for NV MCP61 (10de:03eb) SMBus controller

debrouxl:
P1100828

The computer has survived several seconds of memtest86+ and a reboot :)

Off-topic, but since it shows on the picture: sharp-eyed readers might notice the ~2200 MHz CPU frequency. This isn't a bug in memtest86+: I'm still searching for the maximum frequency for stable operation (as in, absence of thermal shutdown at 100% CPU...), which is much lower when the case is closed. I wiped the old thermal paste and cleaned up the surfaces with IPA, laid brand-new MX-4 thermal paste - enough to spread over the whole surface of the processor's cover, not way too much to put thermal paste all over the sides of the socket - and both the CPU fan and system fan are forced at max speed > 3000 RPM. At thermal shutdown, the heatsink isn't too hot to touch with bare fingers.


x86fr:
Was working on it and missed this PR. Merged with e1508a02be3869354143f71d92eaf0a276f70805

PR #90, opened 2022-05-25
Add github action

bensuperpc:
Add github action:
- Add build with ubuntu 18.04 and 20.04
- Add build with GCC and Clang
- Add dependabot
- Add stalebot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every weekday
+      interval: "daily"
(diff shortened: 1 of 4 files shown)


Fantu:
good addition
about the build I suggest build also on newer system (ubuntu 18.04 and 20.04 are now quite "old"), for example debian unstable to check if build also on latest version of gcc (or others)

bensuperpc:
> good addition about the build I suggest build also on newer system (ubuntu 18.04 and 20.04 are now quite "old"), for example debian unstable to check if build also on latest version of gcc (or others)

This is still in [beta](https://github.com/actions/virtual-environments/issues/5490), i added ubuntu 22.04 ^^

debrouxl:
I'm no Github action expert, but this looks reasonable :)

x86fr:
Let's try it!

x86fr:
Add github action:
- Add build with ubuntu 18.04 and 20.04
- Add build with GCC and Clang
- Add dependabot
- Add stalebot

Issue #40, opened 2022-04-11
PIT Counter broken on Framework Laptop


chirs241097:
On Gentoo, the `iso` target fails because `sys-fs/dosfstools` does not provide `/sbin/mkdosfs`. Instead it uses `/usr/sbin/mkfs.fat`. Simply changing the path to the correct executable would fix it, and produces an image that seems to work on a ThinkPad X60 Tablet (legacy BIOS) and ThinkPad X1 yoga 1st gen (UEFI w/o CSM).

However the said image does not work on a Framework Laptop. After loaded it just gets stuck in a blank screen. I've also tried booting with `smp` or `nosm` (as well as both), none of which has any effect on the result. The machine has an i7-1185G7 processor and 64 GB of DDR4 RAM (in 2*32 GB configuration). I'm not familiar with the process memtest86+ should be debugged, but if provided with instructions I would happily report the results.

x86fr:
Can you try the 32bit build, directly started from the UEFI shell. 

Also a copy of `memmap` and `acpiview` from the UEFI shell can help.

martinwhitaker:
The 32-bit build only supports 32-bit UEFI systems, and I'll be surprised if the Framework laptop is that.

Please try changing the line `#define DEBUG 0` near the top of `boot/efisetup.c` to `#define DEBUG 1`. Does that cause any diagnostic messages to be displayed?

martinwhitaker:
OK, that shows the Framework laptop hasn't provided a usable address for the graphics frame buffer. Unless I've misinterpreted the UEFI spec., that's a firmware bug. Without a usable frame buffer, Memtest86+ can't display anything.

chirs241097:
Thanks. I'll report this to the Framework team.

martinwhitaker:
Could you retest with this patch, just to make sure it isn't a 64-bit address aligned on a 4GB boundary:
````
--- a/boot/efisetup.c
+++ b/boot/efisetup.c
@@ -366,7 +366,7 @@ static efi_status_t set_screen_info_from_gop(screen_info_t *si, efi_handle_t *ha
 
 #if DEBUG
     print_string("FB base   : ");
-    print_hex(si->lfb_base);
+    print_hex((uintptr_t)lfb_base);
     print_string("\n");
     print_string("FB size   : ");
     print_dec(si->lfb_width);
````

**Comment by chirs241097** [comment]
After applying the patch, the message printed at boot changed to:
```
BGR32 mode
[...]

PR #102, opened 2022-06-12
Add ACPI Timer as the primary TSC correction source and PIT Timer as fallback

x86fr:
- Add function to parse additional APIC tables (FADT added and option for HPET, SLIT/SRAT, etc.)
- Split APIC code from SMP into `apic.c` 
- Add timers code to a separate file `timers.c`.
- Change init order to configure timers before USB init.

Here's the story for this PR. To ensure accurate delays and clock/bandwidth measurements, we need a reliable clock source. The easier one is the TSC, but the TSC is not synchronized with the CPU anymore. It is often invariant, even if the CPU clock speed change (by a vast margin on mobile SKU). So we need an accurate clock source to add a "calibration factor" to the TSC. We got it from the old PIT timer (~1.2MHz). 

Unfortunately, some laptop manufacturers don't configure their firmware correctly and enable the PIT Clock Gating by default at boot time. We had only one report (#40), so we didn't care. Recently, I started to work on some new laptops that will be ready for Black Friday later this year. They comes with a development AMI BIOS where this option was enabled by default. 

IMG_2563

AMI told us that this should be disabled for production, but it seems that some new actors on the market lacks the knowledge to understand exactly what to enable and what to disable, so they keep the development BIOS settings on their production BIOS. I suspect this also the case here : https://github.com/memtest86plus/memtest86plus/discussions/93#discussioncomment-2875111

Checking if the PIT Timer is in clock gating mode and potentially re-enable it is not straightforward, because it's a platform-specific setting. So it's better to use another timer source. The ACPI Timer (ticking at a fixed 3.6 MHz) is easy to use. It provides accurate measurements to set the `clks_per_msec` correction factor. If needed later, we can also use the HPET or APIC Timers.

Let me know if you find something wrong.

### DIFF
--- a/system/acpi.c
+++ b/system/acpi.c
@@ -37,7 +37,7 @@
 #define CPEPSignature   ('C' | ('P' << 8) | ('E' << 16) | ('P' << 24)) // Corrected Platform Error Polling Table
 #define HESTSignature   ('H' | ('E' << 8) | ('S' << 16) | ('T' << 24)) // Hardware Error Source Table
 
-#define SLITSignature   ('S' | ('L' << 8) | ('I' << 16) | ('T' << 24)) // System Locality System Information Table (NUMA)
+#define SLITSignature   ('S' | ('L' << 8) | ('I' << 16) | ('T' << 24)) // System Locality Information Table (NUMA)
 #define SRATSignature   ('S' | ('R' << 8) | ('A' << 16) | ('T' << 24)) // System Resource Affinity Table (NUMA)
 
 //------------------------------------------------------------------------------
@@ -326,20 +326,19 @@ int acpi_checksum(const void *data, int length)
 
 void acpi_init(void)
 {
-    // Find ACPI RDST Table Address
     acpi_config.rsdp_addr = find_rsdp();
 
-    // Find ACPI MADT Table Address
+    if (acpi_config.rsdp_addr == 0) {
+        return;
+    }
+
     acpi_config.madt_addr = find_acpi_table(MADTSignature);
 
-    // Find ACPI FADT Table Address
     acpi_config.fadt_addr = find_acpi_table(FADTSignature);
 
-    // Parse FADT
     if (acpi_config.fadt_addr) {
         parse_fadt(acpi_config.fadt_addr);
     }
 
-    // Find ACPI HPET Table Address
     acpi_config.hpet_addr = find_acpi_table(HPETSignature);
 }
--- a/system/acpi.c
+++ b/system/acpi.c
@@ -35,7 +35,7 @@
 #define EINJSignature   ('E' | ('I' << 8) | ('N' << 16) | ('J' << 24)) // Error Injection Table
 #define ERSTSignature   ('E' | ('R' << 8) | ('S' << 16) | ('T' << 24)) // Error Record Serialization Table
 #define CPEPSignature   ('C' | ('P' << 8) | ('E' << 16) | ('P' << 24)) // Corrected Platform Error Polling Table
-#define HESTSignature   ('H' | ('E' << 8) | ('S' << 16) | ('T' << 24)) //  Hardware Error Source Table
+#define HESTSignature   ('H' | ('E' << 8) | ('S' << 16) | ('T' << 24)) // Hardware Error Source Table
 
 #define SLITSignature   ('S' | ('L' << 8) | ('I' << 16) | ('T' << 24)) // System Locality System Information Table (NUMA)
 #define SRATSignature   ('S' | ('R' << 8) | ('A' << 16) | ('T' << 24)) // System Resource Affinity Table (NUMA)
@@ -43,6 +43,14 @@
 //------------------------------------------------------------------------------
 // Types
 //------------------------------------------------------------------------------
+
+typedef struct __attribute__ ((packed)) {
+    uint8_t     address_space;
+    uint8_t     bit_width;
+    uint8_t     bit_offset;
+    uint8_t     access_size;
+    uint64_t    address;
+} acpi_gen_addr_struct;
 
 typedef struct {
     char        signature[8];   // "RSD PTR "
@@ -81,10 +89,7 @@ static const efi_guid_t EFI_ACPI_2_RDSP_GUID = { 0x8868e871, 0xe4f1, 0x11d3, {0x
 
 const char *rsdp_source = "";
 
-uintptr_t rsdp_addr = 0;
-uintptr_t madt_addr = 0;
-uintptr_t fadt_addr = 0;
-uintptr_t hpet_addr = 0;
+acpi_t acpi_config = {0, 0, 0, 0, 0, 0, 0, 0, false};
 
 //------------------------------------------------------------------------------
 // Private Functions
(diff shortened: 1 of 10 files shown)


system/acpi.c
```diff
@@ -22,17 +22,23 @@
...
-#define HPETSignature   ('H' | ('P' << 8) | ('E' << 16) | ('T' << 24))
+#define HPETSignature   ('H' | ('P' << 8) | ('E' << 16) | ('T' << 24)) // High Precision Event Timer
 
-#define EINJSignature   ('E' | ('I' << 8) | ('N' << 16) | ('J' << 24))
+#define EINJSignature   ('E' | ('I' << 8) | ('N' << 16) | ('J' << 24)) // Error Injection Table
+#define ERSTSignature   ('E' | ('R' << 8) | ('S' << 16) | ('T' << 24)) // Error Record Serialization Table
+#define CPEPSignature   ('C' | ('P' << 8) | ('E' << 16) | ('P' << 24)) // Corrected Platform Error Polling Table
+#define HESTSignature   ('H' | ('E' << 8) | ('S' << 16) | ('T' << 24)) //  Hardware Error Source Table
+
+#define SLITSignature   ('S' | ('L' << 8) | ('I' << 16) | ('T' << 24)) // System Locality System Information Table (NUMA)
```
debrouxl, line 40: Excess "System" in the middle of the comment. Also, extra whitespace in line 38 :)

system/acpi.c
```diff
@@ -0,0 +1,345 @@
...
+        if (*(uint32_t *)rt != XSDTSignature) {
+            return 0;
+        }
+        rt = (rsdt_header_t *)map_region(rp->xsdt_addr, rt->length, true);
+        if (rt == NULL || acpi_checksum(rt, rt->length) != 0) {
+            return 0;
+        }
+        // Scan the XSDT for a pointer to the table we're looking for.
+        uint64_t *tab_ptr = (uint64_t *)((uint8_t *)rt + sizeof(rsdt_header_t));
+        uint64_t *tab_end = (uint64_t *)((uint8_t *)rt + rt->length);
```
debrouxl, line 230: What is a reasonable upper bound for rt->length ? Certainly much lower than UINT32_MAX or even 1 GB.

system/acpi.c
x86fr: Copy/parsed from https://wiki.osdev.org/RSDT. Will correct. 😁

system/acpi.c
```diff
@@ -0,0 +1,345 @@
...
+    } else {
+        rt = (rsdt_header_t *)map_region(rp->rsdt_addr, sizeof(rsdt_header_t), true);
+        if (rt == NULL) {
+            return 0;
+        }
+        // Validate the RSDT.
+        if (*(uint32_t *)rt != RSDTSignature) {
+            return 0;
+        }
+        rt = (rsdt_header_t *)map_region(rp->rsdt_addr, rt->length, true);
```
debrouxl, line 249: Ditto about an upper bound for rt->length.

system/acpi.c
x86fr: Probably, but that was set according to ACPI specs.

system/acpi.c
```diff
@@ -0,0 +1,345 @@
...
+static bool parse_fadt(uintptr_t fadt_addr)
+{
+    // FADT is a very big & complex table and we only need a few data.
+    // We use byte offset instead of a complete struct.
+
+    // FADT Header is identical to RSDP Header
+    rsdt_header_t *fadt = (rsdt_header_t *)fadt_addr;
+
+    // Validate FADT
+    if (fadt == NULL || acpi_checksum(fadt, fadt->length) != 0) {
```
debrouxl, line 279: Ditto about fadt->length.

x86fr: Still according to ACPI specs : https://uefi.org/specs/ACPI/6.4/05_ACPI_Software_Programming_Model/ACPI_Software_Programming_Model.html#fixed-acpi-description-table-fadt

system/acpi.c
```diff
@@ -0,0 +1,345 @@
...
+    // Find ACPI FADT Table Address
+    acpi_config.fadt_addr = find_acpi_table(FADTSignature);
+
+    // Parse FADT
+    if (acpi_config.fadt_addr) {
+        parse_fadt(acpi_config.fadt_addr);
+    }
+
+    // Find ACPI HPET Table Address
+    acpi_config.hpet_addr = find_acpi_table(HPETSignature);
```
debrouxl, line 344: AFAICS (I wondered if I'm failing at reading and grepping code & patch on a late Sunday evening, but I'm unable to find evidence of that), hpet_addr is only written to but not read back, so this should be commented out for now ?

system/timers.c
```diff
@@ -0,0 +1,94 @@
...
+
+            // Make sure we have a credible result
+            if (loops >= 10 && run_time >= 50000) {
+               clks_per_msec = run_time / 50;
+               return;
+            }
+        }
+    }
+
+    // Use PIT Timer to find TSC correction factor if APIC not available
```
debrouxl, line 64: An easy mistake to make - I sometimes do it on other acronyms as well - but "PIT Timer" is effectively "Programmable Interval Timer Timer", here, in the commit message and PR title :)

system/acpi.c
debrouxl: Well, when standards do not define safeguards, we'd better invent ours, as a measure against voluntarily or involuntarily specially crafted data, whose processing would consume excessive CPU time, or cause memory accesses beyond the currently mapped memory in some cases :)
In the absence of such checks, fuzzers ought to quickly destroy the parsing code...

system/acpi.c
```diff
@@ -0,0 +1,345 @@
...
+    uint8_t *ptr = (uint8_t *)data;
+    while (length--) {
+        sum += *ptr++;
+    }
+    return sum;
+}
+
+void acpi_init(void)
+{
+    // Find ACPI RDST Table Address
```
martinwhitaker, line 329: Comment should refer to RSDP, not RDST. But the comments are really unnecessary here - the code is clear enough on its own.

system/acpi.c
```diff
@@ -0,0 +1,345 @@
...
+    acpi_config.rsdp_addr = find_rsdp();
+
```
martinwhitaker, line 331: Should test for failure (zero address) here, otherwise the following calls to find_acpi_table will dereference an invalid pointer.

system/acpi.h
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef _ACPI_H_
+#define _ACPI_H_
+/**
+ * \file
+ *
+ * Provides support for multi-threaded operation.
```
martinwhitaker, line 7: Needs new description.

system/acpi.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ *//*
+ * Copyright (C) 2020-2022 Martin Whitaker.
+ * Copyright (C) 2020-2022 Sam Whitaker.
```
martinwhitaker, line 11: Wrong name :-)

system/acpi.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+extern acpi_t acpi_config;
+
+/**
+ * ACPI Table Checksum Function
+ */
+int acpi_checksum(const void *data, int length);
+
+/**
+ * Initialises the SMP state and detects the number of available CPU cores.
```
martinwhitaker, line 52: Need new description.

system/smp.c
```diff
@@ -426,32 +374,17 @@ static bool find_cpus_in_floating_mp_struct(void)
...
-        if (*ptr == RSDPSignature1 && *(ptr+1) == RSDPSignature2 && checksum(ptr, 20) == 0) {
-            if (rp->revision < 2 || (rp->length < 1024 && checksum(ptr, rp->length) == 0)) {
-                return rp;
-            }
-        }
-        ptr += 4;
-    }
-    return NULL;
-}
+    if(acpi_config.madt_addr == 0) return false;
```
martinwhitaker, line 379: Whitespace.

PR #113, opened 2022-07-03
Add much bigger PASS/FAIL banner

x86fr:
The goal is to see the actual test result (PASS or FAIL) far away. 

Pressing any key will remove the banner.

### DIFF
--- a/app/main.c
+++ b/app/main.c
@@ -4,8 +4,8 @@
 // Derived from memtest86+ main.c:
 //
 // MemTest86+ V5 Specific code (GPL V2.0)
-// By Samuel DEMEULEMEESTER, sdemeule@memtest.org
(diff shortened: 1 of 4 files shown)


app/main.c
```diff
@@ -629,7 +629,10 @@ void main(void)
             display_pass_count(pass_num);
             if (error_count == 0) {
                 display_status("Pass   ");
-                display_notice("** Pass completed, no errors **");
+                if (!big_status_displayed) {
```
martinwhitaker, line 632: This condition could be handled inside display_big_status(). Then big_status_displayed wouldn't need to be globally visible.

x86fr: Right. Will be changed before merge.

PR #136, opened 2022-07-28
RFC: Add x86 BIOS boot code to header.S

01e3:
Currently, Memtest86+ provides two (three counting mbr.S) different headers:
Both support loading by an intermediate bootloader. Also, both are derived
from the Linux kernel, but from a very different chapter in its evolution.
FDD loading was removed a very long time ago, many, many years before adding
PE/COFF headers for UEFI. Also, the PE/COFF headers are large (even larger
for x86-64) and take a lot of very limited space, leaving very almost no
headroom for the actual code, making this a bit harder.

Harder but not impossible.

This commit adds BIOS boot code to header.S, which should eventually enable
to maitain a single, all-in-one binary that can be booted either by an
intermediate bootloader (such as GRUB or LILO), UEFI and BIOS. Easier, and less
confusing for users, as I have seen cases when people assume memtest.bin is
needed for non-UEFI systems, even if never intented for FDD loading.

Overall, the code is derived from bootsect.S with changes, most importantly:

- Print "Loading" message immediately after setting up the custom FDD
   parameter table. The sooner the better.

- Instead of overriding the beginning of the code (which assumes BIOS
   sets DH to 0 and DL=0 -> FDD), use stack for the FDD table. Yes, the
   current code in bootsect.S uses DX assuming it is 0 but never initializes it.

- Print "MT86+" after successfully loading setup.S. Not enough space
   for the full name. Yet. :)

- Remove register dump in case of error - no space for it. However,
   print "X" after each failed read. Less spammy and IMO may even
   work better for an average user who may not understand how int13h works.

- Use $MAIN_SEG in the sector count detection routine. Less memory
   touched, and can be immediately re-used later for more compact code.

- Replace %cs with %ds for accessing variables (shorter x86
   instructions) and explicitly prefix all with %ds.

- Don't allocate space for variables that start with 0. Instead,
   locate them in the PC/COFF header. There are plenty of fields set
   to 0 there.

- Several other small changes here and there to make the code more
   compact and save instructions.

Probably not the final version ready to commit, but should be very close to it.
Error handling works well for loading both setup.S and the main program code
and the loader recovers from read failures (such as ejected disk, unreadable
sectors) always booting a working Memtest86+ if only the errors were
transient. Tested multiple times. ;)

Also, while there are several other options for fitting more code, which I plan
to explore later, for now I tried this to be as least disruptive as possible -
no changes have been made to build files, memtest.bin nor anywhere outside
header.S. All it does is to allow BIOS to boot memtest.efi, keeping the existing
PE/COFF headers untouched.

That said, at some point it may make sense to consider removing memtest.bin which
should also allow to simplify the Makefiles. We are definitely not there yet.

### DIFF
--- a/boot/header.S
+++ b/boot/header.S
@@ -1,11 +1,23 @@
 // SPDX-License-Identifier: GPL-2.0
 //
-// header.S supports booting directly from a UEFI BIOS or via an intermediate
-// bootloader that supports the Linux boot protocol. When booted directly from
-// the BIOS, it provides the MS-DOS & PE/COFF headers. When using an intermediate
-// bootloader, it provides the first few bytes of the Linux boot header (at the
-// end of the boot sector), with the remainder of the header being provided by
-// setup.S.
+// header.S supports booting directly from the BIOS/UEFI or via an intermediate
+// bootloader that supports the Linux boot protocol. All-in-one.
+//
+// To support loading from the UEFI, it provides the MS-DOS & PE/COFF headers.
+//
+// When using an intermediate bootloader, it provides the first few bytes of
+// the Linux boot header (at the end of the boot sector), with the remainder
+// of the header being provided by setup.S.
+//
+// Finally, if booted directly from the BIOS, it gets loaded at address 0x7C00
+// with DL register set to the boot drive number: 0x0 - FDD, 0x80 - first HDD.
+// Once started, it creates a new Floppy Disk Drive parameter table and prints
+// boot_msg - "Loading". Next it tries to load setup.S immediately after itself
+// (address 0x7E00). Once succesfull, it prints " MT86+" and starts loading
+// the main program code at segment MAIN_SEG (address 0x10000), using BIOS
+// int 13h services to read the data from the disk. After each successfull read
+// a dot (.) is printed. When read fails, "X" gets printed and read operation is
+// restarted.
 //
 // Copyright (C) 2020 Martin Whitaker.
 //
@@ -32,52 +44,73 @@
 	.section ".header", "ax", @progbits
 	.code16
 
+# The BIOS boot entry point. This will be located at 0x7c00.
+
 	.globl	boot
 boot:
-	# "MZ", the MS-DOS header signature.
+	# "MZ", the MS-DOS header signature. Fortunately it translates
+	# to harmless x86 instructions.
 	.byte 0x4d
 	.byte 0x5a
 
-        # In case we are booted by a legacy BIOS, print an error message.
-        # Fortunately the MS-DOS header translates to harmless instructions.
+	# Initialize the segment registers and the stack.
 
-	ljmp	$BOOT_SEG, $(error - boot)
-error:
+	ljmp	$BOOT_SEG, $init
+init:
 	movw	%cs, %ax
 	movw	%ax, %ds
 	movw	%ax, %es
 	movw	%ax, %ss
-	xorw	%sp, %sp
-	sti
+
+	movw	$(BOOT_STACK_TOP - 12), %bp
+	movw	%bp, %sp
+
+	# Old BIOS's default disk parameter tables will not recognize
+	# multi-sector reads beyond the maximum sector number specified
+	# in the default diskette parameter tables - this may mean 7
+	# sectors in some cases.
+	#
+	# Since single sector reads are slow and out of the question,
+	# we must take care of this by creating new parameter tables
+	# (for the first disk) in RAM. We will set the maximum sector
+	# count to 18 - the most we will encounter on an HD 1.44.
+	#
+	# High doesn't hurt. Low does.
+	#
+	# Segments are as follows:
+	#	ds=es=ss=cs = BOOT_SEG,
+	#	fs = 0, gs = parameter table segment
+	#
+	# The table will be stored at the stack, taking 12 bytes.
+
+	xorw	%bx, %bx
+	movw	%bx, %fs
+	movb	$0x78, %bl
+	lgs	%fs:(%bx),%si		# Load source addr from fs:bx to gs:si
+
+	movw	%sp, %di		# Set es:di from ss:sp as the destination
+	movw	$6, %cx 		# copy 12 bytes
+
 	cld
+	rep	movsw %gs:(%si), %es:(%di)
 
-	movw	$error_msg, %si
-0:	lodsb
-	andb	%al, %al
-	jz	wait
-	movb	$0xe, %ah
-	movw	$7, %bx
+	movb	$18, %ss:4(%bp)		# patch sector count
+
+	movw	%bp, %fs:(%bx)
+	movw	%es, %fs:2(%bx)
+
+	# Called with bh=0
+	movb	$0x03, %ah		# Read cursor pos.
 	int	$0x10
-	jmp	0b
 
-wait:
-	# Allow the user to press a key, then reboot.
-	xorw	%ax, %ax
-	int	$0x16
-	int	$0x19
+	jmp	boot_done		# That's all that fits here, continue...
 
-	# int 0x19 should never return. In case it does, invoke the BIOS.
-        # reset code.
-	ljmp	$0xf000,$0xfff0
+sread: .word 1 + SETUP_SECS		# sectors read of the current track
 
-        # The PE header pointer.
+	# The PE header pointer.
 	.org	0x3c
 	.long	pe_header
 
-error_msg:
-	.ascii	"This is a UEFI bootable image\r\n"
-	.ascii	"\n"
-        .asciz  "Press any key to reboot\r\n"
 
 pe_header:
 	.ascii	"PE"
@@ -141,8 +174,11 @@ extra_header_fields:
 	.long	512				# FileAlignment
 	.word	0				# MajorOperatingSystemVersion
 	.word	0				# MinorOperatingSystemVersion
+head:					# Also variable: current head
 	.word	0				# MajorImageVersion
+track:					# Also variable: current track
 	.word	0				# MinorImageVersion
+sectors:				# Also variable: sector count=18/15/9
 	.word	0				# MajorSubsystemVersion
 	.word	0				# MinorSubsystemVersion
 	.long	0				# Win32VersionValue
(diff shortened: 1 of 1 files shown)


boot/header.S
```diff
@@ -182,6 +218,166 @@ section_table:
...
+	ljmp	$SETUP_SEG, $0
+
+# This subroutine loads the system at address 0x10000, making sure no 64KB
+# boundaries are crossed. We try to load it as fast as possible, loading
+# whole tracks whenever we can.
+#
+# in:	es - starting address segment (normally 0x1000)
+#
+
+read_it:
```
debrouxl, line 300: I see a single reference to this label in this commit. Can this subroutine be inlined ? Although if it can, it remains to be seen whether it actually saves space, of course :)

01e3: Yes, very god point, thanks for a flagging and thank you for a prompt initial review.

Indeed, I also noticed that and I think we can save 4 bytes there (3 from the call, one from ret). Maybe even more with some changes in the code flow and by using more registers. I also would like to reduce the number of variables kept in the memory.

As mentioned, there are several more opportunities, which I will definitely explore. Just wanted to show what I have been working on and get some initial feedback. If we decide it makes sense, I promise we can fit the proper project name there before I consider this done, so more changes will follow. ;)

Perhaps  we can even fit the functionality of mbr.S, but don't want to over-promise.

martinwhitaker:
Take a look at pull request #34. I think you'll need to save quite a bit more space if we plan to support secure boot (not a priority for me, but other people seem to want it).

01e3:
Thanks, looking there now. I think Secure Boot is an important goal, I will need to find a way to support both of the efforts.

Now... If I'm correctly reading the code and did the math correctly it seems we we are extending PE/COFF headers by 48 bytes with DataDirectory and then adding two additional sections ".reloc" and ".sbat", each 40 bytes long, so 128 bytes total. That's indeed a lot but nothing is impossible.

One of the options we have is to follow what Linux kernel did, that merged bootsect.S and setup.S into header.S. This would require a little bit more effort, but this can give us all the space we need, assuming I can fit enough code in the first sector to complete load_setup.

Another one is to move PE/COFF outside the first critical sector.

What is the ETA for merging the Secure Boot changes?

x86fr:
Secure Boot support is really the top priority to make Memtest86+ easily usable by as many people as possible.

Without secure boot support, an increasing number of computer will be unable to run Memtest86+ out of the box. 
Secure Boot might even become mandatory at some point, starting with Tier A brand like Dell and HP.

For the rest, my 2 cents: mixing both (BIOS & UEFI) support in a single .S is technically clever and sounds good at first sight. BUT this is not, by itself, interesting enough to justify the addition of serious code-size restrictions that could led to limitations (or added complexity) in future improvements. I think any actual or future developers will vastly prefer working on two files that can be modified easily, versus a single merged file that requires counting every byte at each modification.

From a betatesting point of view, heavily altering the header files requires a lot of testing as this is a critical part of the boot process. That will be required when Secure Boot support will be added, so it's probably better to wait for secure boot support first, then try to merge both header files.

Finally, I wonder if the best solution is not to split Memtest86+ into two binaries : 
- One binary with BIOS support, i586 and legacy code.
- One binary with EFI support, SSE2 and modern technology.

I'm not sure for how long we will be able to merge technologies split in a 30+ years time line into a single binary.

All opinions welcome. :)

01e3:
Fair point about code-size restrictions being a risk longer term. As mentioned, we have plenty of space, we just need to move some structures around, so it seems this is the direction I would need to take.

Strongly agree about the Secure Boot, including the priority and the sequence. This is why I asked when the code is planned to be merged, as any further work should happen only once this part is landed.

Regarding splitting the binaries - what I have been trying to avoid is "which version do I need" and looking for the proper media when I touching a random PC. That said, to some extend this is what we have already given the 32bit and the 64 bit versions, except we currently have four with the BIOS / UEFI, even five if we count mbr.S as a separate one. Two (32/64) seems to be my goal. One could be done for BIOS, but not for UEFI due to lack of multi-arch support in PE/COFF, unless we would be fine assuming no 32-bit UEFI and only support booting them from BIOS?

Also note that there are many 64 bit systems w/o UEFI support that are less than 10 years old, so putting them into the "legacy" bucket may not be optimal.

debrouxl:
On the one hand, it's true that we don't know how much longer the BIOS and EFI boot capabilities can be merged into a single binary per architecture; on the other hand, neither format can change, otherwise there will be backwards compatibility headaches, probably even for some recent systems, if the given UEFI BIOS implementation is picky.
As such, I'd *think* (and I could be wrong about it) that if - and probably when - all of BIOS, UEFI without Secure Boot and UEFI with Secure Boot can be made to work in a single binary per architecture, then it's going to work for an extended period of time.

I also think that keeping the current 64-bit BIOS binary is desirable for the years to come, for the reason given by 01e3: some 64-bit systems from the 2010s do not support UEFI. Some of these systems will keep serving professional production workloads for the next 5-10 years, at least :)

martinwhitaker:
32-bit UEFI machines exist (Intel Bay Trail). The one I have does not have a CSM, so is UEFI boot only.

debrouxl:
I spent a bit of time rebasing these changes on top of the Secure Boot changes, and trying to shoehorn the code into the significantly reduced available space. As expected, and as much as the concept of a single unified binary image is conceptually interesting and could be interesting in practice for users, the prospects of obtaining such a thing have become bleak, especially on x86_64.

Even after commenting out `int $0x10` calls and the corresponding strings, inlining the functions called once, and hard-coding the number of sectors per track to 18, the code before `.org 497` overflows the available area by 23 bytes on x86 and 39 bytes on x86_64.
Switching from multi-sector reads to single sector reads _may_ save _some_ space, by not having to build a new disk parameter table, at significant load time penalty according to the comments. However, saving _enough_ space is a whole other matter, though it's less hard on x86 than on x86_64.

Diff against 186ef6e91333badfe315c9b2616991bd238da3f6 , a number of comments not updated:
```
diff --git a/boot/header.S b/boot/header.S
index 673bc67..7c4429b 100644
--- a/boot/header.S
+++ b/boot/header.S
@@ -1,11 +1,23 @@
 // SPDX-License-Identifier: GPL-2.0
 //
-// header.S supports booting directly from a UEFI BIOS or via an intermediate
-// bootloader that supports the Linux boot protocol. When booted directly from
-// the BIOS, it provides the MS-DOS & PE/COFF headers. When using an intermediate
-// bootloader, it provides the first few bytes of the Linux boot header (at the
-// end of the boot sector), with the remainder of the header being provided by
-// setup.S.
+// header.S supports booting directly from the BIOS/UEFI or via an intermediate
+// bootloader that supports the Linux boot protocol. All-in-one.
+//
+// To support loading from the UEFI, it provides the MS-DOS & PE/COFF headers.
+//
+// When using an intermediate bootloader, it provides the first few bytes of
+// the Linux boot header (at the end of the boot sector), with the remainder
+// of the header being provided by setup.S.
+//
+// Finally, if booted directly from the BIOS, it gets loaded at address 0x7C00
+// with DL register set to the boot drive number: 0x0 - FDD, 0x80 - first HDD.
+// Once started, it creates a new Floppy Disk Drive parameter table and prints
+// boot_msg - "Loading". Next it tries to load setup.S immediately after itself
+// (address 0x7E00). Once succesfull, it prints " MT86+" and starts loading
+// the main program code at segment MAIN_SEG (address 0x10000), using BIOS
+// int 13h services to read the data from the disk. After each successfull read
+// a dot (.) is printed. When read fails, "X" gets printed and read operation is
+// restarted.
 //
 // Copyright (C) 2020 Martin Whitaker.
 //
@@ -33,52 +45,76 @@
        .section ".header", "ax", @progbits
        .code16
 
+# The BIOS boot entry point. This will be located at 0x7c00.
+
        .globl  boot
 boot:
-       # "MZ", the MS-DOS header signature.
+       # "MZ", the MS-DOS header signature. Fortunately it translates
+       # to harmless x86 instructions.
        .byte 0x4d
        .byte 0x5a
 
-        # In case we are booted by a legacy BIOS, print an error message.
-        # Fortunately the MS-DOS header translates to harmless instructions.
+       # Initialize the segment registers and the stack.
 
-       ljmp    $BOOT_SEG, $(error - boot)
-error:
+       ljmp    $BOOT_SEG, $init
+init:
        movw    %cs, %ax
        movw    %ax, %ds
        movw    %ax, %es
        movw    %ax, %ss
-       xorw    %sp, %sp
-       sti
+
+       movw    $(BOOT_STACK_TOP - 12), %bp
+       movw    %bp, %sp
+
+       # Old BIOS's default disk parameter tables will not recognize
+       # multi-sector reads beyond the maximum sector number specified
+       # in the default diskette parameter tables - this may mean 7
+       # sectors in some cases.
+       #
+       # Since single sector reads are slow and out of the question,
+       # we must take care of this by creating new parameter tables
+       # (for the first disk) in RAM. We will set the maximum sector
+       # count to 18 - the most we will encounter on an HD 1.44.
+       #
+       # High doesn't hurt. Low does.
+       #
+       # Segments are as follows:
+       #       ds=es=ss=cs = BOOT_SEG,
+       #       fs = 0, gs = parameter table segment
+       #
+       # The table will be stored at the stack, taking 12 bytes.
+
+       xorw    %bx, %bx
+       movw    %bx, %fs
+       movb    $0x78, %bl
+       lgs     %fs:(%bx),%si           # Load source addr from fs:bx to gs:si
+
+       movw    %sp, %di                # Set es:di from ss:sp as the destination
+       movw    $6, %cx                 # copy 12 bytes
+
        cld
+       rep     movsw %gs:(%si), %es:(%di)
 
-       movw    $error_msg, %si
-0:     lodsb
-       andb    %al, %al
-       jz      wait
-       movb    $0xe, %ah
-       movw    $7, %bx
-       int     $0x10
-       jmp     0b
-
-wait:
-       # Allow the user to press a key, then reboot.
-       xorw    %ax, %ax
-       int     $0x16
-       int     $0x19
+       movb    $18, %ss:4(%bp)         # patch sector count
+
+       movw    %bp, %fs:(%bx)
+       movw    %es, %fs:2(%bx)
 
-       # int 0x19 should never return. In case it does, invoke the BIOS.
-        # reset code.
-       ljmp    $0xf000,$0xfff0
+       # Called with bh=0
+load_setup:
+       xorw    %dx, %dx                        # dl=drive 0, dh=head 0 for below.
+       xorb    %ah, %ah
+#      movb    $0x03, %ah              # Read cursor pos.
+#      int     $0x10
 
-        # The PE header pointer.
+       jmp     boot_done               # That's all that fits here, continue...
+
+sread: .word 1 + SETUP_SECS            # sectors read of the current track
+
+       # The PE header pointer.
        .org    0x3c
        .long   pe_header
 
-error_msg:
-       .ascii  "This is a UEFI bootable image\r\n"
-       .ascii  "\n"
-        .asciz  "Press any key to reboot\r\n"
 
 pe_header:
        .ascii  "PE"
@@ -138,8 +174,11 @@ extra_header_fields:
        .long   512                             # FileAlignment
        .word   0                               # MajorOperatingSystemVersion
        .word   0                               # MinorOperatingSystemVersion
+head:                                  # Also variable: current head
        .word   0                               # MajorImageVersion
+track:                                 # Also variable: current track
        .word   0                               # MinorImageVersion
+sectors:                               # Also variable: sector count=18/15/9
        .word   0                               # MajorSubsystemVersion
        .word   0                               # MinorSubsystemVersion
        .long   0                               # Win32VersionValue
@@ -226,6 +265,159 @@ section_table:
                | IMAGE_SCN_ALIGN_4096BYTES     \
                | IMAGE_SCN_CNT_INITIALIZED_DATA        # Characteristics (section flags)
 
+boot_done:
+       int     $0x13                           # reset FDC
+       # Print the message. Called with bh=0, and dh, dl already set.
+
+#      movw    $boot_msg, %bp
+#      movw    $(boot_msg_end - boot_msg), %cx
+#      movb    $0x07, %bl                      # Attribute 7 (normal).
+#      movw    $0x1301, %ax                    # Write string, move cursor.
+#      int     $0x10
+
+       # Load the setup sectors directly after the boot block.
+       # Note that 'es' is already set up.
+
+       movw    $0x0002, %cx                    # sector 2, track 0
+       movw    $0x0200, %bx                    # address = 512, in BOOT_SEG
+       movw    $(0x0200 + SETUP_SECS), %ax     # service 2, nr of sectors
+                                               # (assume all on head 0, track 0)
+       int     $0x13                           # read it.
+       jc      load_setup                      # error - repeat
+
+#      xor     %bx, %bx
+#      movb    $0x03, %ah                      # Read cursor pos.
+#      int     $0x10
+
+#      movw    $header_msg, %bp
+#      movw    $(header_msg_end - header_msg), %cx
+#      movb    $0x07, %bl                      # Attribute 7 (normal).
+#      movw    $0x1301, %ax                    # Write string, move cursor.
+#      int     $0x10
+
+load_setup_done:
+       # Get disk drive parameters, specifically number of sectors/track.
+       # It seems that there is no BIOS call to get the number of sectors.
+       # Guess 18 sectors if sector 18 can be read, 15 if sector 15 can be
+       # read. Otherwise guess 9.
+
+       movw    $MAIN_SEG, %ax
+       movw    %ax, %es
+       xorw    %bx, %bx                # bx is starting address within segment
+
+       xorw    %dx, %dx                # drive 0, head 0
+#      movw    $0x0012, %cx            # sector 18, track 0
+#      movw    $0x0201, %ax            # service 2, 1 sector
+#      int     $0x13
+#      jnc     got_sectors
+#      movb    $0x0f, %cl              # sector 15
+#      movw    $0x0201, %ax            # service 2, 1 sector
+#      int     $0x13
+#      jnc     got_sectors
+#      movb    $0x09, %cl
+
+#got_sectors:
+       movw    $0x0012, %ds:sectors
+
+       # Load the main test program at address 0x10000, making sure no 64KB
+       # boundaries are crossed. We try to load it as fast as possible, loading
+       # whole tracks whenever we can.
+       #
+       # in:   es - starting address segment (normally 0x1000)
+       #
+
+read_it:
+       # es:bx is already set to $MAIN_SEG:0 in load_setup_done
+
+rp_read:
+       movw    %es, %ax
+       subw    $MAIN_SEG, %ax          # have we loaded all yet?
+       cmpw    %ds:sys_size, %ax
+       jg      read_end
+
+       movw    %ds:sectors, %ax
+       subw    %ds:sread, %ax
+       movw    %ax, %cx
+       shlw    $9, %cx
+       addw    %bx, %cx
+       jnc     ok2_read
+       je      ok2_read
+       xorw    %ax, %ax
+       subw    %bx, %ax
+       shrw    $9, %ax
+ok2_read:
+#      call    read_track
+read_track:
+       pusha
+#      movw    $0x0e2e, %ax            # 0x2e = "."
+read_track_recover:
+#      movw    $00007, %bx
+#      int     $0x10
+#      popa
+#      pusha
+
+       movw    %ds:track, %dx
+       movw    %ds:sread, %cx
+       incw    %cx
+       movb    %dl, %ch
+       movw    %ds:head, %dx
+       movb    %dl, %dh
+       andw    $0x0100, %dx
+       movb    $2, %ah
+
+       int     $0x13
+       jnc     read_track_end
+bad_rt:
+       xorb    %ah, %ah
+       int     $0x13                   # Reset FDC, dl already points to the drive id
+
+#      movw    $0x0e58, %ax            # Leave 0x58 = "X" to be printed
+       jmp     read_track_recover
+
+read_track_end:
+       popa
+       movw    %ax, %cx
+       add     %ds:sread, %ax
+       cmpw    %ds:sectors, %ax
+       jne     ok3_read
+       movw    $1, %ax
+       subw    head, %ax
+       jne     ok4_read
+       incw    track
+ok4_read:
+       movw    %ax, %ds:head
+       xorw    %ax, %ax
+ok3_read:
+       movw    %ax, %ds:sread
+       shlw    $9, %cx
+       addw    %cx, %bx
+       jnc     rp_read
+       movw    %es, %ax
+       addb    $0x10, %ah
+       movw    %ax, %es
+       xorw    %bx, %bx
+       jmp     rp_read
+
+       # FDD motor will be switches off via floppy_off() called from
+       # global_init().
+
+read_end:
+       # Fix up the Linux boot header to indicate we've loaded into low memory.
+       movl    $LOW_LOAD_ADDR, code32_start
+
+       # After that (everything loaded), we jump to the setup code loaded
+       # directly after the boot block.
+
+       ljmp    $SETUP_SEG, $0
+
+#boot_msg:
+#      .ascii "Loading"
+#boot_msg_end:
+
+#header_msg:
+#      .ascii " MT86+"
+#header_msg_end:
+
 # Emulate the Linux boot header, to allow loading by intermediate boot loaders.
 
        .org    497
```

01e3:
Thanks Lionel!

This is fine and kind expected based on our earlier discussion. Now when we have all the secure boot code in place and we also understand we want to leave some headroom for future changes that me be required, my plan here is to merged bootsect.S and setup.S into header.S, as it was done in the Linux kernel. With this, we will also have more space for the code itself and no longer need to count every single byte.

medhefgo:
I just had a thought (mind you, I have no idea about legacy boot restrictions): Why not just hide the setup headers in the DOS stub? Then you can add as much x86 boot code before/after it and the rest of the PE image follows. This would allow adding any number of new PE sections without any issues too (`.osrel` comes to mind).

```
DOS header (Points to PE header as usual)
Legacy boot code
Setup header
Extra legacy boot code
PE headers
```

Firmware should be loading all of SizeOfHeaders into memory when the image is loaded, so the setup headers would still be available to the PE code. But should there ever be a firmware that doesn't one could always add a duplicate setup header as part of a PE section (one could even try to have PointerToRawData offset directly into the DOS stub).

PR #148, opened 2022-08-20
Last minute patches for 6.00 :)

01e3:
This is another set of changes I have been working on for some time that I think may be good candidates for 6.10.

- "smbus: refactor SPD SKU (part number) handling" is an optimization / cleanup that also fixes two bugs regarding parsing SPD SKU (part number) - handling spaces and the length. While this is clearly a bug fix, I suspect it may be too late for 6.00 given the scope of changes, especially that the bugs are not critical IMO.

Before:
obraz
After:
obraz
(Note that 0x0313 has been added in https://github.com/memtest86plus/memtest86plus/commit/53f61e6b87606244da67d1a05961aed16a18ed44)

- "smbus: optimize parse_spd_* functions" is a further optimization / cleanup, so no functional changes.
 
- "selftest: Introduce selftests" adds some initial self-tests, including testing fixed SPD part number parsing code from the first patch above. The other goal was to uncover a number of problems with the current code for DDR/DDR2/DDR3/DDR4 (note that I have not tested SDRAM, RDRAM and DDR5), where the timing reported by Memtest86+ is wrong when compared to all of: vendor's specification, decode-dimms and CPU-Z - please see the comments in the file. Please let me know if I should file a single bugs for all of the issues, or separate bug for each of DDR/DDR2/DDR3/DDR4?

- I also would like to hear feedback regarding the change itself - the concept of adding selftests and the implementation. Given this is a bare-metal C code (that needs freestanding / no-builtin) with a lot of static functions, the tests require a little bit of hacks, but my goal has been to keep the changes to the main codebase as small as possible. Alternatively, some "ifdefs" could be added that could simplify the selftests, so I would appropriate an advice regarding the direction, and the overall implementation.

Thanks!

### DIFF
--- a/system/smbus.c
+++ b/system/smbus.c
@@ -54,6 +54,11 @@ static uint8_t ich5_process(void);
 static uint8_t ich5_read_spd_byte(uint8_t adr, uint16_t cmd);
 static uint8_t nf_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
 
+static inline uint8_t bcd_to_ui8(uint8_t bcd)
+{
+    return bcd - 6 * (bcd >> 4);
+}
+
 void print_smbus_startup_info(void)
 {
     uint8_t spdidx = 0, spd_line_idx = 0;
@@ -385,12 +390,8 @@ static void parse_spd_ddr5(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 521, 30);
 
-    // Week & Date (BCD to Int)
-    uint8_t bcd = get_spd(slot_idx, 515);
-    spdi->fab_year =  bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 516);
-    spdi->fab_week =  bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 515));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 516));
 
     spdi->isValid = true;
 }
@@ -508,12 +509,8 @@ static void parse_spd_ddr4(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 329, 20);
 
-    // Week & Date (BCD to Int)
-    uint8_t bcd = get_spd(slot_idx, 323);
-    spdi->fab_year =  bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 324);
-    spdi->fab_week =  bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 323));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 324));
 
     spdi->isValid = true;
 }
@@ -650,11 +647,8 @@ static void parse_spd_ddr3(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 128, 18);
 
-    uint8_t bcd = get_spd(slot_idx, 120);
-    spdi->fab_year =  bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 121);
-    spdi->fab_week =  bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 120));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 121));
 
     spdi->isValid = true;
 }
@@ -790,11 +784,8 @@ static void parse_spd_ddr2(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 73, 18);
 
-    uint8_t bcd = get_spd(slot_idx, 93);
-    spdi->fab_year = bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 94);
-    spdi->fab_week = bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 93));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 94));
 
     spdi->isValid = true;
 }
@@ -882,11 +873,8 @@ static void parse_spd_ddr(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 73, 18);
 
-    uint8_t bcd = get_spd(slot_idx, 93);
-    spdi->fab_year = bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 94);
-    spdi->fab_week = bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 93));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 94));
 
     spdi->isValid = true;
 }
@@ -958,19 +946,14 @@ static void parse_spd_rdram(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 73, 18);
 
-    uint8_t bcd = get_spd(slot_idx, 93);
-    spdi->fab_year = bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 94);
-    spdi->fab_week = bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 93));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 94));
 
     spdi->isValid = true;
 }
 
 static void parse_spd_sdram(spd_info *spdi, uint8_t slot_idx)
 {
-    uint8_t bcd;
-
     spdi->type = "SDRAM";
 
     uint8_t spd_byte3  = get_spd(slot_idx, 3) & 0x0F; // Number of Row Addresses (2 x 4 bits, upper part used if asymmetrical banking used)
@@ -1029,11 +1012,8 @@ static void parse_spd_sdram(spd_info *spdi, uint8_t slot_idx)
 
     read_sku(spdi->sku, slot_idx, 73, 18);
 
-    bcd = get_spd(slot_idx, 93);
-    spdi->fab_year = bcd - 6 * (bcd >> 4);
-
-    bcd = get_spd(slot_idx, 94);
-    spdi->fab_week = bcd - 6 * (bcd >> 4);
+    spdi->fab_year = bcd_to_ui8(get_spd(slot_idx, 93));
+    spdi->fab_week = bcd_to_ui8(get_spd(slot_idx, 94));
 
     spdi->isValid = true;
 }
--- a/system/smbus.c
+++ b/system/smbus.c
@@ -28,6 +28,7 @@ static int8_t spd_page = -1;
 static int8_t last_adr = -1;
 
 // Functions Prototypes
+static void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len);
 
 static spd_info parse_spd_rdram (uint8_t slot_idx);
 static spd_info parse_spd_sdram (uint8_t slot_idx);
@@ -153,9 +154,8 @@ static void print_spdi(spd_info spdi, uint8_t lidx)
     }
 
     // Print SKU
-    for(i = 0; i < spdi.sku_len; i++) {
-        curcol = printc(LINE_SPD+lidx, i ? curcol : ++curcol, spdi.sku[i]);
-    }
+    if (*spdi.sku)
+        curcol = prints(LINE_SPD+lidx, curcol + 1, spdi.sku);
 
     // Check manufacturing date and print if valid.
     // fab_year is uint8_t and carries only the last two digits.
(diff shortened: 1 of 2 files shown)


x86fr:
I'll review the whole PR next week. Many changes look nice without adding complexity. Great! 

About selftest, I'm not sure if this must be added to main. A separate debug branch could be useful too. I also use my own tools to do the same things (SPD debugging). Will recheck everything as soon as I'm back home.

debrouxl:
In the next few weeks, I meant to work a bit on related code bits, by rebasing / rewriting the fuzzing branch for SPD data and SMBIOS data. Also, I felt that my quick & dirty debugging patch posted inline in #140 should be turned into a usable debug front-end for feeding table data to some ACPI table data parsers. Looks like great minds think alike :)

IMO, from the get go, we should attempt to implement a generic solution able to deal with SPD data for three relevant purposes:
* implementing unit tests, which you did here;
* implementing fuzzing tests, which I did months ago, but left by the wayside (for merging, that is) until after the 6.00 release. Those require a single function dealing with memory-mapped arguments, and the memory type parsing and code dispatching is performed below that entry point;
* implementing a debugging front-end, for feeding SPD data from a file to memtest86+'s SPD data parsing code, producing human-readable output (and why not some icing on the cake while at it, namely producing ready-made C code to be checked by humans and copy-pasted into unit tests). This should also end up calling a single function with memory-mapped arguments, after either reading a file's data e.g. through a full hosted implementation of stdio.h, or just mapping it to memory through POSIX/*nix `mmap`, and in any case, dealing with the allowed data sizes and various I/O errors.

We know that it's not just SPD data: parsing code for SMBIOS data and ACPI data (all of top-level - RSDT, XSDT - and underlying tables), all of which are memory-mapped, unlike SPD data, eventually need to be organized the same way, for the same reasons.
But obviously, while a common design taking into account the mid-term idea of multiple functions possibly sharing some common bits (not sure there will be much code to share, actually) ought to be desirable, let's work on one thing at a time: SPD first, SMBIOS second and ACPI later, I suppose :)

01e3:
> I'll review the whole PR next week. Many changes look nice without adding complexity. Great!

Thanks! I have several more changes like this, should I wait for 6.10 or keep sending them? Also, when are we planning to have the 6.10 branch?
 
> About selftest, I'm not sure if this must be added to main. A separate debug branch could be useful too. I also use my own tools to do the same things (SPD debugging). Will recheck everything as soon as I'm back home.

There are multiple benefits of having unit-tests included in the same branch:
 - It is easier to keep the tests in sync with the code, otherwise they pretty quickly stop working with the ongoing changes.
 - Other contributors can quickly and easily check their changes before sending PRs.
 - We should eventually be able to setup pre-commit tests automatically testing submitted patches.

01e3:
Hi!

> In the next few weeks, I meant to work a bit on related code bits, by rebasing / rewriting the fuzzing branch for SPD data and SMBIOS data. Also, I felt that my quick & dirty debugging patch posted inline in #140 should be turned into a usable debug front-end for feeding table data to some ACPI table data parsers. Looks like great minds think alike :)

Ah, indeed. Perhaps done in similar way, as you described below for the SPD.

> IMO, from the get go, we should attempt to implement a generic solution able to deal with SPD data for three relevant purposes:
> 
>     * implementing unit tests, which you did here;

Sure. Just that I would not call them fully implemented yet, but we need to start somewhere.

>     * implementing fuzzing tests, which I did months ago, but left by the wayside (for merging, that is) until after the 6.00 release. Those require a single function dealing with memory-mapped arguments, and the memory type parsing and code dispatching is performed below that entry point;

Oh, did not know that. Is this https://github.com/memtest86plus/memtest86plus/commit/bf03cf79188a810928dcaa11a721d24c2cda6328?
I think with the changes I have made, it should be very easy to handle that part - almost all the callbacks are there already.

>     * implementing a debugging front-end, for feeding SPD data from a file to memtest86+'s SPD data parsing code, producing human-readable output (and why not some icing on the cake while at it, namely producing ready-made C code to be checked by humans and copy-pasted into unit tests). This should also end up calling a single function with memory-mapped arguments, after either reading a file's data e.g. through a full hosted implementation of stdio.h, or just mapping it to memory through POSIX/*nix `mmap`, and in any case, dealing with the allowed data sizes and various I/O errors.

Great idea! Assuming we are talking about something that just pretty prints data from spd_info for a given SPD blob (similar to what decode-dimms can also do), it should be also very easy to implement based on what I have wrote already.

> We know that it's not just SPD data: parsing code for SMBIOS data and ACPI data (all of top-level - RSDT, XSDT - and underlying tables), all of which are memory-mapped, unlike SPD data, eventually need to be organized the same way, for the same reasons. But obviously, while a common design taking into account the mid-term idea of multiple functions possibly sharing some common bits (not sure there will be much code to share, actually) ought to be desirable, let's work on one thing at a time: SPD first, SMBIOS second and ACPI later, I suppose :)

Ah, yes, definitely. Eventually, most if not all functions should be tested, especially ones that handle complex BIOS / UEFI supplied data, so on top of SMBIOS/ACPI, we should also things like e801/e820/88, etc. What makes it a little bit more complicated is that some of that logic lives in ASM files, but nothing is impossible.

x86fr:
Anyway, please keep it simple.

We must think twice before adding a lot of complexity on tests to solve non-existent issues, or to solve minor issues that can be solved with minimum efforts. On my side, I never spend a lot of time on "potential" bugs that have not been reported by a real user. I know this is very bad practice, but time is limited and I must prioritize tasks. 

Also, having complex tests and fuzzing code in the main branch means we need someone to maintain this code over an extended period of time. And this is painful: I use a wrapper to inject SPD dumps from CPUZ to the Memtest86+ SPD code and it requires a lot of work. As long as tests are fully independent from the main code, it's fine, but we really don't want to stop someone from solving an issue just because he also have to understand and fix the related test.

debrouxl:
> Oh, did not know that. Is this bf03cf79188a810928dcaa11a721d24c2cda6328?
Yup, bf03cf79188a810928dcaa11a721d24c2cda6328 is the current HEAD of the fuzzing branch, dating from before SPD-related refactoring and improvements were merged.

> Great idea! Assuming we are talking about something that just pretty prints data from spd_info for a given SPD blob (similar to what decode-dimms can also do), it should be also very easy to implement based on what I have wrote already.
Indeed. Bonus points for generating both human-readable and unit test C code, which is just a different form of pretty-printing anyway.

> Eventually, most if not all functions should be tested, especially the ones that handle complex BIOS / UEFI supplied data, so on top of SMBIOS/ACPI, we should also things like e801/e820/88, etc.
Right. I did not envision fuzzing e801, e820, memory ranges code just yet, but that's an interesting idea as well, for the even longer term. Not that issues in these code bits would seem much of a relevant security issue to me: pure DoS upon malformed BIOS-provided data is annoying for users but harmless, and as for arbitrary code execution... AFAICT, if attackers can boot into memtest86+, they should usually be able to boot other OS in the first place, especially if Secure Boot is disabled.

> Anyway, please keep it simple.
Sure. I'm not against simplicity in itself, but I know that it can be hard to make things as simple as possible, while avoiding the pitfall of constructs simplistic to the point that they're limited and will require later refactoring (i.e. more maintainer / contributor time spent on the matter) to deal with foreseen use cases :)
Put another way: if the development and maintenance cost of designing better infrastructure for the next step is low, I'm all for doing it and reducing future technical debt, even if it takes a bit of extra effort from myself. That goes in my day job as well.

> Also, having complex tests and fuzzing code in the main branch means we need someone to maintain this code over an extended period of time.
AFAICT from experience on other projects maintained by other persons where I contributed fuzzing effort and sometimes initial fuzzing code (usually improved upon by the respective maintainers), such code is low maintenance after the initial refactoring, if any... especially if it's part of the main branch and it is thereby exercised by automatic build and testing systems.
Pre-commit tests mentioned by 01e3, which reduce maintainer burden, and later integration into the likes of OSS-Fuzz, which has been shown on other projects to prevent some security-relevant bugs from hitting end users, after being automatically reported and acted upon by humans before a release was cut, are good.

> And this is painful: I use a wrapper to inject SPD dumps from CPUZ to the Memtest86+ SPD code and it requires a lot of work.
Good. Could you post that wrapper to inject CPU-Z SPD dumps into the memtest86+ SPD code, so that the requirement can be integrated into the current design ? :)

> As long as tests are fully independent from the main code, it's fine, but we really don't want to stop someone from solving an issue just because he also have to understand and fix the related test.
Keeping the entry barrier to contribution low is desirable, but I'm not convinced that maintainers and users have much to gain by not requiring that automated tests do not regress before a PR is merged. By now, that's standard best practice for many projects, and memtest86+ is currently behind the curve on that front... but that's partially going to be solved by this PR :)

x86fr:
At this time, it's probably better to split this PR in 2 (PR#1 tests PR#2 all others).

debrouxl:
About the optimization commits: since all functions are static, I expected the compiler to do a better job with the code generation related to inlining and returning spd_info by value. Likewise, the impact of outlining the SKU reading code is higher than I'd have expected. Trust but verify, I guess - good thing you did it :)

Beyond the SKU reading bugfix, in relative terms, cutting over 1% of the size of `memtest_shared`, for both 32-bit and 64-bit binaries, is nothing to sneer at. I have already mentioned another optimization on the `jep106` array of structs, which is currently well over a KB large on x86_64, and whose size can be reduced by over three quarters on x86_64, without impairing maintainability... I need to spend some time on the matter.

01e3:
> About the optimization commits: since all functions are static, I expected the compiler to do a better job with the code generation related to inlining and returning spd_info by value. Likewise, the impact of outlining the SKU reading code is higher than I'd have expected. Trust but verify, I guess - good thing you did it :)

I think most of the savings come from the removal of the duplicated code which became possible by the change. I was actually once again surprised how smart compilers are these days when I looked at the asm code.

> Beyond the SKU reading bugfix, in relative terms, cutting over 1% of the size of `memtest_shared`, for both 32-bit and 64-bit binaries, is nothing to sneer at. I have already mentioned another optimization on the `jep106` array of structs, which is currently well over a KB large on x86_64, and whose size can be reduced by over three quarters on x86_64, without impairing maintainability... I need to spend some time on the matter.

Which brings another questions - does it make sense to consider having an "init" part of the binary, that gets dropped after the program starts? For example, all that SPD parsing / printing code and the mentioned jep106 struct provide value only once. That would be a large change of course, but I wonder if this is something worth looking into?

Thanks!

01e3:
> At this time, it's probably better to split this PR in 2 (PR#1 tests PR#2 all others).

Sure, good idea, done.

Dropped the tests from this PR and will create a different branch and a PR for it. I expect much more discussion will be needed.

At the same time, I added one more (small) change here. ;)

x86fr:
> Right. I did not envision fuzzing e801, e820, memory ranges code just yet, but that's an interesting idea as well, for the even longer term. Not that issues in these code bits would seem much of a relevant security issue to me: pure DoS upon malformed BIOS-provided data is annoying for users but harmless, and as for arbitrary code execution... AFAICT, if attackers can boot into memtest86+, they should usually be able to boot other OS in the first place, especially if Secure Boot is disabled.

If you find a way to compromise a firmware with forged malicious tables and then sign it with a valid key, the whole Secureboot process is already broken and you really don't need to mess with Memtest86+ to do anything you want. The only strings that are user writable without recompiling the firmware package are DMI strings, and the input values are checked very strictly by all current UEFI.

> Could you post that wrapper to inject CPU-Z SPD dumps into the memtest86+ SPD code, so that the requirement can be integrated into the current design ? :)

The method I use requires an admin access to CPU-Z with an auth mechanism I can't fully disclose.

But OK, let's discuss and compare my quick&dirty simple way vs a more complex method like this one to determine the best "time spent vs user benefit" ratio of both. 

We can take this exact issue (SPD parsing) as an example. I can disclose almost all the code I wrote internally to exercise the patch, and explain exactly how I solved the SPD issues related to timings computation. My commit is already being tested right now. The betatesting phase was planned to end on Tuesday, with the commit being pushed on Wednesday. I'll take a few more day to post a detailed listing on the new PR related to tests.

x86fr:
> At the same time, I added one more (small) change here. ;)

Great! I'll check with the betatesters tomorrow if they have time to test this PR for inclusion in 6.00. Unfortunately, many of them are on vacations and the whole team is moving to a bigger warehouse in a couple weeks, so their schedule is tight. :-/

system/smbus.c
```diff
@@ -182,14 +182,38 @@ static void print_spdi(spd_info spdi, uint8_t lidx)
...
+
+        // Stop on the first non-ASCII char
+        if (sku_byte < 0x20 || sku_byte > 0x7F)
+            break;
+
+        sku[sku_len] = (char)sku_byte;
+    }
+
+    // Trim spaces from the end
+    while (sku_len && sku[sku_len - 1] == 0x20)
```
x86fr, line 212: Many DRAM manufacturers do not follow the JEDEC SPD specs and right-trim their SKU string with 0x00 or other non-ASCII char (especially 0xFF like Corsair).

Can you change this to: `while (sku_len && (sku[sku_len - 1] <= 0x20 || sku[sku_len - 1] > 0x7F))`

01e3:
As discussed, I created a separate branch and https://github.com/memtest86plus/memtest86plus/pull/150 for the selftests.

system/smbus.c
x86fr: That's correct. ;)

01e3: Do you want me to add a comment to make it clear?

01e3:
Not a big deal, I can do this over the weekend. Thanks!

x86fr:
#152 merged. This one's next.

01e3:
I have updated my upstream-next branch with adapted patches, but it was a long day for me and is getting late, so would I like to look at them again, hopefully tomorrow, and run some more tests.

Feel free to take a look, but I will ping this PR once ready.

x86fr:
Sure! I will not build the RC until Wednesday I think.

Thnks for your efforts.

01e3:
I may not be able to do much during the next week so it is now or never. ;)

I assume I should drop the last patch for the DDR3 XMP parsing? Should I keep the part that removes the extra ";" or you will take care of this while fixing the code that calculates tRAS?

x86fr:
Both are fine. I'll correct soon anyway.

01e3:
I dropped the incorrect DDR3 XMP patch, so back to the original set. I think we are ready. Thanks!

Issue #83, opened 2022-05-21
Selecting the error reporting mode through the command line


debrouxl:
At the moment, this is another configuration option which can only be set through the menu.

martinwhitaker:
Do we really want multiple issues for this? Memtest86+ doesn't currently have a proper CLI - the boot command line options are only there to work around hardware quirks that would otherwise prevent you getting to the interactive configuration menu. If this is a desired feature, it would be best considered as a whole rather than piecemeal.

x86fr:
Agree. Also, I can't find a way to easily pass arguments to the efi binary loaded by iPXE. The config menu is much more efficient for complex settings.

x86fr:
IMO, to avoid flooding the issues list, we should convert all these enhancements ideas without an active developer working on it to discussion (#7, #58, #72,#73, #83 and maybe some others). So users can vote and we will spend some time on the most popular ones.

debrouxl:
This one was the last of the "scriptability" tracking issues. Scriptability is the correct way to provide the best building blocks for automated, non-interactive regression testing, instead of some more brittle TTY-based workaround for insufficient scriptability, which would involve waiting a bit, then sending keypresses. Some of these issues, especially #82 and #83 , are easy to implement.

I posted them because building a proper CLI for memtest86+ is _far_ less complex (and yields much smaller code) than the alternatives, e.g. reading a configuration file from a FS on an internal or external mass storage device, or getting configuration information from some form of network server. For a portable implementation of these, rewriting the memory testing as a task in an embedded RTOS would be an option...

The command line length limit is high, as 300+ characters routinely work ( https://clonezilla.org/livehd.php et. al) and we don't need that many for reasonable inputs to #7 + #58 + #72 + #82 + #83 on the same command line :)

x86fr:
- Considering that there is a demand from users for scripting that way, we should regroup all of these to a single "scripting" issue with sub-requests
- reading/writing on a FS is the ultimate solution and will be needed at some point.
- Still convert #7 (and future ones like this one) to discussion so we can get feedbacks from users with the voting system

Issue #82, opened 2022-05-21
Selecting the CPU sequencing mode through the command line


debrouxl:
For instance, currently, SEQ / PAR can be selected through the usage of `nosmp`, but RR can't, at the time of this writing.
This interacts with the CLI for #9 and #73 .

Issue #73, opened 2022-05-17
Selecting the initial set of cores / sockets / nodes through the command line


debrouxl:
At the time of this writing, only interactive configuration can select the set of cores.
I was thinking about a comma-separated list of cores similar to the one I suggested for tests in #72 , enabling both positive and negative selection.
This interacts with the CLI for #9; should we really try to build a unified CLI syntax to handle both core number selection and SMT/P-Core/E-Core enablement/disablement ?

Distant wishlist item: `socket<n>` and `node<n>` shortcuts for selecting respectively the cores of the given CPU socket and the cores of the given NUMA node, after parsing valid information from the ACPI SRAT for the latter. In the meantime, `<m>-<n>` range and `<m>,<n>`list syntaxes for selecting cores do the job for a sufficiently homogeneous set of computers.

Issue #72, opened 2022-05-17
Setting the memory address range through the command line


debrouxl:
This is another configuration option which can only set interactively, at the time of this writing.
Suggested CLI: `memrange=<low>-<high>`, both parameters mandatory.
Distant wishlist item: `memrange=node<n>` for selecting the memory address range of the given NUMA node.

PR #149, opened 2022-08-21
Optimize the JEP106 list

debrouxl:
I've finally scratched a long-time itch, already mentioned in March in a comment from #2 : optimizing an array of structs containing strings, here the `jep106` struct array, which ends up taking a sizable amount of space in the program, due to the pointer, the padding, and the relocation information.
The core idea is nothing clever, I used it on the TI-68k series 15+ years ago with the exact same goal: avoiding relocated 32-bit / 64-bit pointers by switching to 16-bit offsets.
Over the size of memtest_shared, on x86, this change saves 1+ KB (1-% of the size without BSS), and on x86_64, it saves 3.5+ KB (2.5+% of the size without BSS).

In order to keep a single list of entries, and the ability to uncomment single lines, i.e. keep the maintenance of the table as easy as with the current code, I went the external code generation route. I used Perl because I'm more familiar with Perl than with Python, and Perl is still a standard part of many Linux distros; however, the code is very simple, so it could be easily converted to Python.

NOTE: as I don't have access to my usual test computers tonight, this is not tested on real hardware yet...
However, after successful builds, I examined `build64/memtest_shared` and `build64/memtest.efi`: in the former, `objdump` shows that the data part (struct array, string data) is reasonable and that the disassembled code uses the extern labels and also looks reasonable; in the latter, `strings` shows that the manufacturer strings are part of one of the final binary products. Therefore, I definitely did not screw up to the point that the JEP106 manufacturer information printing code and data would be getting elided by the toolchain, which would save even more space :)


x86fr:
I really don't want to be rude, but we already had a debate about this : I can't find any justification to add complexity in the build process with obfuscated perl script run at makefile (against a standard c array) with the unique goal to gain a few KB in code size. 

There is no impact of a bigger code size, up to several hundreds of KB. The time needed to relocate the binary even on a 30 year old computer is under the millisecond. We really don't need to be obsessed about a few KB "lost" here and there. 

Refactoring the code to remove duplication like on the SPD PR is desirable, but because it makes the code clearer and easier to maintain, not because the binary is reduced by 1000 or 1500 bytes. We're not in Z80 era anymore :)

debrouxl:
It's not just for the sake of size optimization: reducing the size of the binary is also about increasing the ability for end users to run memtest86+ on some MBs, as shown by #49 - an issue reported by a real user. #54 / 93c9c8ded587323ff842d9ea3da6de4983c2e3b2 is a more effective workaround than optimization at the source code level, but still ;)

Also, I have a hard time finding myself guilty of obfuscating the Perl script. I mean, 8 lines (counting the } and a blank line) which reformat the information contained in an array of arrays aren't rocket science :) Generating output to stdout and taking advantage of shell output redirection was a way to keep the script bare-bones and easy to understand.
One of the ways to better document the script would be to paraphrase what the script generates, but it's easier to run the script and see what happens.

I can think of two approaches for reducing the current space waste by eliminating the padding (6 bytes per entry):
* a one-liner: using compiler-specific `__attribute__((packed))` on `const struct spd_jedec_manufacturer jep106[]`, which I'd expect to work because the x86 / x86_64 ISA supports transparent unaligned memory access;
* in pure portable C, using separate arrays for the JEDEC ID and the manufacturer name. Unless code generation is used, maintainability will suffer, though.

x86fr:
> It's not just for the sake of size optimization: reducing the size of the binary is also about increasing the ability for end users to run memtest86+ on some MBs, as shown by #49 - an issue reported by a real user. #54 / [93c9c8d](https://github.com/memtest86plus/memtest86plus/commit/93c9c8ded587323ff842d9ea3da6de4983c2e3b2) is a more effective workaround than optimization at the source code level, but still, this patch saves nearly a 4 KB memory page on x86_64 ;)

Really, about #49, when you have 3 TERABYTES of RAM and a 200 KILOBYTES binary can't fit in memory, the issue does not come from a lack of optimization in your software. The issue was from the overhead due to APs and the initial load address in low memory, and has been solved the right way. The program size by itself was not the cause of the issue and shrinking it was not the solution. 

So, there is still no justification to add complexity to the build process by involving perl to generate .s file, just to gain a KB or so. (& also, I include jedec_id.h in my debugging tools 😅)

What I don't really understand in the end is why spend time on this kind of side optimizations now rather than on the one of many problems and new features that users ask for? There is so many things to do with direct impact on users in the issues list, including many requests that you opened without active developer working on it.

debrouxl:
Indeed, we had a debate about optimization... but several smaller pure optimization commits, even some of mine, have been integrated since then, so I didn't understand that much larger optimizations such as this one, with minimal maintenance cost, wouldn't get in :)
And I could hardly have known that you use jedec_id.h in your internal tools ;)

Anyhow, I wrapped up this particular line of work, fixing the issues from the first, untested version, and introducing a first commit which adds `__attribute__((packed))`, from my second comment above. Both commits work for me on a 32-bit PIII 1 GHz with 2 x 128 MB of RDRAM on VC820 in BIOS mode, and a 64-bit P4 660 with 2 x 1 GB of DDR2 on Dell Precision 370 MB in BIOS mode as well.

Point taken about working on other, more useful requests. However:
* progress on the most of the remaining open issues I posted, #58, #72, #83 or #9 + #73 + #82 , would benefit from technical direction and/or design review on the expected CLI (I posted suggestions, or questions, in most of these issues), when you, Martin and others have time. Granted, I didn't make it explicit before.
* parsing ACPI SRAT and building up information about NUMA nodes and memory ranges is something that I'm interested in, but doing that work isn't much use without also refactoring how tests are implemented and executed in order to introduce optional NUMA awareness, which is a larger endeavour.
* I was waiting for the release to resume work on fuzzing, for instance.

x86fr:
My point of view is that optimizations added so far (#148 being a good example) improve the code readability and/or remove redundancies and/or enhance maintainability. Some of them also decrease code size as a side effect of factorization.

The one and only goal of this PR is to reduce the code size by a few KB at the expense of increased complexity, involving perl in the make process for file generation, and removing the ability to simply use jedec.h array directly. 

I don't expect the code size from the JEP106 list to grow more than 5 KB in the next decade (new RAM manufacturers are VERY rare) and we don't have any RAM size limitation, so I only see drawbacks with merging this.

x86fr:
JEP106 now packed (#335)

debrouxl:
And I have a better version of the second commit, FWIW.

PR #150, opened 2022-08-22
RFC: Introduce selftests and spd-decode tool

01e3:
As discussed, I moved the selftests into a separate branch. I'm marking this PR as a draft as I anticipate [much] more discussion. 

Please ignore the first three commits, they are already in https://github.com/memtest86plus/memtest86plus/pull/148 and are required for the sefltests.

As mentioned earlier, one of the other goal was to uncover a number of problems with the current code for DDR/DDR2/DDR3/DDR4 (note that I have not tested SDRAM, RDRAM and DDR5), where the timing reported by Memtest86+ is wrong when compared to all of: vendor's specification, decode-dimms and CPU-Z - please see the comments in [smbus_test.c](https://github.com/memtest86plus/memtest86plus/blob/bf92cea184d70dc92b6c764a6e4059b2dbc77073/selftests/smbus_test.c). 

As a bonus, I also added spd-decode tool that we discussed https://github.com/memtest86plus/memtest86plus/pull/148 in - https://github.com/memtest86plus/memtest86plus/commit/cae51bdaee04ff39c36fbfb69445f14cfb82e830, which was very trivial to implement with the existing testing code. Because of this, no additional changes were needed to smbus.c.

```
$ ./spd-decode data/DDR2-GoldenEmpire-1.spd
Loaded data/DDR2-GoldenEmpire-1.spd, size=256
Type: DDR2-800 3-5-5-14
Size: 1GiB
Vendor: Golden Empire (0x0313)
Part Number: "CL4-4-4  DDR2-800"
Capabilities:
 - ECC: No
 - XMP: 0

```

I expect it should be also possible to implemented fuzzing in a very similar fashion, perhaps just with some trivial changes to spd-decode.c?


01e3:
Created https://github.com/memtest86plus/memtest86plus/issues/151 for the timing / freq problem.

selftests/spd-decode.c
```diff
@@ -0,0 +1,120 @@
...
+
+    load_spd(argv[1]);
+    parse_spd(&spdi, 0);
+
+    if (!spdi.isValid) {
+        printf("Invalid SPD data.\n");
+        return 1;
+    }
+
+    mprintf("Type: %s-%u %u-%u-%u-%u-%u\n",
```
01e3, line 91: Ah, there is an extra "-%u" here. Will fix once making other changes.

x86fr:
Please, wait a bit before spending time on this. I already have a lot of code to do barely the same thing, much dirtier, but tested at a very wide scale with proven efficiency to detect issues.

01e3:
Ah, sure, just wanted to show how easy it was and that we can avoid making any changes to to "main" code.

x86fr:
I know :) OK, let's post here the quick&dirty code I use since many weeks.

It wasn't intended to be published because
1.  I don't think it will be used anymore as soon as the bugs are fixed. I see it as a temporary tool.
2.  the efficiency is not related to the code but to the amount of SPD dumps on which it can be tested (and I can't release that).

Again, don't expect clean code or any fancy tricks, it's very basic.  It took less than 2 hours to write but did the job well. The code is 100% independent from the main code. I put it on `debug/spd_check` but it can go anywhere. It doesn't write anything outside of its path. All the 'run' code is on a 50-line bash script :

```
#!/bin/bash

# Check if smbus.c or smbus.h changed
smd5_old="$(cat smbus.md5)"
smd5_new="$(md5sum ../../system/smbus.c ../../system/smbus.h ../../system/jedec_id.h)"

if [ "$smd5_old" != "$smd5_new" ]; then
    # grab new files and regen md5
    echo "smbus.c/h changed, rebuilding..."
    echo "$smd5_new" > smbus.md5
    cp ../../system/smbus.h . && cp ../../system/smbus.c . && cp ../../system/jedec_id.h .

    # extract parse code that need exercise from smbus.c
    func_start="$(grep -n "static spd_info parse_spd_ddr5(uint8_t slot_idx)" smbus.c | head -n 1 | cut -d: -f1)"
    func_end="$(grep -n "SMBUS Controller Functions" smbus.c | head -n 1 | cut -d: -f1)"
    func_end=$((func_end - 2))
    sed -n "$func_start"','"$func_end"'p' smbus.c > smbus_exercise.c

    # assemble the new .c
    cat smbus_header.c smbus_exercise.c smbus_footer.c > smbus_chk.c

    # compile
    gcc smbus_chk.c -o smbus_chk

    # clean
    rm smbus_exercise.c smbus_chk.c
fi

cnt=1

while [ $cnt -le 1000 ]
do
    # Get SPD dump sample from CPU-Z database. Param is DDR type (from 0 for SDR to 5 for DDR5)
    curl -D headers.txt --output dump.spd --silent https://valid.x86.fr/api/spd/4

    # Display result of SPD parsed by CPU-Z using the header we received with the binary dump
    cpuz_spd="$(sed -n 's/^Content-Description:\s\(.*\)$/\1/p' headers.txt | tr -d '\r')"
    echo "CPU-Z 2.01 : ${cpuz_spd}"

    # Display result of SPD parsed by Memtest86+
    mt_spd="$(./smbus_chk dump.spd | tr -d '\0')"
    echo -n "Memtest86+ : ${mt_spd}"

    if [ "$mt_spd" != "$cpuz_spd" ]; then
        fid="$(sed -n 's/^Content-DumpID:\s\(.*\)$/\1/p' headers.txt)"
        echo " - MISMATCH FOR ID ${fid}"
    else
        echo " - OK!"
    fi

    ((cnt++))
done
```

Very easy to read.

It just extract the block of code from `smbus.c` and put it in untouched in a new .c file called `smbus_chk.c`.

A very simple header is added before the block :

```
#include <stdio.h>
#include <stdbool.h>
#include <stdint.h>
#include "smbus.h"
#include "jedec_id.h"

uint8_t spd_bytes[1024];

static uint8_t get_spd(uint8_t slot_idx, uint16_t spd_adr)
{
    return spd_bytes[spd_adr];
}

// All untouched parse_xx() functions below
```

and a very simple "footer" is added after the code to display the results using the parse_xxx functions. It's mainly copy/paste from other functions that I was too lazy to extract properly. 

```
// All untouched parse_xx() functions before this

int main(int argc, char* argv[])
{
    uint8_t spdidx = 0;
    int i;

    spd_info curspd;

    if(argv[1] == NULL) {
        printf("No file specified\r\n");
        return 1;
    }

    // SPD content to spd_bytes array
    FILE *fptr;
    long flen;

    fptr = fopen(argv[1], "rb");
    if (fptr == NULL) {
        printf("The file %s does not exist\r\n", argv[1]);
        return 2;
    }
    fseek(fptr, 0, SEEK_END);
    flen = ftell(fptr);
    rewind(fptr);

    fread(spd_bytes, flen, 1, fptr);
    fclose(fptr);

    // Parse SPD byte array according to the type
    switch(get_spd(spdidx, 2)) {
        default:
            printf("Unable to find DRAM type\r\n");
            return 3;
        case 0x12: // DDR5
            curspd = parse_spd_ddr5(spdidx);
            break;
        case 0x0C: // DDR4
            curspd = parse_spd_ddr4(spdidx);
            break;
        case 0x0B: // DDR3
            curspd = parse_spd_ddr3(spdidx);
            break;
        case 0x08: // DDR2
            curspd = parse_spd_ddr2(spdidx);
            break;
        case 0x07: // DDR
            curspd = parse_spd_ddr(spdidx);
            break;
        case 0x04: // SDRAM
            curspd = parse_spd_sdram(spdidx);
            break;
        case 0x01: // RAMBUS - RDRAM
            if (get_spd(spdidx, 1) == 8) {
                curspd = parse_spd_rdram(spdidx);
            }
            break;
    }

    // Display all infos we need to check
    printf("%uGB %s-%i%s", curspd.module_size / 1024,
                    curspd.type,
                    curspd.freq,
                    curspd.XMP ? " XMP " : " ");

    for (i = 0; i < JEP106_CNT; i++) {
        if (curspd.jedec_code == jep106[i].jedec_code) {
            printf( "- %s ", jep106[i].name);
            break;
        }
    }

    if (curspd.jedec_code == 0) {
        printf("- Noname ");
    } else if (i == JEP106_CNT) {
        printf("- Unknown (0x%x) ", curspd.jedec_code);
    }

    for(i = 0; i < curspd.sku_len; i++) {
        printf("%c", curspd.sku[i]);
    }

    if (curspd.fab_week < 53 && curspd.fab_week != 0 &&
        (curspd.fab_year < 32 || (curspd.fab_year >= 96 && curspd.fab_year <= 99))) {
        printf(" (%02i%02i-W%02i)", curspd.fab_year >= 96 ? 19 : 20, curspd.fab_year, curspd.fab_week);
    }

    printf(" - CAS %u-%u-%u-%u", curspd.tCL, curspd.tRCD, curspd.tRP, curspd.tRAS);

    return -1;
}
```

Then the bash script download a random binary dump for a specific type of DRAM from CPU-Z and run the Memtest86+ code on the byte array as if it was from a real SPD module. A string formatted identically but parsed by CPU-Z is also retrieved from the API call headers. IE:

```
sandbox:/home/memtest/memtest86plus/debug/spd_check# cat headers.txt
HTTP/1.1 200 OK
Date: Tue, 11 Aug 2022 00:31:26 GMT
Server: Apache/2.4.25 (Debian)
Cache-Control: no-cache private
Content-Description: 8GB DDR5-4800 - Samsung M425R1GB4BB0-CQKOD (2022-W05) - CAS 40-39-39-77
Content-DumpID: um7nsp
Content-disposition: attachment; filename=dump.spd
Content-Transfer-Encoding: binary
Content-Length: 1024
Content-Type: application/octet-stream
```

Both are then compared to help find the issues.

Small log extract:

```
#1024 - CPU-Z 2.01 : 16GB DDR5-6000 XMP - Kingston KF560C40-16 (2022-W16) - CAS 40-40-40-80
#1024 - Memtest86+ : 16GB DDR5-6000 XMP - Kingston KF560C40-16 (2022-W16) - CAS 40-40-40-80 - OK!
#1025 - CPU-Z 2.01 : 16GB DDR5-5600 XMP - Corsair CMK32GX5M2B5600C36 - CAS 36-36-36-76
#1025 - Memtest86+ : 16GB DDR5-5600 XMP - Corsair CMK32GX5M2B5600C36 - CAS 36-36-36-76 - OK!
#1026 - CPU-Z 2.01 : 16GB DDR5-5600 XMP - Kingston KF556C40-16 (2022-W29) - CAS 40-40-40-80
#1026 - Memtest86+ : 16GB DDR5-5600 XMP - Kingston KF556C40-16 (2022-W29) - CAS 40-40-40-80 - OK!
#1027 - CPU-Z 2.01 : 16GB DDR5-5600 XMP - Kingston KF556C40-16 (2022-W29) - CAS 40-40-40-80
#1027 - Memtest86+ : 16GB DDR5-5600 XMP - Kingston KF556C40-16 (2022-W29) - CAS 40-40-40-80 - OK!
#1028 - CPU-Z 2.01 : 16GB DDR5-4800 XMP - Crucial CT16G48C40U5.M8A1 (2022-W04) - CAS 40-39-39-77
#1028 - Memtest86+ : 16GB DDR5-4800 XMP - Crucial CT16G48C40U5.M8A1 (2022-W04) - CAS 40-38-38-76 - MISMATCH FOR ID iumb9h
#1029 - CPU-Z 2.01 : 16GB DDR5-4800 XMP - Crucial CT16G48C40U5.M8A1 (2022-W04) - CAS 40-39-39-77
#1029 - Memtest86+ : 16GB DDR5-4800 XMP - Crucial CT16G48C40U5.M8A1 (2022-W04) - CAS 40-38-38-76 - MISMATCH FOR ID 5xkvri
#1030 - CPU-Z 2.01 : 16GB DDR5-6200 XMP - Corsair CMT32GX5M2X6200C36 - CAS 36-39-39-76
#1030 - Memtest86+ : 16GB DDR5-6200 XMP - Corsair CMT32GX5M2X6200C36 - CAS 36-39-39-76 - OK!
#1031 - CPU-Z 2.01 : 16GB DDR5-5200 XMP - Kingston KF552C40-16 (2022-W24) - CAS 40-40-40-80
#1031 - Memtest86+ : 16GB DDR5-5200 XMP - Kingston KF552C40-16 (2022-W24) - CAS 40-40-40-80 - OK!
```

To solve the issues, I have access to many other tools I wrote for CPU-Z to display the various timings tables in a easily human readable format. Ie:

```
DIMM #				1
	SMBus address		0x51
	Memory type		DDR5
	Module format		UDIMM
	Module Manufacturer(ID)	A-Data Technology (7F7F7F7FCB000000000000000000)
	SDRAM Manufacturer (ID)	Samsung (CE00000000000000000000000000)
	Size			16384 MBytes
	Max bandwidth		DDR5-6000 (3000 MHz)
	Max JEDEC		DDR5-4800 (2400 MHz)
	Channels		2
	Part number		AX5U6000C4016G-B
	Serial number		00062980
	Manufacturing date	Week 10/Year 22
	Nominal Voltage		1.10 Volts
	EPP			no
	XMP			yes, rev. 3.0
	AMP			no
JEDEC timings table		CL-tRCD-tRP-tRAS-tRC @ frequency
	JEDEC #1		22.0-23-23-43-65 @ 1333 MHz
	JEDEC #2		26.0-27-27-51-77 @ 1566 MHz
	JEDEC #3		28.0-28-28-54-82 @ 1666 MHz
	JEDEC #4		30.0-30-30-58-88 @ 1800 MHz
	JEDEC #5		32.0-33-33-62-95 @ 1933 MHz
	JEDEC #6		36.0-37-37-70-106 @ 2166 MHz
	JEDEC #7		40.0-40-40-77-117 @ 2400 MHz
	JEDEC #8		42.0-40-40-77-117 @ 2400 MHz
XMP profile			XMP-6000
	Specification		6000 CL40
	VDD Voltage		1.350 Volts
	VDDQ Voltage		1.350 Volts
	VPP Voltage		1.800 Volts
	Memory Controller Volt.	1.200 Volts
	Min Cycle time		0.333 ns (3000 MHz)
	Max CL			40.0
	Min tRP			13.32 ns
	Min tRCD		13.32 ns
	Min tRAS		25.31 ns
	Min tRC			38.63 ns
XMP timings table		CL-tRCD-tRP-tRAS-tRC-CR @ frequency (voltage)
	XMP #1			20.0-20-20-38-58-n.a @ 1501 MHz (1.350 Volts)
	XMP #2			22.0-22-22-42-64-n.a @ 1651 MHz (1.350 Volts)
	XMP #3			24.0-24-24-46-70-n.a @ 1801 MHz (1.350 Volts)
	XMP #4			26.0-26-26-50-76-n.a @ 1951 MHz (1.350 Volts)
	XMP #5			28.0-29-29-54-82-n.a @ 2102 MHz (1.350 Volts)
	XMP #6			30.0-31-31-58-87-n.a @ 2252 MHz (1.350 Volts)
	XMP #7			32.0-32-32-61-93-n.a @ 2402 MHz (1.350 Volts)
	XMP #8			34.0-34-34-65-99-n.a @ 2552 MHz (1.350 Volts)
	XMP #9			36.0-36-36-69-105-n.a @ 2702 MHz (1.350 Volts)
	XMP #10			38.0-38-38-73-111-n.a @ 2852 MHz (1.350 Volts)
	XMP #11			40.0-40-40-76-116-n.a @ 3000 MHz (1.350 Volts)
	XMP #12			42.0-40-40-76-116-n.a @ 3000 MHz (1.350 Volts)
	XMP #13			44.0-40-40-76-116-n.a @ 3000 MHz (1.350 Volts)
	XMP #14			46.0-40-40-76-116-n.a @ 3000 MHz (1.350 Volts)
	XMP #15			48.0-40-40-76-116-n.a @ 3000 MHz (1.350 Volts)
	XMP #16			50.0-40-40-76-116-n.a @ 3000 MHz (1.350 Volts)
```

This code has been run on a bit less than 10.000 SPD dump (out of millions available) and identified almost all the issues. I don't think I will need it anytime soon, but as a quick debug tool, it worked as expected. Now I still need a bit more time to fix some timing formulas, but the patch should be committed soon. 

I'm moving soon and I have a lot of stuff to do at work these weeks, so my time is a bit limited :-/

selftests/spd-decode.c
```diff
@@ -0,0 +1,120 @@
...
+    (void)slot_idx;   // Silent "unused parameter" warning
+
+    if (spd_adr < spd_size)
+        return spd[spd_adr];
+
+    fprintf(stderr, "Unbound SPD read - addr %u\n", spd_adr);
+    exit(1);
+}
+
+static char *no_yes[] = {"No", "Yes"};
```
debrouxl, line 71: const char * ?

debrouxl (commented):
I'd swap "smbus: introduce and use parse_spd" and "selftest: Introduce selftests", so that all of the changes to smbus.c are done up front, and the selftests are added in one fell swoop. Maybe "smbus: introduce and use parse_spd" could even be moved one step earlier, before the bcd_to_ui8 introduction ?

01e3:
> I know :) OK, let's post here the quick&dirty code I use since many weeks.
> 
> It wasn't intended to be published because
> 
>     1. I don't think it will be used anymore as soon as the bugs are fixed. I see it as a temporary tool.
> 
>     2. the efficiency is not related to the code but to the amount of SPD dumps on which it can be tested (and I can't release that).
> 
> 
> Again, don't expect clean code or any fancy tricks, it's very basic. It took less than 2 hours to write but did the job well. The code is 100% independent from the main code. I put it on `debug/spd_check` but it can go anywhere. It doesn't write anything outside of its path. All the 'run' code is on a 50-line bash script :

<CUT>

Wow, very nice write up, thanks! In particular, I was not aware of https://valid.x86.fr/api/spd", but it seems it is not public as I get 401 Unauthorized / "NOT AUTHORIZED"?

Now... While I agree that if the currently known SPD bugs are being fixed then the value of the tests goes down, at least short term, I still would still point out that longer term, there may be more issues discovered in the future, especially if new module types get added.

Also, the way how I implemented the SPD tests requires very minimal changes to the original files (so far, essentially one #undef for getspd in smbus.c, which if preferred, can be replaced by one #ifdef/#endif - please let me know) so the cost of having them in the main tree should be very small. Also, while I have not looked at this yet, I suspect it should be also possible to add similar tests to the remaining parts of the code accessing other data structures that we have discussed.

> I'd swap "smbus: introduce and use parse_spd" and "selftest: Introduce selftests", so that all of the changes to smbus.c are done up front, and the selftests are added in one fell swoop. Maybe "smbus: introduce and use parse_spd" could even be moved one step earlier, before the bcd_to_ui8 introduction ?

Ah sure, happy to do it this way id this is preferred. My original motivation for the current patch structure is that the parse_spd change makes most sense if we intend to test the code and/or enable accessing this from the outside (for example the spd-decode tool), so I included it here. Otherwise the change itself does not bring much value to the original code especially when compared to my other changes, where I mostly tried to remove all duplicated code and/or fix bugs.

selftests/spd-decode.c
01e3: Definitely, thanks for catching!

selftests/smbus_test.c
```diff
@@ -0,0 +1,312 @@
...
+    fprintf(stderr, "Unbound SPD read slot %u, addr %u\n", spd_slot, spd_adr);
+    exit(1);
+}
+
+void test_parse_spd_ddr4() {
+
+    spd_info spdi;
+    uint8_t slot_idx = 0;
+
+    // BUG: shuld be reported as DDR4-2133 15-15-15-36 not DDR4-2132 14-14-14-35
```
debrouxl, line 78: Typo: "shuld" (here and below)

debrouxl:
Nice mass testing setup indeed, for enabling a set of bug fixes in the SPD decoders :)
Besides a small improvement of the main testing program for the mass testing setup, this PR has immediate value for ensuring that the relevant testcases for each bug fix currently being pushed do not regress in the future, and later, with a small expansion, value for fuzzing (see below).

BTW: some other generic testing frameworks I've seen have more states for test outcomes, e.g. PASS, FAIL, XPASS (this test is expected to pass, but known to be flaky or otherwise problematic), XFAIL (this testcase is deemed correct, but expected to fail until a bug is fixed) and SKIP.
For our purposes, I can't see a relevance for XPASS and SKIP; however, the mere addition of an XFAIL-type state for some of the check_* lines corresponding to most SPD dumps added by this PR would enable documenting the expected behaviour from the get go, instead of having to fix the tests alongside the code when that is fixed. But the impending bug fixes reduce the value of XFAIL.

About fuzzing of code (preferably built with AddressSanitizer, UndefinedBehaviourSanitizer and MemorySanitizer instrumentation, one at a time): on your side, Sam, differential code fuzzing between the (private ?) CPU-Z parsing code and the memtest86+ parsing code, i.e. running both functions, comparing the output (which ought to be identical), and `abort()`-ing the execution of the fuzzer when the output diverges, is a useful (sometimes, surprisingly so) way to 1) generate and 2) pinpoint discrepancies, usually fixing bugs in at least one.
I'm less sure about cooperative fuzzing, i.e. concurrently running a fuzzer exercising CPU-Z parsing code and a fuzzer exercising memtest86+ parsing code, on the same input, and having them expand each other's corpus, bringing much value. The control flow of the parsing code is simple enough, so finding coverage for virtually all branches should be done in quick order...

selftests/smbus_test.c
01e3: Ah, yes, thanks. Luckily, they are no longer needed as the SPD parsing now works as expected for every testcase I have. :)

01e3:
I wonder if there is any appetite on getting something like this integrated one day, and if so - should I continue working on this, or should I just close this PR?

x86fr:
There are currently no reported issues regarding timings, so I don't think it's particularly useful. The timing computation formulas have always been a mess. While the JEDEC specifications are somewhat acceptable, both XMP and EXPO specifications are filled with constants and odd rounding tricks. Some values are even hardcoded, making adjustments a manual process.

01e3:
Thanks, closing.

Issue #126, opened 2022-07-22
ALi SMBus controllers not supported


InfoLibre:
When I test these 2 computers with these RAM (see joined files for RAM model and name), RAM name isn't displayed
[Description (avec carte Wi-Fi  PCMCIA D-Link).odt](https://github.com/memtest86plus/memtest86plus/files/9172314/Description.avec.carte.Wi-Fi.PCMCIA.D-Link.odt)
[Description (avec carte Wi-FI PCMCIA OvisLink).odt](https://github.com/memtest86plus/memtest86plus/files/9172315/Description.avec.carte.Wi-FI.PCMCIA.OvisLink.odt)
.

x86fr:
These laptops seem to use ATI chipsets, which are unsupported for SPD decofing (RAM specs display) at this time.

x86fr:
PS : Please confirm the chipset used by these laptop.

InfoLibre:
Here are lspci and lsusb command result files. Do you need another info?

[lspci.txt](https://github.com/memtest86plus/memtest86plus/files/9175199/lspci.txt)

[lsusb.txt](https://github.com/memtest86plus/memtest86plus/files/9175200/lsusb.txt)

debrouxl:
I didn't notice it previously, but the output of `lspci -nn` would be more accurate than the output of `lspci`, though the latter can be encoded back to numeric IDs, at least if there are no duplicates :)

InfoLibre:
```
00:00.0 Host bridge [0600]: Advanced Micro Devices, Inc. [AMD/ATI] RS100 Host Bridge [1002:cab0] (rev 13)
00:01.0 PCI bridge [0604]: Advanced Micro Devices, Inc. [AMD/ATI] RS100 AGP Bridge [1002:700f] (rev 01)
00:02.0 USB controller [0c03]: ULi Electronics Inc. USB 1.1 Controller [10b9:5237] (rev 03)
00:06.0 Multimedia audio controller [0401]: ULi Electronics Inc. M5451 PCI AC-Link Controller Audio Device [10b9:5451] (rev 02)
00:07.0 ISA bridge [0601]: ULi Electronics Inc. M1533/M1535/M1543 PCI to ISA Bridge [Aladdin IV/V/V+] [10b9:1533]
00:08.0 Modem [0703]: ULi Electronics Inc. M5457 AC'97 Modem Controller [10b9:5457]
[...]

PR #154, opened 2022-09-02
Add several PCI vendor and device IDs for SMBus controllers; minor refactor

debrouxl:
This was inspired by #126 . It's a no-op on the binary code, at least with GCC 12: the useless cases are optimized away.

I moved the PCI VID defines to pci.h, so that they can be reused by e.g. hwquirks.c; maybe the defines' names should be prefixed by "PCI_" ?

### DIFF
--- a/system/hwquirks.c
+++ b/system/hwquirks.c
@@ -79,9 +79,9 @@ void quirks_init(void)
     //  -------------------------
     //  -- ALi Aladdin V Quirk --
     //  -------------------------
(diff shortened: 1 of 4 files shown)


martinwhitaker:
I would prefer the defines to be prefixed by PCI_. One day we may also need to take USB vendor IDs into account.

debrouxl:
Very good point, improved.

x86fr:
Can be merged for 6.00 if you remove the draft flag ;)

Issue #71, opened 2022-05-16
No keyboard detected on Dell Optiplex 580, no SPD data


01e3:
I have a Dell Optiplex 580 DT where keyboard does not work on Memtest86+. It is connected to a dell KVM via a dongle, but just in case - I also plugged a "real" keyboard. Memtest86+ is booted in BIOS via LILO - this machines does not support UEFI. I also tested the official USB/ISO "GRUB" image - no changes.

Also, there is no SPI data but I think this may be WAI given the ATI/AMD SMBus controller that may not be yet supported. Just a note that SPD works on Linux:

```
[    9.003538] piix4_smbus 0000:00:14.0: SMBus Host Controller at 0xb00, revision 0
[    9.008553] piix4_smbus 0000:00:14.0: Auxiliary SMBus Host Controller at 0xb20

# i2cdetect -l|grep SMBus
i2c-6   smbus           SMBus PIIX4 adapter at 0b20             SMBus adapter
i2c-5   smbus           SMBus PIIX4 adapter at 0b00             SMBus adapter

# i2cdetect -y 5
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          -- -- -- -- -- -- -- -- -- -- -- -- --
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
30: 30 31 32 33 -- -- -- -- 38 -- -- -- -- -- -- --
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --
50: UU UU UU UU -- -- -- -- -- -- -- -- -- -- -- --
60: -- -- -- -- -- -- -- -- -- 69 -- -- -- -- -- --
70: -- -- -- -- -- -- -- --

# decode-dimms|grep ^Number
Number of SDRAM DIMMs detected and decoded: 4
```

Attaching the screenshots and the output from lspci/lsusb/lsusb.py:

[dello580-lsusb.py.txt](https://github.com/memtest86plus/memtest86plus/files/8701488/dello580-lsusb.py.txt)
[dello580-lsusb.txt](https://github.com/memtest86plus/memtest86plus/files/8701489/dello580-lsusb.txt)
[dello580-lspci.txt](https://github.com/memtest86plus/memtest86plus/files/8701490/dello580-lspci.txt)

[...]

PR #167, opened 2022-09-13
Take ownership of all USB controllers before probing for devices.

martinwhitaker:
When two controllers are attached to a physical port (e.g. in the case of EHCI and its companion controllers, problems can occur if the BIOS still has control of one controller when we try to use the other one. So perform a first pass to scan the PCI bus and take ownership of and reset all the controllers we find, and perform a second pass to initialise the controllers and probe for attached devices.

As we don't support hot plugging, split the second pass into two, with the first probing the EHCI controllers and handing over any low and full speed devices to the companion controllers, and the second probing the remaining controller types.

This fixes a problem I found on my old AMD system, where connecting an external USB1 hub to an external USB2 hub caused the EHCI controller to detect the USB2 hub as a low speed device. It also brings us more into line with the way the Linux USB subsystem works.

### DIFF
--- a/system/ehci.c
+++ b/system/ehci.c
@@ -486,7 +486,7 @@ static const hcd_methods_t methods = {
 // Public Functions
 //------------------------------------------------------------------------------
 
-bool ehci_init(int bus, int dev, int func, uintptr_t base_addr, usb_hcd_t *hcd)
+bool ehci_reset(int bus, int dev, int func, uintptr_t base_addr)
 {
     ehci_cap_regs_t *cap_regs = (ehci_cap_regs_t *)base_addr;
 
@@ -512,6 +512,15 @@ bool ehci_init(int bus, int dev, int func, uintptr_t base_addr, usb_hcd_t *hcd)
     // Ensure the controller is halted and then reset it.
     if (!halt_host_controller(op_regs)) return false;
     if (!reset_host_controller(op_regs)) return false;
+
+    return true;
+}
+
+bool ehci_probe(uintptr_t base_addr, usb_hcd_t *hcd)
+{
+    ehci_cap_regs_t *cap_regs = (ehci_cap_regs_t *)base_addr;
+
+    ehci_op_regs_t *op_regs = (ehci_op_regs_t *)(base_addr + cap_regs->cap_length);
 
     // Record the heap state to allow us to free memory.
     uintptr_t initial_heap_mark = heap_mark(HEAP_TYPE_LM_1);
--- a/system/ehci.h
+++ b/system/ehci.h
@@ -15,12 +15,35 @@
 #include "usbhcd.h"
 
 /**
- * Initialises the EHCI device identified by bus, dev, func, and base_addr,
- * scans all the attached USB devices, and configures any HID USB keyboard
- * devices it finds to generate periodic interrupt transfers that report key
- * presses. Initialises hcd and returns true if the device was successfully
- * initialised and one or more keyboards were found.
+ * If necessary, takes ownership of the EHCI device at the specified base
+ * address, then resets it.
+ *
+ * \param bus       - the PCI bus number for accessing the device
+ * \param dev       - the PCI device number for accessing the device
+ * \param func      - the PCI function number for accessing the device
+ * \param base_addr - the base address of the device in virtual memory
+ *
+ * \returns
+ * true if ownership was acquired and the device was successfully reset,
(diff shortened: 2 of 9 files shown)


martinwhitaker:
It would be interesting to test whether this change allows us to remove the workaround I added to fix issue #71. I don't have the necessary hardware to try it.

01e3:
> It would be interesting to test whether this change allows us to remove the workaround I added to fix issue #71. I don't have the necessary hardware to try it.

Ehm... that would be me. ;)

Sadly, the said HW is on a different continent than where I am currently, so it is not going to be easy.... but with some effort I can arrange testing it. How urgent is it? 

Also, I assume you want me to revert https://github.com/memtest86plus/memtest86plus/commit/3a107018e54d285208885a5b69995382b5afa999 and see what happens with this commit applied?

martinwhitaker:
@01e3 , not at all urgent. The previous fix causes all USB devices to be handed over to the companion controller(s) if no keyboards are attached via a USB >= 2 hub, which means the companion controllers then probe any USB >= 2 devices again. So if we can remove that fix, it will reduce the time taken to probe for keyboards on such systems, but otherwise it does no harm to keep it.

Yes, to test, check out this branch and revert 3a10701.

@debrouxl also has hardware that needed 3a10701, so may be willing to test.

Unless @x86fr is planning another round of beta testing, we should probably wait until after the 6.00 release before making such a change.

debrouxl:
Oops, I forgot to post this comment on Wednesday.
Yup, I meant to test this commit on several of my computers, but not before Saturday or Sunday. The tests will be incomplete this week-end, as the R815 is doing real work I can't interrupt for several more days, and one of the two RS904A is currently down because of PSU failure.

debrouxl:
Relatively quick first test on the RS904A, single run, under the following conditions:
* code: HEAD of early-usb-handoff + a manual revert of the single line from 3a10701;
* BIOS settings: "Legacy USB Support" and "BIOS EHCI Hand-Off" both Disabled;
* memtest86+ command line: `usbdebug keyboard=usb`;
* both PS/2 and USB keyboard plugged in, the former being useful in the GRUB menu under these conditions.

Result: the Logitech K120 is correctly detected by memtest86+, enumerated under the OHCI controller under the EHCI controller, and it works well enough: that is, dozens of F1 / F10 pairs in quick succession, without a hitch. Keypresses on the PS/2 keyboard are not taken into account, as requested on the command line.

Remaining tests:
- [ ] repeat runs on the RS904A under the same conditions;
- [ ] repeat runs on the RS904A with "Legacy USB Support" disabled and "BIOS EHCI Hand-Off" enabled;
- [ ] several other sufficiently recent computers with an EHCI controller.

debrouxl:
I can see two successful detections of Logitech K120 in a row on the RS904A, with the same binary as in my previous message, "Legacy USB Support" disabled, "BIOS EHCI Hand-Off" enabled, memtest.bin `usbdebug keyboard=usb` parameters. The keyboard works nominally for triggering and exiting the memtest86+ menu repeatedly.

01e3:
Was able to coordinate some testing on Dell Optiplex 580 (AMD 785G, AMD SB710), see https://github.com/memtest86plus/memtest86plus/issues/71 for more context.

All tests were done with "keyboard=usb"

Summary:
 - current git (ddbee66c854f0e88ddebaeed46011bea78526c7d): keyboard works
 - current git, but with the [workaround](https://github.com/memtest86plus/memtest86plus/commit/3a107018e54d285208885a5b69995382b5afa999) removed [1]: keyboard does not work
- current git, but with https://github.com/memtest86plus/memtest86plus/pull/167/commits/0c9d7ad8044f0846ad7b21edeaef27b208de9810: keyboard works
- current git, but with https://github.com/memtest86plus/memtest86plus/pull/167/commits/0c9d7ad8044f0846ad7b21edeaef27b208de9810 and the [workaround](https://github.com/memtest86plus/memtest86plus/commit/3a107018e54d285208885a5b69995382b5afa999) removed [1]: keyboard *still works* :)

[1]
```
@@ -645,7 +654,7 @@ bool ehci_init(int bus, int dev, int func, uintptr_t base_addr, usb_hcd_t *hcd)

     if (num_keyboards == 0) {
         (void)halt_host_controller(op_regs);
-        (void)reset_host_controller(op_regs);
+//        (void)reset_host_controller(op_regs);
         goto no_keyboards_found;
     }

```

PR #177, opened 2022-09-30
Script to debug memtest86plus (efi-version) with GDB in QEMU

koenigr:
When I was working on my own memtest-project, it was very usefull to retrace what happens in the code by debug it with gdb in qemu. I think this could help other memtest-developers to comprehend how to load the symbol-tables at the correct loading address.
This solution contains a script (debug_memtest.sh) with the whole procedure and some modifications at the Makefile and memtest_shared.lds to create a debug symbol table. More explanations and suggestions for further improvements can be found in HOW_TO_DEBUG_WITH_GDB.md

### DIFF
--- a/HOW_TO_DEBUG_WITH_GDB.md
+++ b/HOW_TO_DEBUG_WITH_GDB.md
@@ -1,0 +1,97 @@
+#		How to debug memtest.efi in QEMU with GDB
+
+`debug_memtest.sh` is a script that allows memtest86plus developers to set up a debugging environment with GDB in QEMU.  
+It calls the `make debug` target of a modified Makefile to create an additional debug-symbol file called `memtest.debug`.  
+The symbols of this file are loaded with an offset (in accordance with the loading location of the efi-image) into GDB to match the exact addresses of the symbols in memory.
+
+
+##	Prerequisites
+
+* this approach was tested on Ubuntu 18.04 - 22.04
+* the debug script is created only for the efi 64-bit version
+* qemu-system-x86_64 and ovmf must be installed
+
+##	How to run
+
+* navigate to build64 directory
+* run `./debug_memtest.sh`
+* or type `./debug_memtest.sh -h` for help
+
+### Remarks - create own gdbscript
+
+It is possible to provide an own gdb-script. Name it 'gdbscript' and place it in the build64 directory.
+This script will automatically be loaded when gdb starts.
+!! But be careful when cleaning the directory by './debug_memtest.sh -c'. It also removes 'gdbscript'.
+!! Make sure that you have made a copy of 'gdbscript' when running this command.
+
+##	Navigate inside Qemu/UEFI
+
+* wait until UEFI-Shell is loaded
+* type "fs0:" - Enter
+* type "memtest.efi" - Enter
+
+###	Inside GDB
+
+When GDB is running, it stops at the first breakpoint at main(). Feel free to add further breakpoints or continue with `c`.
+
+###	Remarks - auto-boot memtest86+
+
+In step **Navigate inside QEMU/UEFI**, you have to navigate to the directory which contains memtest.efi and manually launch it.
+
+If you want to automatically boot from memtest.efi, there is an additional step required to add memtest to the first place at the bootorder:
+
+When the UEFI Shell is running, type
+		`bcfg boot add 0 FS0:\EFI\boot\BOOT_X64.efi "memtest"`
+and confirm with Enter.
+The directory "\EFI\boot" and the file "BOOT_X64.efi" are automatically
+created by the debug-script.
+
+When you run the script the next time, memtest.efi should run without
+previous user interaction.
+
+!! But be careful when cleaning the directory by './debug_memtest.sh -c'. It also removes this setting.
+!! Make sure that you have made a copy of 'OVMF*'-files when running this command.
+
+##	Clean directory
+
+'debug_memtest.sh' has an own clean procedure which cleans additional files not mentioned in Makefile's
+'make clean' target. When you run this command, make sure that you have saved 'gdbscript' and/or OVMF* files if there are custom changes.
+
+To clean the directory, type `./debug_memtest.sh -c`
+
+##	Possible features/alternatives and further considerations
+
+###	Detection of Image Base 
+
+To assign the correct address for all debugging symbols, it is neccessary to add an offset to the values in memtest.debug (the file containing the debug symbols). This offset consists of the IMAGE_BASE and the BASE_OF_CODE.  
+Both values are defined in `memtest86plus/boot/header.S` 
+
+* IMMUTABILITY OF ALL CONDITIONS
+
+if you assume, that these values will never change during the development phase of memtest86plus AND memtest.efi is always loaded at this preferred address in qemu-system-x86_64 (which seems to be the case) then it is possible to hardcode the offset in the script (for the implementation see debug_memtest_simple.sh)
+
+* ADAPTABILITY TO DEVELOPMENT CHANGES
+
+if there is a chance, that these values WILL change during the development phase but memtest.efi is always loaded at this preferred address then the value can be read from header.S by the debug script just right before starting the debugging (for an example, see debug_memtest_full.sh)
+
+* EXPECTED ERRATIC BEHAVIOUR OF QEMU
+
+If it is expected that memtest.efi is NOT always loaded at the same address, it is inevitable to determine the actual loading address first. This approach comprises a DEBUG-build of OVMF.fd (which requires the cloning of the whole edk2-repository and manually build OVMF.fd). With this DEBUG-version of OVMF.fd it is possible to write the loading addresses of all modules into a debug.log.
+This proceeding has been tested successfully but is actually not implemented in one of the srcipts.
+
+###	Handle relocation of memtest
+
+memtest86plus relocates itself during the test procedures. As the script loads the symbol table with a given offset, debugging is only possible when the code is located at the original position. There are several ways to deal with relocation:
+
+* IGNORE RELOCATION
+
+Just ignore the fact that at a part of the time the symbols are not recognized by gdb as gdb has no information about the symbols, when memtest86plus has been relocated. It is still possible to debug the code since memtest86plus jumps sooner or later back to the original position and all precedures which are executed at one location are also executed at the other position.
+BUT: If a bug is position-dependent (i.e. it occurs only at the relocated position), you are not able to debug it.
+
+* DISABLE RELOCATION
+
+TODO: Is it possible to deactivate relocation? E.g. by outcommenting some code or setting a flag? Does it have benefits over the first approach?
+
+* FOLLOW RELOCATION
+
+If the position after relocation is expected to be always the same, then you can just load the symbol table twice. This is done in debug_memtest_simple.sh (the offsets are 0x201000 and 0x400000). If the locations can vary then the offsets must be determined dynamically ... todo: how?
--- a/build64/Makefile
+++ b/build64/Makefile
@@ -10,6 +10,18 @@ else
   GIT_AVAILABLE = true
 endif
 
+ifeq ($(DEBUG), 1)
+<<<<<<< HEAD
+  CFLAGS+=-ggdb3 -DDEBUG_GDB
+  OPT_SMALL=-Og
+  OPT_FAST=-Og
+  MS_LDS=memtest_shared_debug.lds
+else
+  OPT_SMALL=-Os
+  OPT_FAST=-O3
+  MS_LDS=ldscripts/memtest_shared.lds
+endif
+
 CFLAGS = -std=c11 -Wall -Wextra -Wshadow -m64 -march=x86-64 -mno-mmx -mno-sse -mno-sse2 \
          -fpic -fno-builtin -ffreestanding -fomit-frame-pointer -fno-stack-protector
 
@@ -73,6 +85,14 @@ OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(IMC_OBJS) $(LIB_OBJS) $(TST_
 
 all: memtest.bin memtest.efi
 
+check:
+	@if [ -z ${DEBUG} ]; then\
+		echo "Macro DEBUG is not defined. Run debug_memtest.sh to invoke debug target";\
+		exit 1;\
+	fi
+
+debug: check memtest.debug memtest.efi
+
 -include boot/efisetup.d
 -include $(subst .o,.d,$(SYS_OBJS))
 -include $(subst .o,.d,$(IMC_OBJS))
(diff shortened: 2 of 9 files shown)


debrouxl:
Until now, I have always tested memtest86+ on real computers, and I knew it was possible to use QEMU, but I didn't have a procedure for that... so I consider this to be useful documentation and scripting to have :)

martinwhitaker:
I've used qemu (and VirtualBox) a lot during the development of PCMemTest and Memtest86+, but I find it quicker and easier to insert a few print calls than to use gdb. I did try using qemu + gdb to track down one particularly stubborn bug in the startup code, but as gdb didn't track the changes between real/protected/long mode, it wasn't that helpful. But I'm not against adding this script if other people will find it useful.

build64/Makefile
```diff
@@ -66,6 +213,8 @@ APP_OBJS = app/badram.o \
 
 OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(LIB_OBJS) $(TST_OBJS) $(APP_OBJS)
 
+debug: clean memtest.debug memtest.efi
+
```
martinwhitaker, line 217: The `debug` target should come after the `all` target, to keep `all` as the default.

build64/Makefile
```diff
@@ -12,6 +14,151 @@ endif
...
          -fpic -fno-builtin -ffreestanding -fomit-frame-pointer -fno-stack-protector
 
+ifeq ($(DEBUG_S), 1)
+  CFLAGS+=-ggdb3
+endif
+
+# In Progress. This option must not be activated as the right flags are not chosen yet.
+ifeq ($(DEBUG_OPT), 1)
+  OPT = -Og
+  OPT_SYSTEM = \
```
martinwhitaker, line 24: I agree with @debrouxl , we need to make the code compile with -Og instead of doing this.

build64/ldscripts/memtest_shared.lds
```diff
@@ -52,5 +52,4 @@ SECTIONS {
 		. = ALIGN(256);
 		_end = .;
 	}
-	/DISCARD/ : { *(*) }
```
martinwhitaker, line 55: This adds ~12KB to the final binary size even without the -Og option, so I'd rather it wasn't done on a non-debug build.

build64/Makefile
```diff
@@ -66,6 +213,8 @@ APP_OBJS = app/badram.o \
 
 OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(LIB_OBJS) $(TST_OBJS) $(APP_OBJS)
 
+debug: clean memtest.debug memtest.efi
```
martinwhitaker, line 216: Do you really want to do a clean build every time?

build64/ldscripts/memtest_shared.lds
debrouxl: It's due to .eh_frame and .comment, the latter being tiny. To keep a single linker script, they could still quit being discarded, but be stripped out individually by `objcopy`.

debrouxl:
Two more ideas:
* many code bases build with debugging flags, then strip / split the debugging information after linking. There's a build time penalty in doing so, but the slowdown shouldn't be _that_ bad for a freestanding C code base like memtest86+, as opposed to some C++ code bases in which every single file includes dozens of thousands of lines of complicated non-modular headers. I once significantly optimized the build of a code base by switching from convenient `#include <QtCore>` to `#include <QString> #include <QList>...` - just the select few classes that the code actually used. That code base had a lower build-to-test time than the mere fact of installing a new memtest86+ build and rebooting the physical server to test memtest86+.
* an alternate way to write $(OPT) would be to use
```
ifeq ($(DEBUG_S), 1)
  OPT_SMALL=-Og
  OPT_FAST=-Og
else
  OPT_SMALL=-Os
  OPT_FAST=-O3
endif
```
and use these instead of hard-coded -Os / -O3 conditionally overridden by $(OPT) later.

koenigr:
Thank you for all of your feedback. I will react to the single comments /suggestions step by step during the next days

build64/Makefile
koenigr: You are right. I have changed it

build64/Makefile
koenigr: Thank you for the remark. It was indeed not the intention to do a clean build every time. I just added it at the end to make sure that the whole script runs from the beginning without errors. I have removed it now

koenigr:
I have adapted your suggestions in the following way:

1. Build with level -Og
I added a preprocessor directive -DDEBUG_GDB to CFLAGS. If it is defined (in a debug build), then memset and memcpy are implemented in string.c. Otherwise they are used as builtin functions. So there are no changes in a non-debug build.

2. Supported terminals
Now, x-terminal-emulator, gnome-terminal and xterm are supported. If the user wishes to use an other terminal (to not to be forced to install one the supported three), it is possible to pass an own terminal by ./debug_memtest.sh -t <terminal> <exec_cmd>.
There is also a help information callable by ./debug_memtest.sh -t help

3. Discard sections
It seems that it does not help to strip .eh_frame and .comment manually. I still have a big difference in the size. I just have some rough ideas but can't really explain why it is as it is. So the new solution is that I let /DISCARD/ in the original linkerscript and create a temporal linkerscript without /DISCARD/ when executing a debug build. So here is also no change to the original non-debug build

Open tasks:
1. One question is if it would be better to add the files created by the script to "make clean" instead of deleting them by "./debug_memtest.sh -c"

2. As soon as there are no more suggestions for improvement, I would finish the documentation/comments and resolve some small remaining "todos" in the code

HOW_TO_DEBUG_WITH_GDB.md
```diff
@@ -0,0 +1,99 @@
...
+
+If the position after relocation is expected to be always the same, then you can just load the symbol table twice. This is done in debug_memtest_simple.sh (the offsets are 0x201000 and 0x400000). If the locations can vary then the offsets must be determined dynamically ... todo: how?
+
+###	Build process
+
+The creation of the debug symbol table is realized by calling `make debug` with a modified Makefile. If a modification of the Makefile is not desired, it is also possible to imitate the build procedure directly in the debug-script. But this means to adjust the build procedure every time when the procedure of the original Makefile changes.
+
+###	Alternatives to gnome-terminal
+
+Todo: add alternatives to gnome-terminal, if gnome-terminal is not installed
```
debrouxl, line 99: This is partially fixed now.

build64/debug_memtest.sh
```diff
@@ -0,0 +1,201 @@
...
+        \?) # invalid option
+            Info
+            exit;;
+	esac
+
+done
+
+Check() {
+	# Check if QEMU and OVMF is installed
+	if ! dpkg -l qemu-system* > /dev/null 2>&1; then
```
debrouxl, line 77: Wouldn't a check based on `command -v` be enough here ? That would be much more portable.

build64/debug_memtest.sh
```diff
@@ -0,0 +1,201 @@
...
+		echo "Qemu not installed"
+		exit 1
+	fi
+
+	if ! dpkg -l ovmf > /dev/null 2>&1; then
```
debrouxl, line 82: Checking for the existence of *.fd files in /usr/share/OVMF would avoid querying a package manager and therefore gain portability. AFAICS, all of the Debian, CentOS / RHEL / Fedora and Alpine packages contain files matching that pattern.

build64/debug_memtest.sh
koenigr: Yes, that's right. I have added this some time before the other checks. Now all checks are consistent

build64/debug_memtest.sh
koenigr: this is also fixed now

HOW_TO_DEBUG_WITH_GDB.md
koenigr: I have made some little changes to this file. As for me, it is the final version now

koenigr:
I have marked this pull request as ready for review. Let me know if there are any more suggestions for improvement

debrouxl:
Works for me after manually rebasing the combined diff, which yields the following diff:
```
--- 177.diff.orig       2023-01-17 12:27:19.433828139 +0100
+++ 177.diff    2023-01-17 12:31:15.919836680 +0100
@@ -8,7 +8,7 @@
  
 -# Generated file
 +# Generated files
- githash.h
+ build_version.h
 +gdbscript
 +memtest_shared_debug.lds
  
@@ -171,7 +171,7 @@
  -include boot/efisetup.d
  -include $(subst .o,.d,$(SYS_OBJS))
  -include $(subst .o,.d,$(LIB_OBJS))
-@@ -87,27 +107,27 @@ boot/%.s: ../boot/%.S ../boot/boot.h
+@@ -86,27 +106,27 @@ boot/%.s: ../boot/%.S ../boot/boot.h
  
  boot/efisetup.o: ../boot/efisetup.c
        @mkdir -p boot
@@ -198,14 +198,14 @@
 -      $(CC) -c $(CFLAGS) -O3 $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
 +      $(CC) -c $(CFLAGS) -O3 $(INC_DIRS)  -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
  
- app/%.o: ../app/%.c app/githash.h
+ app/%.o: ../app/%.c app/build_version.h
        @mkdir -p app
 -      $(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
 +      $(CC) -c $(CFLAGS) $(OPT_SMALL)  $(INC_DIRS)  -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
  
- app/githash.h: FORCE
+ app/build_version.h: FORCE
        @mkdir -p app
-@@ -126,13 +146,18 @@ FORCE:
+@@ -127,13 +147,18 @@ FORCE:
  # Link it statically once so I know I don't have undefined symbols and
  # then link it dynamically so I have full relocation information.
  
```

Looks good enough for me after you add a newline at the end of build64/debug_memtest.sh :)

Passing SMP flags to `qemu-system-x86_64` works as it should. I noticed that the debugger not catch execution when passing `-cpu host -accel kvm` to `qemu-system-x86_64`, but this should not prevent from merging this PR.

Technically, we _could_ move -ggdb3 to CFLAGS to unconditionally build debug info, and strip the debug sections for the non-debug binaries, with a build time penalty... but most of the time, memtest86+ is not being debugged, so it's usually a waste of developer time.

Sam, Martin ?

build64/debug_memtest.sh
```diff
@@ -0,0 +1,222 @@
...
+# Build
+Make
+
+# Create needed directories and move efi binary to appropriate location
+Prepare_Directory
+
+# Run QEMU and launch second terminal,
+# wait for connection via gdb
+$TERMINAL gdb -x $GDB_FILE &
+$QEMU $QEMU_FLAGS -s -S
```
debrouxl, line 222: Newline at EOF ?

build64/debug_memtest.sh
```diff
@@ -0,0 +1,222 @@
...
+	BASEOFCODE=0x1000
+	RELOCADDR=0x400000
+
+	printf -v OFFSET "0x%X" $(($IMAGEBASE + $BASEOFCODE))
+
+
+	GDB_FILE="gdbscript"
+
+	# Check if gdbscript exists. If not, create one.
+    if [ ! -f $GDB_FILE ]
```
debrouxl, line 145: This script is indented by a mix of tabs and spaces.

lib/string.c
```diff
@@ -43,6 +43,17 @@ int memcmp(const void *s1, const void *s2, size_t n)
     return 0;
 }
 
+#ifdef DEBUG_GDB
+void *memcpy (void *dest, const void *src, size_t len)
+{
+  char *d = dest;
```
debrouxl, line 49: 4 spaces indentation, like in the rest of the code, please :)

lib/string.c
```diff
@@ -125,4 +146,4 @@ char *itoa(int num, char *str)
     reverse(str);
 
     return str;
-}
+}
```
debrouxl, line 149: Newline at EOF, please, though newer GCC versions do no longer warn about the lack thereof :)

koenigr:
Hi,

I have rebased this fork onto the last state original memtest86plus:main branch. So there should be no merge-conflicts. Additionally, I have corrected the inconsisted identation in the script.

To this comment:
> Ah, I noticed one usability issue: code breakpoints work on the relocated addresses, but data addresses do not. For instance, I ran `p acpi_config`: that printed the memory at 0x29e40, instead of 0x229e40.
I have made numerous test and I realized the following:

Indeed, there was a little part missing. To the command  `add-symbol-file` I have add the addresses of the respective data sections `-s .data $DATAOFFSET"`

BUT:
It only shows you the address with the first loaded Offset. You can switch both lines
`        echo "add-symbol-file memtest.debug $OFFSET -s .data $DATAOFFSET" >> $GDB_FILE
        echo "add-symbol-file memtest.debug $RELOCADDR -s .data $RELOCDATA" >> $GDB_FILE` 
to either get the variable addresses at the loading address 0x200000 or 0x400000

AND a much bigger "BUT":
the given example with acpi_config is a very special case. It is declared as extern and initialized to {0,0,0.....,false}. I absolutley have no clue for the reason and how to solve it, but it seems to be not possible to get addresses for this combination.

Try it out:
set the line 92 of acpi.c from 
`acpi_t acpi_config = {0, 0, 0, 0, 0, 0, 0, false};`
to
`acpi_t acpi_config = {1, 0, 0, 0, 0, 0, 0, false};`

and voila ... you get an address with the offset

koenigr:
Could you please check, if the additions for 32-bit support work on our system/environment or if I have forgotten to include something? When I install ovmf-ia32 on my system, I get OVMF32_CODE_4M.secboot.fd instead of OVMF32_CODE.fd, and what I get seems to not work with the debug_memtest.sh script...

debrouxl:
Exactly, users of the Debian ovmf-ia32 packages get an incomplete set of edk2 UEFI BIOS images, where only the secure boot version is available. Last time I had a look at the matter, it seemed that a simple change in debian/rules ought to fix this (adding entries to the "OVMF32_IMAGES  :=" line), but I need to check for real.

debrouxl:
Martin, Sam: any feedback before I merge this ? I wouldn't have been able to work on NUMA support or the badram reporting fix using pure physical hardware.

x86fr:
I have really no experience working with GDB on Memtest86+, so if you think it's ready to merge without any impact on real-world platforms needing tests, feel free to do it.

debrouxl:
Common code is untouched by this PR. The addition and usage of `OPT_SMALL`, `OPT_FAST` and `MS_LDS` in the Makefiles are no-ops, and when `DEBUG_GDB` is not defined, the changes to `lib/string.*` are no-ops as well. 
Also, FWIW, the SHA-256 hashes of the binaries produced before and after this change are identical on my Debian sid amd64 environment - after hex-editing the string containing the Git commit hash prefix to the same value on both sides, that is.

A triple-check is always welcome nevertheless :)

martinwhitaker:
Finally found some time to look at this. A couple of comments:

1. I think HOW_TO_DEBUG_WITH_GDB.md should live in the doc directory, not the top level.
2. The script uses OVMF.fd, OVMF_CODE.fd, and OVMF_VARS.fd. On my system only the latter two exist. Searching for an explanation, I found https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=764918#47, which explains that OVMF.fd is just the concatenation of OVMF_CODE.fd and OVMF_VARS.fd, and states "In addition, the unified OVMF.fd should not be provided at all, if possible; it only leads to confusion.". FWIW I just use `-bios /usr/share/edk2/ovmf/OVMF_CODE.fd` when using qemu to debug memtest86+ and ignore OVMF_VARS.fd.

Issue #58, opened 2022-04-30
Selection of initial set of tests through the command line


debrouxl:
This came up in #57 : non-interactive selection of tests through the command line without modifying and recompiling the program.

EDIT: suggested CLI:
* `tests=<m>(,<n>,<p>...)`: enable only tests numbers m, (n, p...);
* `tests=-<m>(,-<n>,-<p>...)`: enable all tests but tests numbers m, (n, p...);
* `tests=<m>-<n>(,<p>-<q>...)`: enable only tests numbers m to n inclusive, (p to q inclusive...);
* `tests=-<m>-<n>(,-<p>-<q>...)`: enable all tests but tests numbers m to n inclusive, (p to q inclusive...);
* combinations of single positive tests and positive test ranges, negative tests and negative test ranges.
In case of an error, e.g. empty resulting set of tests: enable all tests.

x86fr:
> `tests=<m>(,<n>,<p>...):` enable only tests numbers m, (n, p...); 

Looks the most easy to understand.

IMO, before adding all these options to config.c, we should explore the possibility of using UEFI Boot services to read a simple configuration file on the boot media early in the startup process.

x86fr:
Added in #570

PR #186, opened 2022-10-15
CLI improvements, part 1

debrouxl:
This fixes the two easiest scriptability issues I posted several months ago, providing slightly better functionality at a much lower development cost than reading configuration from a filesystem on mass storage or network, as indicated in #83 .
This doesn't have to be merged before 6.00.

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -49,7 +49,7 @@
 
 static const char spin_state[NUM_SPIN_STATES] = { '|', '/', '-', '\\' };
 
-static const char *cpu_mode_str[] = { "PAR", "SEQ", "RR " };
+static const char cpu_mode_str[3][4] = { "PAR", "SEQ", "RR " };
 
 //------------------------------------------------------------------------------
 // Private Variables
--- a/app/config.c
+++ b/app/config.c
@@ -185,6 +185,16 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "rr", 3) == 0 || strncmp(params, "one", 4) == 0) {
             cpu_mode = ONE;
         }
+    } else if (strncmp(option, "reportmode", 11) == 0) {
+        if (strncmp(params, "none", 5) == 0) {
+            error_mode = ERROR_MODE_NONE;
+        } else if (strncmp(params, "summary", 8) == 0) {
+            error_mode = ERROR_MODE_SUMMARY;
+        } else if (strncmp(params, "address", 8) == 0) {
+            error_mode = ERROR_MODE_ADDRESS;
+        } else if (strncmp(params, "badram", 7) == 0) {
+            error_mode = ERROR_MODE_BADRAM;
+        }
     } else if (strncmp(option, "keyboard", 9) == 0 && params != NULL) {
         if (strncmp(params, "legacy", 7) == 0) {
             keyboard_types = KT_LEGACY;
--- a/app/config.c
+++ b/app/config.c
@@ -175,7 +175,17 @@ static void parse_option(const char *option, const char *params)
 {
     if (option[0] == '\0') return;
 
-    if (strncmp(option, "keyboard", 9) == 0 && params != NULL) {
+    if (strncmp(option, "console", 8) == 0) {
+        parse_serial_params(params);
+    } else if (strncmp(option, "cpuseqmode", 11) == 0) {
+        if (strncmp(params, "par", 4) == 0) {
+            cpu_mode = PAR;
+        } else if (strncmp(params, "seq", 4) == 0) {
+            cpu_mode = SEQ;
+        } else if (strncmp(params, "rr", 3) == 0 || strncmp(params, "one", 4) == 0) {
+            cpu_mode = ONE;
+        }
+    } else if (strncmp(option, "keyboard", 9) == 0 && params != NULL) {
         if (strncmp(params, "legacy", 7) == 0) {
             keyboard_types = KT_LEGACY;
         } else if (strncmp(params, "usb", 4) == 0) {
@@ -183,16 +193,6 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "both", 5) == 0) {
             keyboard_types = KT_USB|KT_LEGACY;
         }
-    } else if (strncmp(option, "powersave", 10) == 0) {
-        if (strncmp(params, "off", 4) == 0) {
-            power_save = POWER_SAVE_OFF;
-        } else if (strncmp(params, "low", 4) == 0) {
-            power_save = POWER_SAVE_LOW;
-        } else if (strncmp(params, "high", 5) == 0) {
-            power_save = POWER_SAVE_HIGH;
-        }
-    } else if (strncmp(option, "console", 8) == 0) {
-        parse_serial_params(params);
     } else if (strncmp(option, "nobench", 8) == 0) {
         enable_bench = false;
     } else if (strncmp(option, "nobigstatus", 12) == 0) {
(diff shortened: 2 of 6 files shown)


x86fr:
I fear that we will have the same debate about [#7439703](https://github.com/memtest86plus/memtest86plus/pull/186/commits/74397034c38004b46dc8d0c7d8de01a6280574e7) as we previously had.

IMO, this breaks the coding style and add unwanted complexity to the code hierarchy for an insignificant ~300 bytes gain in the code size. I don't strictly adhere to the 'Clean Code' orthodoxy, but finding `memcmp()` code in a string header looks odd to me. 

Maybe we should let @martinwhitaker decide if this is compatible with the (great!) coding style he originally defined, or if this deserves some optimizations for a better code hierarchy.

debrouxl:
In fact, my work on building memtest86+ with LTO, which I pushed yesterday on a new branch (*), was started in the same time frame as this commit, being another way to potentially help with the efficiency of simple string operations without moving the implementation of these functions to the header (for the same reason why I moved several others previously with similar size gains, with Martin merging the changes at the time, FTR: calls to out of line functions instead of the likes of `rep cmpsb` or simple arithmetic operations suck) :)

Still - moving the string functions to the header is a much simpler, yet pretty low maintenance (no API or ABI change), way to offset the cost of new minor features added by the two subsequent commits. The benefits roughly scale with the number of occurrences of these functions, too.

*: I failed to make the LTO build work for a dynamic binary, and gave up, so I published the raw changes on the "lto" branch. The current set of changes includes commenting out a bunch of seemingly unrelated global definitions, as a way to try and avoid linker errors on these symbols, only to hit the next blocker, until a more serious error comes.
LTO has an impact on build time, and compiler reliability in LTO mode used to be less than stellar (though it shouldn't be an issue nowadays), so it's not necessarily something you or Martin would accept in the end anyway, despite the size savings on the static binary, which make this optimization commit, or even my string array optimization, pale.

x86fr:
@martinwhitaker If you pass by here, can you give us your opinion about commit 6d11069 (the move of the functions, including memcmp, to string header) ?

app/display.c
```diff
@@ -49,7 +49,7 @@
 
 static const char spin_state[NUM_SPIN_STATES] = { '|', '/', '-', '\\' };
 
-static const char *cpu_mode_str[] = { "PAR", "SEQ", "RR " };
+static const char cpu_mode_str[3][4] = { "PAR", "SEQ", "RR " };
```
x86fr, line 52: Just for my knowledge, what's the point of doing this? Is it just to avoid the array of pointers and reduce the size of the code by 12 bytes in 64-bit mode?

debrouxl: Correct, the point is to avoid the array of pointers (for storing strings which were already padded to the same length) and therefore especially the relocations for these pointers, which take up significantly more space than the pointers and strings:

```
git log --oneline | head -1; cd build32; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..; cd build64; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..
6d11069 Move more of the simple string functions to the header, to allow inlining and further optimization.
  10105      19    2712   12836    3224 app/config.o
 106580   26720   13344  146644   23cd4 memtest_shared
   8653      19    2712   11384    2c78 app/config.o
 110969   28392  294688  434049   69f81 memtest_shared
  7 .rela.data    00001368  0000000000019e78  0000000000019e78  0001ae78  2**3
  8 .rela.got     000000c0  000000000001b1e0  000000000001b1e0  0001c1e0  2**3

git log --oneline | head -1; cd build32; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..; cd build64; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..
88843af Add support for configuring the CPU sequencing mode through the command line. Reorder tests in app/config.c::parse_option alphabetically. Fixes #82.
  10266      19    2712   12997    32c5 app/config.o
 106705   26720   13344  146769   23d51 memtest_shared
   8811      19    2712   11542    2d16 app/config.o
 111059   28360  294720  434139   69fdb memtest_shared
  7 .rela.data    00001320  0000000000019f18  0000000000019f18  0001af18  2**3
  8 .rela.got     000000c0  000000000001b238  000000000001b238  0001c238  2**3
```

On the debrouxl/experiments branch, there's a commit containing this change and two other one-liners switching from array of pointers to strings which were already padded to the same length to an array of fixed-length strings. The combined effect is stronger:
```
git log --oneline | head -1; cd build32; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..; cd build64; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..
d1c332b Further optimize the JEP106 list by switching from relocated string pointers to offsets from the beginning of a string table.
  10374      19    2712   13105    3331 app/config.o
 106377   25636   13408  145421   2380d memtest_shared
   8734      19    2712   11465    2cc9 app/config.o
 108873   26256  294784  429913   68f59 memtest_shared
  7 .rela.data    000006d8  000000000001a2c0  000000000001a2c0  0001b2c0  2**3
  8 .rela.got     000000d8  000000000001a998  000000000001a998  0001b998  2**3

git log --oneline | head -1; cd build32; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..; cd build64; make clean &> /dev/null; make -j &> /dev/null; size app/config.o memtest_shared; objdump -x memtest_shared | grep '\.rela\.'; cd ..
1b339de Optimize two arrays of strings and an array of structs containing same-, fixed-length strings by switching to fixed-length char arrays, which avoids pointers + relocations.
  10374      19    2712   13105    3331 app/config.o
 105792   25988   13312  145092   236c4 memtest_shared
   8731      19    2712   11462    2cc6 app/config.o
 107992   26512  294784  429288   68ce8 memtest_shared
  7 .rela.data    00000528  000000000001a100  000000000001a100  0001b100  2**3
  8 .rela.got     000000d8  000000000001a628  000000000001a628  0001b628  2**3
```

I did disassemble the binary to check the generated code and data :)

martinwhitaker:
There are plenty of precedents for including small functions in the headers (e.g. in spinlock.h), so I can't really object to adding more. The disadvantage is that it obscures the API, but Doxygen helps with that.

debrouxl:
When this one is merged, and you're less busy with Secure Boot testing, could you give some design feedback for the desired CLI on #58, #72 and #9 + #73 ? :)

Selection of cores and memory ranges is fairly important to me:
* on the maintainer side, I've already used the interactive selection quite a number of times on my 4S Opterons, for benchmarking purposes;
* real-world users could non-interactively disable SMT (or enable it, if/when we flip the default), or work around the current lack of NUMA awareness in memtest86+ and yield faster semi-automated tests, especially with the help of PXE boot. Per the figures in https://github.com/memtest86plus/memtest86plus/discussions/12#discussioncomment-3745454 , we can expect to test memory over twice faster on that model (<10h vs. ~24h) by testing one NUMA node at a time, even when taking into account the fact that the server shall boot 9 times instead of 1 (8 NUMA nodes + a 9th boot to run tests 0-3 with all cores on the entire physical memory range).

PR #206, opened 2022-11-10
add autoreboot=n commandline option for automation

g-v-egidy:
Automatically reboot after n successful passes. Memtest will continue to run when a failure is detected.

This can be used to automate running memtest.

As previously discussed in https://github.com/memtest86plus/memtest86plus/discussions/138

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,11 @@ recognised:
       * 54600
       * 115200 (default if not specified or invalid)
       * 230400
+  * autoreboot=*n*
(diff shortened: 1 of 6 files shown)


01e3:
Note that it is pretty common for system with HW problems to reboot on its own.

With "reboot when OK", how one can distinguish between "my system is fine" vs "Memtest86+ reported hundreds of errors and then the system rebooted"?

x86fr:
These changes are already present in #207. Please only create one PR.

PR #207, opened 2022-11-10
add autoreboot=n and rtc_result commandline options for automation 

g-v-egidy:
autoreboot=n option:
Automatically reboot after n successful passes. Memtest will continue to run when a failure is detected.

rtc_result option:
When set and an automatic reboot due to the "autoreboot" option should be triggered,
the date in the hardware real time clock (RTC) is set to 2012-01-01. This acts
as indicator that the memtest really was successful and the reboot did not occur
due to faulty hardware, power loss or similar things.

Both options are designed as building blocks for automating memtest.

Includes extensive documentation about automating Memtest86+ in README_automation.md
  
As previously discussed in https://github.com/memtest86plus/memtest86plus/discussions/138

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,11 @@ recognised:
       * 54600
       * 115200 (default if not specified or invalid)
       * 230400
+  * autoreboot=*n*
+    * automatically reboot after *n* successful passes. memtest will continue to
+      run when a failure is detected. This can be used to automate running memtest.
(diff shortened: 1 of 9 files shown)


x86fr:
While nice in specific controlled test environments, this can't be considered for main as potential impact on the target computer is too high.

Issue #155, opened 2022-09-03
"address test, own address" behavior difference between 32-bit and 64-bit versions


01e3:
Forked from https://github.com/memtest86plus/memtest86plus/issues/153

In the "address test, own address" the "top" half has all zeros (00000000-b3f99d00), which is partially due to the 64-bit binary using 64-bit variables.

Would we have a better coverage if we put there something else then just zeros? Most systems have xx GiBs of RAM, but even for ones with z TiB we will still have soo many zeros. 

As the behavior is different compared to the 32 bit binary, where we use 32-bit variables, perhaps the best way would be to only use the "bottom" part of the address in 32-bit variables - in a consistent manner between the 32bit and 64 bit versions?

Also the value  does not match the address (00000006 73f99d00) - we have 0xb instead of 0x7 in the highest bits of the "bottom" 32 bit number, and for the "top" part of the value it is 0x0 instead of 0x6? 

_Originally posted by @01e3 in https://github.com/memtest86plus/memtest86plus/issues/153#issuecomment-1234965355_

x86fr:
Main question : is this still present in b3 ? 

I can theoretically imagine a race condition between e-core and p-core but this should not happen anymore as e-core are now skipped. 

About filling the top bits of address with something (usually 0xAA or 0x55) to create a "mixed" test, I don't think it will be really useful. Address tests are really basic and quick tests. Using the address as the "test pattern" make it easier to debug. if any issue, it will be detected by moving inversion quickly after.

martinwhitaker:
There are two "own address" tests, tests 1 and 2. Test 1 comes from Memtest86+ v5 and works one (1GB) test window at a time. I added test 2 to quickly identify faults in the address lines above 1GB. Two different tests are needed for 32-bits, because you can't uniquely tag more than 4G 32-bit words. But for 64-bits we could speed up testing by amalgamating these two tests. I didn't do that because:

1. I wanted to minimise the differences between the 32-bit and 64-bit code, to ease maintenance
2. I only deal with low core-count machines, and these tests are fast in that case

But because these tests are run sequentially on all cores, they take far more time on high core-count machines. So it might be worth implementing an amalgamated 64-bit version. Something to consider for v6.1.

01e3:
Right, I agree that 0xAA or 0x55 may not be the best way of handling this. I have been thinking about doing something like:

```
diff --git a/tests/own_addr.c b/tests/own_addr.c
index b5a9223..e297a12 100644
[...]

PR #219, opened 2022-12-11
Improved own addr test

martinwhitaker:
This modifies the own address + window test to use the physical address as the test pattern in 64-bit builds. This should make it easier to diagnose faults. Obviously we can't do this for 32-bit builds because the test word size isn't big enough to store a physical address.

This also disables test 1 by default, as discussed in issue #155. Test 2 should catch any faults caught by test 1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,8 +481,9 @@ selected by the user.
 
 ### Test 2 : Address test, own address + window
 
-Across all memory regions, each address is written with its own address plus
-the window number and then each address is checked for consistency. This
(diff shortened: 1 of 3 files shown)


x86fr:
After 6.01 release, I would like to merge #34 first, then all "ready" PRs including this one. With Black Friday and xmas, betatesters on SI production lines are very busy. They have less time to spend on testing betas, so they prefer to check all changes at once.

x86fr:
Tested OK. No regression on error detection rate/time. Merged

Issue #128, opened 2022-07-22
 Temperature reporting does not work on older AMD K8 CPUs


01e3:
Temperature reporting does not work on older AMD K8 CPUs

So far I have tested the following CPUs that don't work:
 - AMD Athlon(tm) Dual Core Processor 5400B
 - AMD Athlon(tm) 64 X2 Dual Core Processor 5000+

Both of them are family: 0xf, model: 0x6b, stepping: 0x2

![image](https://user-images.githubusercontent.com/105136494/180575651-2489a6de-6ec2-48fc-b4d8-6e9a66ebecc1.png)

x86fr:
Check first if they're not concerned by the infamous Errata 319 (see #3)

01e3:
Ah, was not aware about #3, reading...

That said, both Linux and HwMonitor seem to be happy:

```
# sensors k8temp-*
k8temp-pci-00c3
Adapter: PCI adapter
Core0 Temp:   +28.0Â°C
Core0 Temp:   +32.0Â°C
Core1 Temp:   +32.0Â°C
Core1 Temp:   +29.0Â°C

# ls -l /sys/bus/pci/drivers/k8temp/????:??:*
lrwxrwxrwx 1 root root 0 Jul 23 15:42 /sys/bus/pci/drivers/k8temp/0000:00:18.3 -> ../../../../devices/pci0000:00/0000:00:18.3
```


![image](https://user-images.githubusercontent.com/105136494/180610137-f508129b-765b-4fe2-8cd7-0cf5b23aa7a6.png)

PR #220, opened 2022-12-12
Enable temperature on K8

montjoie:
Add support for K8 temp by using the formula from linux/k8temp. This is tested on a K8T800-8237 with a AMD Athlon(tm) 64 Processor 3000+.

Fixes: #128

Signed-off-by: Corentin Labbe <clabbe.montjoie@gmail.com>

Issue #70, opened 2022-05-15
Support for 256+ threads (x2APIC & AMD Extended APIC ID)


debrouxl:
At the time of this writing, x86_64 computers with 256+ threads already exist: 2S EPYC 77x[23], 8S Xeon Platinum 82xx for select models. Though they'll remain fairly niche, their market share and ease of access will keep increasing, e.g. with the Zen 4 series, which is rumored to contain some 96-core models - i.e. 384 threads in a 2S configuration, as those models are likely to offer SMT=2 like their EPYC Zen 2 and Zen 3 predecessors did - or probably some Sapphire Rapids Xeon models. There are higher-priority features, but the 256 thread limit is going to hit some new users in the near future.

Technically, the solution to the 256 core limit issue of the original 8-bit APIC is available from Nehalem onwards in Intel processors ("x2APIC", CPUID 1 ecx register bit 21 & CPUID B & CPUID 1F preferred to B if available), and more or less differently on AMD processors (at least "Extended APIC space", with two CPUID-based methods depending on the generation).
The 6272 Bulldozer and 6378 Piledriver Opterons have support for AMD Extended APIC space but do not advertise the x2APIC bit in CPUID 1 ecx. AMD processors from Bulldozer onwards are supposed to set CPUID 80000001 ecx bit 22 (topology extensions) and provide the 32-bit extended APIC ID in CPUID 8000001E eax. On the Ryzen 7 PRO 5750GE (Zen 3), the CPUID data advertises both x2APIC and AMD Extended APIC space.

x86fr:
Adding x2APIC support is nice, but there is a real debate here. For a Memory Testing Tool, does using as much core as possible make the software more efficient to detect memory issues? Massive multi-threading is great to use as much raw computing power as available. There is no doubt about this. But considering that a single core of any modern CPU can retire data at a much higher bandwidth than available from its IMC (even a 4 channel DDR5), why do we need to split a work that can be done on a single thread to 64, 256 or 1024 threads?

[...]

PR #234, opened 2023-01-04
Enable temperature on K8/K10

montjoie:
Add support for K8/K10 temp by using the formula from linux drivers (k8temp/k10temp).
For K8, this is tested on a K8T800-8237 with an AMD Athlon(tm) 64 Processor 3000+.
For K10, this is tested on a K10N78 with an AMD Sempron(tm) 145 Processor

Fixes: #128

Signed-off-by: Corentin Labbe <clabbe.montjoie@gmail.com>

### DIFF
--- a/system/temperature.c
+++ b/system/temperature.c
@@ -21,6 +21,141 @@
 
 #include "temperature.h"
 
+static uint16_t pci_read_pid(int bus, int dev, int func)
+{
+    return pci_config_read16(bus, dev, func, 2);
+}
+
+static uint16_t pci_read_vid(int bus, int dev, int func)
+{
+    return pci_config_read16(bus, dev, func, 0);
+}
+
+static int k8dev, k8fun;
+
+// find the first PCI device with the related IDs
+// On a multi-socketed K8, it grab only the first one since memtest only show
+// one temp
+static bool find_k8_temp(void)
+{
+    uint32_t pid, vid;
+
+    for (k8dev = 0; k8dev < 32; k8dev++) {
+        for (k8fun = 0; k8fun < 8; k8fun++) {
+            vid = pci_read_vid(0, k8dev, k8fun);
+            pid = pci_read_pid(0, k8dev, k8fun);
+            if (vid == PCI_VID_AMD && pid == 0x1103)
+                return true;
+        }
+    }
+    return false;
+}
+
+static int k8_temp(void)
+{
+    // k8temp is stored on a PCI device ([AMD] K8 [Athlon64/Opteron] Miscellaneous Control)
+#define K8_REG_TEMP   0xe4
+#define K8_SEL_CORE   0x04
+    uint32_t reg;
+    int temp = 0;
+
+    if (!find_k8_temp())
+       return 0;
+
+    if (cpuid_info.version.model == 4 && cpuid_info.version.stepping == 0)
+        return 0;
+    if (cpuid_info.version.model == 5 && cpuid_info.version.stepping <= 1)
+        return 0;
+    // TODO: backport Linux test is_rev_g_desktop() which add 21°
+
+    // We want temperature of core0
+    if (cpuid_info.version.model >= 0x40)
+        temp = K8_SEL_CORE;
+    pci_config_write8(0, k8dev, k8fun, K8_REG_TEMP, temp);
+    reg = pci_config_read32(0, k8dev, k8fun, K8_REG_TEMP);
+    // formula is taken from k8temp Linux driver
+    temp = (reg >> 16) & 0xFF;
+
+    // nobody will use memtest in a frozen env
+    if (temp < 49)
+        return 0;
+
+    return temp - 49;
+}
+
+// K10 temperature is stored in a PCI device [AMD] Family 10h Processor Miscellaneous Control
+// There is one PCI device per CPU socket
+static int k10dev, k10fun;
+
+#define CPUID_PKGTYPE_MASK      0xF0000000
+#define CPUID_PKGTYPE_F         0x00000000
+#define CPUID_PKGTYPE_AM2R2_AM3 0x10000000
+#define REG_DCT0_CONFIG_HIGH    0x094
+#define DDR3_MODE 1 << 8
+static bool has_erratum_319(void)
+{
+    uint32_t eax, ebx, ecx, edx;
+    uint32_t pkg_type, reg_dram_cfg;
+
+    if (cpuid_info.version.family != 0x10)
+        return false;
+
+    cpuid(0x80000001, 0, &eax, &ebx, &ecx, &edx);
+
+    pkg_type = ebx & CPUID_PKGTYPE_MASK;
+
+    if (pkg_type == CPUID_PKGTYPE_F)
+        return true;
+    if (pkg_type != CPUID_PKGTYPE_AM2R2_AM3)
+        return false;
+
+    reg_dram_cfg = pci_config_read32(0, k10dev, 2, REG_DCT0_CONFIG_HIGH);
+    if (reg_dram_cfg & DDR3_MODE)
+        return false;
+
+    return false;
+}
+
+static const uint32_t k10temp_id_table[] = {
+    0x1203,
+};
+static bool find_k10_temp(void)
+{
+    uint32_t pid, vid, i;
+
+    for (k10dev = 0; k10dev < 32; k10dev++) {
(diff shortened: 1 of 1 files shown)


montjoie:
This is a follow up of https://github.com/memtest86plus/memtest86plus/pull/220

system/temperature.c
```diff
@@ -21,6 +21,141 @@
...
+#define DDR3_MODE 1 << 8
+static bool has_erratum_319(void)
+{
+    uint32_t eax, ebx, ecx, edx;
+    uint32_t pkg_type, reg_dram_cfg;
+
+    if (cpuid_info.version.family != 0x10)
+        return false;
+
+    cpuid(0x80000001, 0, &eax, &ebx, &ecx, &edx);
```
debrouxl, line 103: Not directly related to your patch, but I think that this is a good opportunity to store more extended feature flags in cpuid_feature_flags_t , as cpuid 80000001 is also needed for the AMD extended APIC space feature ( #70 et al.), and it's silly to call cpuid twice, then partially discard the output on both sides. Sam, Martin ?

system/temperature.c
```diff
@@ -21,6 +21,141 @@
...
+        return false;
+
+    reg_dram_cfg = pci_config_read32(0, k10dev, 2, REG_DCT0_CONFIG_HIGH);
+    if (reg_dram_cfg & DDR3_MODE)
+        return false;
+
+    return false;
+}
+
+static const uint32_t k10temp_id_table[] = {
```
debrouxl, line 119: I suppose that there are more PCI PIDs coming in followup versions of the patch ? Otherwise, a single comparison like in `find_k8_temp()` above will do the job - and make it easier to merge `find_k8_temp()` and `find_k10_temp()` together.

system/temperature.c
```diff
@@ -21,6 +21,141 @@
 
 #include "temperature.h"
 
+static uint16_t pci_read_pid(int bus, int dev, int func)
```
debrouxl, line 24: With such trivial implementations, I think that `pci_read_pid()` and `pci_read_vid()` ought to be static inline functions in system/pci.h .

system/temperature.c
```diff
@@ -21,6 +21,141 @@
...
+
+static int k8_temp(void)
+{
+    // k8temp is stored on a PCI device ([AMD] K8 [Athlon64/Opteron] Miscellaneous Control)
+#define K8_REG_TEMP   0xe4
+#define K8_SEL_CORE   0x04
+    uint32_t reg;
+    int temp = 0;
+
+    if (!find_k8_temp())
```
debrouxl, line 62: Besides the potential merge of `find_k8_temp()` and `find_k10_temp()` I mentioned in another comment, this implementation of `k8_temp()` + `find_k8_temp()` and `k10_temp()` + `find_k10_temp()` reads the PCI bus config space every time, without caching the result of the enumeration. I can live with spending several bytes to avoid this repeated enumeration :)

system/temperature.c
x86fr: This patch is basically a copy/paste of the related k8temp function without taking into consideration the use done by Memtest86+ (which is very different to k8temp) and its already-implemented functions. This can be extremely simplified without any feature loss. I'll try to do it later today.

system/temperature.c
```diff
@@ -21,6 +21,141 @@
...
+{
+    uint32_t pid, vid;
+
+    for (k8dev = 0; k8dev < 32; k8dev++) {
+        for (k8fun = 0; k8fun < 8; k8fun++) {
+            vid = pci_read_vid(0, k8dev, k8fun);
+            pid = pci_read_pid(0, k8dev, k8fun);
+            if (vid == PCI_VID_AMD && pid == 0x1103)
+                return true;
+        }
```
x86fr, line 49: Enumeration is useless as the needed device will always be available at 0/24/3. No exception. 
On a multi-socketed K8, the BSP will be found here.

system/temperature.c
x86fr: Yes, this should be done on cpuid.c

system/temperature.c
montjoie: Yes there are more, but the K10temp handle lot of different way of getting temperature, so I cannot add them without testing.

system/temperature.c
montjoie: it is why k8dev,k8fun are static, I just forgot to add a if (!=0)return

system/temperature.c
montjoie: I created a qemu emulation of k8/k10temp PCI device, so this is usefull in that case (I dont find how to fix bus/dev in qemu). I understand this can be simplified. Unless you want a day to handle multi socket temperature.

system/temperature.c
```diff
@@ -47,14 +182,11 @@ int get_cpu_temperature(void)
...
+    if (cpuid_info.vendor_id.str[0] == 'A' && cpuid_info.version.extendedFamily == 0) {
+        return k8_temp();
+    }
     if (cpuid_info.vendor_id.str[0] == 'A' && cpuid_info.version.extendedFamily > 0 && cpuid_info.version.extendedFamily < 8) {
-
-        // Untested yet
-        uint32_t rtcr = pci_config_read32(0, 24, 3, 0xA4);
-        int raw_temp = (rtcr >> 21) & 0x7FF;
-
-        return raw_temp / 8;
```
x86fr, line 56: Is there anything wrong with this simple code (expect errata 319 handling)?
It basically do the same thing (read 0/24/3/0xA4, shift 21, divide by 8) as all code added before.

system/temperature.c
```diff
@@ -21,6 +21,141 @@
...
+
+    if (cpuid_info.version.family != 0x10)
+        return false;
+
+    cpuid(0x80000001, 0, &eax, &ebx, &ecx, &edx);
+
+    pkg_type = ebx & CPUID_PKGTYPE_MASK;
+
+    if (pkg_type == CPUID_PKGTYPE_F)
+        return true;
```
x86fr, line 108: Starting here, below code is useless because it can only return false.

k10temp.c final return is based on cpu model and mask, which has been removed here.

system/temperature.c
x86fr: 0/24/3 is an internal register from the BSP. It can't be absent, hidden or anywhere else on any CPU targeted by the previous case (`cpuid_info.vendor_id.str[0] == 'A' && cpuid_info.version.extendedFamily > 0 && cpuid_info.version.extendedFamily < 8`).

Are you aware of any real-word platform where 0/24/3 does not exist?

montjoie: From the few hardware I have, no.
I don't see any reason to not add a little test for being safe but if you say you know for sure it will exists at that place, I will dont do any objections.

system/temperature.c
x86fr: Finally not so silly because the (useful) result of CPUID 0x80000001[EBX] is specific to K8, so it does make sense to run cpuid again later, only when this specific CPU has been detected.

system/temperature.c
```diff
@@ -21,6 +21,141 @@
...
+    // k8temp is stored on a PCI device ([AMD] K8 [Athlon64/Opteron] Miscellaneous Control)
+#define K8_REG_TEMP   0xe4
+#define K8_SEL_CORE   0x04
+    uint32_t reg;
+    int temp = 0;
+
+    if (!find_k8_temp())
+       return 0;
+
+    if (cpuid_info.version.model == 4 && cpuid_info.version.stepping == 0)
```
x86fr, line 65: This can't be translated like this. The "model" var on k8temp is the full cpuid model (Ext_model << 4 | model) while the cpuid_info.version.model only contain the base model. So this will trigger on model 0x24 for example)

x86fr:
There's too many things wrong on this PR. Implementing correctly K8/K10 temp reading is not as trivial. 
Errata 319 should be handled correctly (even if implementation is partial) as well as Rev G Mobile & Desktop detection.

I already wrote some code for this, but I need to set up various old K8/K10 platforms to exercise the code.
Hope to be able to test it later this week

debrouxl:
FWIW, I've just put my "AMD Athlon(tm) 64 X2 Dual Core Processor 5600+" K8 on Gigabyte M61P-S3 back into operation, after moving the MB to a new ATX case.

k8ieone:
@x86fr I have a few different K8/K10 systems, let me know if you want me to test anything ;)

Sempron LE-1150
Turion 64 X2 (Laptop)
Phenom II X4 945
FX-8320

Issue #243, opened 2023-01-21
Direct boot on 32-bit EFI systems is broken


martinwhitaker:
When booting the standard memtest.iso on a real 32-bit EFI system or using qemu with the OVMF ia32 BIOS, boot fails with the message "failed to get handle for loaded image protocol". Booting via GRUB is not affected.

git bisect indicates this regression was introduced by commit 04980dfda346fd0f7e6cb51d8e5b1ea88327b9c4.

x86fr:
Beside Atom N270/N280s, are you aware of any other 32-bit UEFI platform? I'm trying to find a board to check.

martinwhitaker:
My understanding is that anything based on the Intel Bay Trail platform was obliged (by Intel) to use 32-bit EFI. I have an Asus X205TA which has a Z3735 CPU.

x86fr:
I've just plugged a MSI J1800I board featuring a Celeron J1800 (Bay Trail). Unfortunately, the EFI is 64-bit (v2.31).
I have an older D510 board (Pineview 64-bit) and also a N270 board (Diamondville 32-bit). Will try.

x86fr:
The Intel D510MO board (based on Atom D510) can't boot anymore :-/
The Intel D945GSEJT board (based on Atom N270) does not use EFI (BIOS only, which works fine).
Will dig more in the closet... 🤔

x86fr:
Finally resurrected the D510MO using black magic. Its EFI BIOS (circa 2001) is very archaic, with very few options related to EFI and no shell. 

The PXE/gPXE options still uses legacy BIOS, but I'm able to boot a .efi file when placed in /EFI/BOOT. 
Unfortunately, it only load bootx64.efi (works fine) and ignore bootia32.efi, so I suppose the UEFI on this board is 64 bit.

martinwhitaker:
You want an Intel Atom Bay Trail tablet from circa 2013 to 2015. I bought mine to enable me to add/test support for 32-bit UEFI in Mageia.

x86fr:
Finding one able to boot the 32bit .efi via PXE would be great. I'll dig eBay...

martinwhitaker:
Fixed in 6.10.

Issue #233, opened 2023-01-02
Unexpected interrupt on CPU 0


johanehnberg:
## Memtest86+
* Fails immediately with `Unexpected interrupt on CPU 0`
* Running v6.01 from public 64-bit iso (32-bit also produces the error)
* Booting as legacy BIOS boot over virtual USB CDROM (ISO file mounted by BMC)
* v5.01 runs smoothly

![Screenshot from 2023-01-02 15-45-04](https://user-images.githubusercontent.com/6415888/210240936-9e5e0866-17cf-4927-95f0-2809a7820d42.png)
![Screenshot from 2023-01-02 15-42-32](https://user-images.githubusercontent.com/6415888/210241005-fc65c780-d182-4482-b91f-c5b93bf97872.png)

## Hardware
I have access to two of these and can do cross-checking as requested. The other one still has an older 2.0a BIOS version for example. This system tested now is running the latest BIOS and BIOS settings left at their defaults except boot.

```
    description: System
    product: Super Server (To be filled by O.E.M.)
    vendor: Supermicro
    width: 64 bits
    capabilities: smbios-2.8 dmi-2.8 smp vsyscall32
    configuration: boot=normal chassis=server family=To be filled by O.E.M. sku=To be filled by O.E.M.
  *-core
       description: Motherboard
       product: X10SDV-4C-TLN4F
       vendor: Supermicro
       physical id: 0
       version: 2.00
       slot: To be filled by O.E.M.
     *-firmware
          description: BIOS
          vendor: American Megatrends Inc.
          physical id: 0
          version: 2.3
          date: 06/04/2021
          size: 64KiB
          capacity: 16MiB
          capabilities: pci upgrade shadowing cdboot bootselect socketedrom edd int13floppy1200 int13floppy720 int13floppy2880 int5printscreen int9keyboard int14serial int17printer acpi usb biosbootspecification uefi
     *-memory
          description: System Memory
          physical id: 19
          slot: System board or motherboard
          size: 64GiB
          capabilities: ecc
          configuration: errordetection=multi-bit-ecc
        *-bank:0
             description: DIMM DDR4 Synchronous 2133 MHz (0.5 ns)
             product: M393A4K40BB0-CPB
             vendor: Samsung
             physical id: 0
             slot: DIMMA1
             size: 32GiB
             width: 64 bits
             clock: 2133MHz (0.5ns)
        *-bank:1
             description: DIMM DDR4 Synchronous [empty]
[...]

Issue #202, opened 2022-11-08
memtest86+ EFI PE image does not load on some firmware


medhefgo:
On some firmware, the memtest efi binary trips up the firmware when loaded using the provided boot service `LoadImage` (for example by systemd-boot).

This is likely due to the PE optional header being too small (sbverify doesn't like this binary because of this). The efi binary can be fixed and made to boot with `objecopy memtest.efi memtest-fixed.efi`.

Grub does not exhibit this behavior as it presumably does not use `LoadImage` to boot.

Discussed in https://github.com/memtest86plus/memtest86plus/discussions/199
Arch bug: https://bugs.archlinux.org/task/76390

x86fr:
This might be solved by #34

mattia-b89:
> This might be solved by #34
> 
> Update: It seems fixed with #34, but another confirmation would be nice :)

As soon as it hits upstream release

mattia-b89:
Just an update: v6.01 does not fix the issue.

debrouxl:
Indeed, the commits from #34 were only integrated after v6.01 . You may test the nightly build, available from https://memtest.org/dev/  :)

medhefgo:
Yikes, from that nightly build:
```sh
$ sbverify --list bootx64.efi 
warning: gap in section table:
    datadir[CERT]->headers: 0x0000012a - 0x00000200,
    .text   : 0x00000600 - 0x00022400,
warning: gap in section table:
    .reloc  : 0x00022400 - 0x0002240a,
    .sbat   : 0x00022600 - 0x00022800,
gaps in the section table may result in different checksums
warning: data remaining[139786 vs 141312]: gaps between PE/COFF sections?
No signature table present
```

That really should be fixed, as there is firmware that will choke on that. (Calling `strip` would do.)

x86fr:
Do you have any real-world exemple? We started a lengthy betatesting process on numerous Secureboot platforms without related issues so far. Maybe we're missing something.

medhefgo:
https://github.com/systemd/systemd/issues/21956

WinterStar2:
Signing and booting the .EFI file works with the latest version (6.10, on Arch).

mattia-b89:
Latest 6.10 version fixes issue.  
If @medhefgo confirm, we can close here.

medhefgo:
Looks good.

PR #254, opened 2023-01-31
EFI image file header fixes

martinwhitaker:
When the reloc and sbat sections were added by PR #34, three bugs were introduced:
    
1. The virtual address and size fields in the PE headers were set to the same values as the raw address and size fields. This is incorrect, because the sections in the image file are aligned on 512 byte boundaries, but when loaded into memory they need to be aligned on 4096 byte boundaries.

2. The value programmed into the SizeOfImage field was too large, as it double-counted the region before the start of the .text section.

3. The value programmed into the SizeOfImage field no longer included the bss size. That potentially allowed the EFI loader to load the image immediately before a reserved region of memory without leaving enough space for the bss section.
    
This PR fixes those bugs by calculating both file and virtual memory offsets & sizes in the ld script. Note that we can't add a bss section to the EFI image because many EFI loaders fail to load images that have uninitialised data sections. Instead the text region size in virtual memory is increased to include the bss size.

Whilst testing, I also spotted that the AP stack size wasn't being added to the bss size in 32-bit builds, so this PR also fixes that, along with a few cosmetic issues.

This fixes issue #243. It also eliminates the gaps between sections remarked upon in issue #202.

### DIFF
--- a/boot/header.S
+++ b/boot/header.S
@@ -191,7 +191,6 @@ section_table:
 	.word	0				# NumberOfLineNumbers
 	.long	IMAGE_SCN_MEM_READ		\
 	      | IMAGE_SCN_MEM_EXECUTE		\
-	      | IMAGE_SCN_ALIGN_16BYTES		\
 	      | IMAGE_SCN_CNT_CODE		# Characteristics (section flags)
 
 	.ascii	".reloc"
@@ -206,7 +205,6 @@ section_table:
 	.word	0				# NumberOfRelocations
 	.word	0				# NumberOfLineNumbers
 	.long	IMAGE_SCN_MEM_READ		\
-	      | IMAGE_SCN_ALIGN_4BYTES		\
 	      | IMAGE_SCN_CNT_INITIALIZED_DATA	# Characteristics (section flags)
 
 	.ascii	".sbat"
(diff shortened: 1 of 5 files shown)


x86fr:
Great! 

I'll ask betatesters to do a quick check tomorrow for non-regression on Secureboot signing. That will be done on only ~30 platforms but should be enough to be sure everything's correct.

#If all goes well, it will be merged later today.

martinwhitaker:
> Is there a chance that these changes, e.g. the BSS size restored in SizeOfImage, also fix #233 ?

No, because the initial comment there says it's booting from a legacy BIOS and happens with both 32-bit and 64-bit images.

x86fr:
Everything looks good so far, but we still have some gaps on the unstripped binary:
```
memtest-dev:/home/memtest/SecureBoot# sbverify --list memtest-sign.efi
warning: gap in section table:
    datadir[CERT]->headers: 0x0000012a - 0x00000200,
    .text   : 0x00000600 - 0x00023400,
gaps in the section table may result in different checksums
warning: data remaining[146352 vs 147376]: gaps between PE/COFF sections?
```
Should we strip the efi binary in Makefile?

martinwhitaker:
That gap contains the setup sectors (setup.S). If stripping removes it, the image will no longer be bootable via GRUB and all the file offsets in the PE/COFF header will be wrong.

If we need to eliminate that warning, we can probably do it by changing the SizeOfHeaders field in the PE header.

x86fr:
Yes! Can you push a commit to increased them by 1KB? (tested working and warnings vanished).

martinwhitaker:
> Yes! Can you push a commit to increased them by 1KB? (tested working and warnings vanished).

Done.

cmp -l saved-memtest.efi memtest.efi
   208   2   6

PR #259, opened 2023-02-06
Support cache and temperature info for VIA/Centaur/Zhaoxin CPUs

jonathan-teh:
Use extended CPUID for VIA C3/C7/Nano cache information.

Use MSR reads for Nano/Zhaoxin and VIA C7 processor temperature.

Tested on VIA C7-D 1.5GHz.

### DIFF
--- a/system/cpuid.c
+++ b/system/cpuid.c
@@ -128,6 +128,26 @@ void cpuid_init(void)
             );
         }
         break;
+      case 'C':
+        if (cpuid_info.vendor_id.str[5] == 'I') break; // Cyrix
+        // VIA / CentaurHauls
+        if (cpuid_info.max_xcpuid >= 0x80000005) {
+            cpuid(0x80000005, 0,
+                &reg[0],
+                &reg[1],
+                &cpuid_info.cache_info.raw[0],
+                &cpuid_info.cache_info.raw[1]
+            );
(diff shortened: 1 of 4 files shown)


jonathan-teh:
memtest-c7-d

debrouxl:
Thanks. Looks reasonable to me, as the C7-D 1500/400 has 64K Instruction & 64K Data cache.
Sam ?

x86fr:
Looks good. I'm also working on VIA CPU to solve issues on core count. I'll do a test pass on various VIA and zhaoxin CPUs  later this week.

jonathan-teh:
Reading the rest of the CPU detection code, I think that in `cpuid_init()` it would be better to always populate `cpuid_info.cache_info` from the extended CPUID, gated by `max_xcpuid`, instead of also filtering it by vendor. `determine_cache_size()` doesn't use it for Intel CPUs anyway and it'll make things easier for detecting cache sizes for other CPUs. I've looked at `determine_cpu_model()` and I'd like to improve the VIA/Centaur/Cyrix section, after this is merged.

x86fr:
Tested OK on various VIA platforms from C3 Nehemiah 533 MHz to Zhaoxin KX-5640 2.0 GHz

PR #263, opened 2023-02-09
Fix overallocation of cpuid_cache_info_t union

jonathan-teh:
Saves 32B per instance.


x86fr:
Merged with #259 (much easier for testing purpose)

PR #267, opened 2023-02-12
[cpuinfo] Fix old CPUs (P5/P6-class) name and cache info

jonathan-teh:
Always populate the cache info from extended CPUID, it is not used for Intel CPUs, even though it is present, and is useful for non-Intel CPUs.

Fix the CPU name and cache sizes for Centaur and Cyrix/NSC CPUs without brand string, which are the WinChip C6 and all Cyrix CPUs except the Media GXm.

For the Media GXm and Geode GXm/GXLV/GX1, which are available with both Cyrix and NSC vendor strings, hardcode the L1 cache size. The Geode GX2 uses standard cache info.

### DIFF
--- a/system/cpuid.c
+++ b/system/cpuid.c
@@ -24,6 +24,8 @@ void cpuid_init(void)
     uint32_t reg[4];
     char *p, *q;
 
+    return;
+
     // Get the max standard cpuid & vendor ID.
     cpuid(0x0, 0,
         &cpuid_info.max_cpuid,
--- a/system/cpuid.c
+++ b/system/cpuid.c
@@ -24,8 +24,6 @@ void cpuid_init(void)
     uint32_t reg[4];
     char *p, *q;
 
-    return;
-
     // Get the max standard cpuid & vendor ID.
     cpuid(0x0, 0,
         &cpuid_info.max_cpuid,
--- a/system/cpuid.c
+++ b/system/cpuid.c
@@ -108,50 +108,21 @@ void cpuid_init(void)
     }
 
     // Get cache information.
-    switch (cpuid_info.vendor_id.str[0]) {
-      case 'A':
-        // AMD Processors
-        if (cpuid_info.max_xcpuid >= 0x80000005) {
-            cpuid(0x80000005, 0,
-                &reg[0],
-                &reg[1],
-                &cpuid_info.cache_info.raw[0],
-                &cpuid_info.cache_info.raw[1]
-            );
-        }
-        if (cpuid_info.max_xcpuid >= 0x80000006) {
-            cpuid(0x80000006, 0,
-                &reg[0],
-                &reg[1],
-                &cpuid_info.cache_info.raw[2],
-                &cpuid_info.cache_info.raw[3]
-            );
-        }
-        break;
-      case 'C':
-        if (cpuid_info.vendor_id.str[5] == 'I') break; // Cyrix
-        // VIA / CentaurHauls
-        if (cpuid_info.max_xcpuid >= 0x80000005) {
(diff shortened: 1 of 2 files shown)


jonathan-teh:
Tested on NSC Geode GX1 with Cyrix vendor string.
memtest_gx1
This is based on Centaur and Cyrix datasheets, and the raw CPUID dumps at http://instlatx64.atw.hu/.
I'll work on the more "exotic" SiS, Transmeta, and Vortex86 CPU detection, seen in #20, next as there are raw CPUID dumps for those.

x86fr:
Ok, but please note there is a LOT of erratas and wrong specs on old Cyrix/IDT datasheets. All code paths must be tested on real hardware before addition. Please provide a screenshot for every added CPUID family.

PS: I was about to push a commit for SiS550 and Vortex86, bit if you can do it, I'll happily leave it to you (I can't find my Vortex86EX anymore)

jonathan-teh:
I'm afraid I can't test the non-brand string codepaths as I don't have any of the hardware. I understand that the datasheets may not be correct but the raw CPUID dumps (I also checked some of the processed CPUIDs at [cpu-world](https://www.cpu-world.com/cgi-bin/CPUID.pl)) should give some confidence that it should work. In particular, the current Cyrix CPU detection doesn't match up with either source of information.

> PS: I was about to push a commit for SiS550 and Vortex86, bit if you can do it, I'll happily leave it to you (I can't find my Vortex86EX anymore)

Feel free to push it if it's already done, I don't have that hardware either.

x86fr:
Testing on real hardware is important on these old platforms, because unexpected or side effects are common.

e.g. cpuid is present but often disabled on 6x86, Processor Name String is writable on C7 via MSR and some BIOS set it incorrectly, hardcoding the cache size will trigger the bandwidth test code which may fail on odd platforms, etc.

Anyway, I still have all these platforms somewhere (and many related prototype/unreleased boards) and it's always a pleasure to power them up again... 😍

jonathan-teh:
OK, I'm aware of Cyrix weirdness, which usually have their own CPU detection routines, but I guess if CPUID is disabled in BIOS this whole section would skipped anyway. None of the CPUs added here have L2 cache so they won't trigger the bandwidth test (I did wonder why it was disabled).

> Anyway, I still have all these platforms somewhere (and many related prototype/unreleased boards) and it's always a pleasure to power them up again... :heart_eyes:

I'll leave the testing to you then, I don't have such a large collection :smile:

x86fr:
> None of the CPUs added here have L2 cache so they won't trigger the bandwidth test (I did wonder why it was disabled).

They were just generic examples, but for this one ... except if you run it on a motherboard based on the popular ALi Aladin V chipset (where onboard L2 is detected later in a specific hwquirks) 😉

x86fr:
PS: I will hijack this PR with some commits soon. It's easier for testing than a separate one 😉

x86fr:
Memtest86+ does not seem to boot at all on MediaGXi (instant reboot). Can't see any obvious solution. Maybe something wrong with my motherboard (5.31b also fail).

jonathan-teh:
> MediaGX is fine (brand string is supported)

That's actually a Media GXm, same CPUID 540h shared with the Geode GX1. The original Media GX (non-MMX) doesn't have the brand string.

BTW, will you be adding the CPUID display in the lower-left? I spotted it on some pictures but the code's not in the main branch.

x86fr:
Yes, the original non-mmx Gx86 (or MediaGX) is very rare and BGA only. The MXi is more common but still BGA only and the GXm is the most "common" (BGA and PGA). CPUID display is from by debug libraries, it's not intended for release.

PR #269, opened 2023-02-13
[Temperature] Add support for CPUs with specific TjMax

x86fr:
Solve an issue where reading MSR_IA32_TEMPERATURE_TARGET makes the system crash (Celeron M 215 / Intel D201GLY board)

### DIFF
--- a/system/temperature.c
+++ b/system/temperature.c
@@ -33,32 +33,55 @@ float cpu_temp_offset = 0;
 // Public Functions
 //------------------------------------------------------------------------------
 
+int TjMax = 0;
+
+uint32_t regl, regh;
+
+void get_specific_TjMax(void)
+{
+    // The TjMax value for some Mobile/Embedded CPUs must be read from a fixed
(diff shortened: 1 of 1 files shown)


system/temperature.c
```diff
@@ -33,58 +33,81 @@ float cpu_temp_offset = 0;
 // Public Functions
 //------------------------------------------------------------------------------
 
+int TjMax = 0;
```
debrouxl, line 36: +static ?

system/temperature.c
```diff
@@ -33,58 +33,81 @@ float cpu_temp_offset = 0;
...
+
+uint32_t regl, regh;
```
debrouxl, line 38: Unless you need to export them later, why not keep them local variables ?

system/temperature.c
x86fr: Yep!

system/temperature.c
x86fr: 1 line saved 😅

Seriously, yes, will move them back.

PR #276, opened 2023-02-24
temperature: Refactor to separate detection and read

jonathan-teh:
Move the detection of the temperature sensor type based on CPUID to temperature_init and store the type in an enum. This allows get_cpu_temperature to use a simpler switch on that enum.

### DIFF
--- a/system/temperature.c
+++ b/system/temperature.c
@@ -44,7 +44,6 @@ typedef enum {
 } cpu_temp_t;
 
 static cpu_temp_t cpu_temp_type;
-static cpu_temp_t get_cpu_temp_type(void);
 
 static int TjMax = 0;
 
@@ -58,16 +57,6 @@ static void get_specific_TjMax(void)
     if (cpuid_info.version.raw[0] == 0x6E8) {
         TjMax = 100;
     }
-}
-
-void temperature_init(void)
-{
-    // Process temperature-related quirks
-    if (quirk.type & QUIRK_TYPE_TEMP) {
-        quirk.process();
-    }
-
-    cpu_temp_type = get_cpu_temp_type();
 }
 
 static cpu_temp_t get_cpu_temp_type(void)
@@ -118,6 +107,16 @@ static cpu_temp_t get_cpu_temp_type(void)
     return None;
 }
 
+void temperature_init(void)
+{
+    // Process temperature-related quirks
+    if (quirk.type & QUIRK_TYPE_TEMP) {
+        quirk.process();
+    }
+
+    cpu_temp_type = get_cpu_temp_type();
+}
+
 int get_cpu_temperature(void)
 {
     uint32_t regl, regh;
@@ -158,6 +157,9 @@ int get_cpu_temperature(void)
       case ViaC7:
         rdmsr(MSR_VIA_TEMP_C7, regl, regh);
         return (int)(regl & 0xffffff);
+
+      case None:
+        break;
     }
 
     return 0;
--- a/system/temperature.c
+++ b/system/temperature.c
@@ -33,9 +33,22 @@ float cpu_temp_offset = 0;
 // Public Functions
 //------------------------------------------------------------------------------
 
+typedef enum {
+    None,
+    Intel,
+    AmdZen,
+    AmdK10,
+    AmdK8,
+    ViaNano,
+    ViaC7
+} cpu_temp_t;
+
+static cpu_temp_t cpu_temp_type;
+static cpu_temp_t get_cpu_temp_type(void);
+
 static int TjMax = 0;
 
-void get_specific_TjMax(void)
+static void get_specific_TjMax(void)
 {
     // The TjMax value for some Mobile/Embedded CPUs must be read from a fixed
     // table according to their CPUID, PCI Root DID/VID or PNS.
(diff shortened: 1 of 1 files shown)


jonathan-teh:
Tested the K8, C7, and no-sensor (Geode) codepaths, screenshots are the same as what I've previously posted.
Also tested the Intel:
memtest-x50gl
and Zen:
memtest-avita
codepaths.

This should make it easier to add new temperature sensors and perhaps move the K8/K10 temperature hwquirks here instead of repeating the tests for K8/K10 processors.

A previous attempt used a separate function for each sensor and a function pointer but that had "interesting" runtime issues from page faults to overall test pass prematurely counting to 100%. Perhaps it's due to memtest86+ relocating itself during execution.

Anyway, the compiler (gcc-12.2.1) appears to split get_temperature into separate functions anyway, at least for the 32-bit binary.

debrouxl:
Interesting :)

You could avoid the forward declaration for `get_cpu_temp_type()` by moving it above `temperature_init()`.

Separate functions for each sensor model and function pointers are pretty bad for size efficiency, due to the size of the pointer itself, the relocation records (which require special declaration in memtest86+, like in the USB code), the call / return / function prologue / function epilogue (/ in general, function argument passing and retrieval, but there's no function argument here), etc.
When possible (and in this case, it is easy), small functions and function pointers should be avoided in memtest86+, and in a number of other code bases as well, in fact :)

jonathan-teh:
> You could avoid the forward declaration for `get_cpu_temp_type()` by moving it above `temperature_init()`.

I was trying to keep the relative positions of the functions the same but I guess it's going to be a big diff either way. Anyway, pushed a follow-up.

x86fr:
To me, there is very few advantages adding this. Refactoring is nice when the code actually needs it, but the temp code is already very simple and quite understandable. There is no evidence at this point that it will grow significantly in a near future.

debrouxl:
Maybe the hwquirks related to stupid / broken temperature reports on K8 & K10 could be folded into this refactor ? `temperature_init()` is the last hardware `_init()` function called by `global_init()`.
This may make the code simpler and easier to understand, if that's your concern :)

jonathan-teh:
The code is reasonably simple but I think it can be made simpler :smiley:  Before the VIA temperature sensors were added there were already 4 sensor types (plus none). Now there are 2 more sensor types and 4 quirks (if you also count the Intel TjMax) in the detection phase. There are some immediate advantages:

1. Saves CPU cycles from not repeatedly detecting sensor type in `get_cpu_temperature()`.
2. Saves an additional test for Intel CPU type.

and future ones:

1. Moving the AMD K8/K10 quirks to this file removes additional tests for K8/K10 and removes `amd_k8_revfg_temp()` from the symbol table if it's made `static` here. It also frees up `quirk.process` in case a platform requires it in addition to the K8/K10 temperature quirk.
2. Makes it easier to add additional current sensor types e.g. Hygon family 18h simply returns `AmdZen`. Someone else will have to add it though since I don't have access to such a platform as it wasn't sold outside China.
3. There may be additional sensor types in future, AMD K10 looked stable for a while until they changed it again for family 17h (Zen).
4. Easier to add temperature sensors from the past, e.g. Pentium III thermal diode read by an external sensor, which is motherboard specific (depends on how far down that rabbit hole you want to go, I do have a platform with that :smile: ).

x86fr:
> The code is reasonably simple but I think it can be made simpler 😃 

Every code can be made simpler. The problem here is significant code change requires tests (for non-regression, etc), so a PR must add something significant like solving a bug or add something to end user. if it ain't broke, don't fix it. 😉

> There are some immediate advantages:
> 
>     1. Saves CPU cycles from not repeatedly detecting sensor type in `get_cpu_temperature()`.
>     2. Saves an additional test for Intel CPU type.

On any CPU, even the oldest and slowest supported like a 486, you will save maybe a microsecond on a pass duration. It's irrelevant.

> and future ones:

Refactoring is great when the code has gotten too big and messy, or as a side enhancement when you add code that leads to the previous situation. Refactoring for refactoring, just to gain 10 bytes of code size or to save 100 CPU cycles every second is often useless if there's nothing useful added for the end user. Even worst, this may delay the addition of feature (or fixing a bug) by wasting the beta-testing time (which is not infinite).
 
>     1. Moving the AMD K8/K10 quirks to this file removes additional tests for K8/K10 and removes `amd_k8_revfg_temp()` from the symbol table if it's made `static` here. It also frees up `quirk.process` in case a platform requires it in addition to the K8/K10 temperature quirk.

Erratas belong to quirks. Only one quirks is currently handled, but a quirks array will be implemented as soon as required.

>     2. Makes it easier to add additional current sensor types e.g. Hygon family 18h simply returns `AmdZen`. Someone else will have to add it though since I don't have access to such a platform as it wasn't sold outside China.

You can do this with the actual code without a new enum (e.g. target AMD with `imc_type >= IMC_K8`)

>     3. There may be additional sensor types in future, AMD K10 looked stable for a while until they changed it again for family 17h (Zen).

Maybe soon, maybe later. Nobody knows today how this will change and if this refactoring is relevant or not. Leaving the choice how to refactor to the contributor who will actually add this new code is usually the better option.

>     4. Easier to add temperature sensors from the past, e.g. Pentium III thermal diode read by an external sensor, which is motherboard specific (depends on how far down that rabbit hole you want to go, I do have a platform with that 😄 ).

Good exemple. if you want to add Pentium III thermal diode reading via something like a ASUS ASB µC, you will need quirks and lot of side code that probably doesn't belong to temperature.c. Same remark than 3. 😉

jonathan-teh:
OK, I guess we have different opinions on what constitutes "simple". Although what I understand is if something like reading thermal diode via the Asus ASB µC is implemented, a refactoring like this would be more acceptable?

x86fr:
Basically, It will be acceptable when needed. But please follow the coding rules if you have to re-submit this on a bigger PR (eg: enum values must be all caps).

PR #277, opened 2023-02-25
Revert pre-enabled x2APIC by BIOS to xAPIC

x86fr:
Here's something we need to discuss. I'm working on full x2APIC support, but debugging is long and painful (messing with MSRs is a good way to experience a massive amount of GPFs...). 

Some weeks ago, a betatester needed a way to test an ES server (Intel SPR) with SMP enabled, but x2APIC was also enabled (and no way to disable it with a BIOS option), so I wrote a quick patch to disable x2APIC and revert back to xAPIC. It worked fine. 

I tuned it a bit and removed a lot of superfluous ASM checks for APs. It's now extremely small and seems to work as expected on platforms with x2APIC pre-enabled. 

To me, it solved 95+% of the issues we currently have with x2APIC (no SMT at all when x2APIC is enabled), leaving the remaining 5% (support for > 255 cores, which I still question the advantage for Memtest86+) to do. 

Should we start betatesting this for addition in main (before a cleaner but much longer x2APIC implementation in the future)?

### DIFF
--- a/boot/startup64.S
+++ b/boot/startup64.S
@@ -152,6 +152,25 @@ startup:
 
 	leaq	startup_stack_top(%rip), %rsp
 
+    # Disable x2APIC if supported & enabled
+
+    movl	$0x1, %eax
+    cpuid
+    notl	%ecx
+    andl	$0x200000, %ecx
+    jnz	nox2apic				# x2APIC supported?
+    movq	$0x1b, %rcx
+    rdmsr
+    notl	%eax
+    andl	$0xc00, %eax		# x2APIC enabled?
+    jnz	nox2apic
+    rdmsr
+    andl	$0xfffff3ff, %eax	# Disable xAPIC & x2APIC
+    wrmsr
+    orl	$0x800, %eax			# Enable xAPIC only
+    wrmsr
+nox2apic:
+
 	# Pick the correct stack.
 
 	xorq	%rax, %rax
--- a/boot/startup64.S
+++ b/boot/startup64.S
@@ -152,23 +152,23 @@ startup:
 
 	leaq	startup_stack_top(%rip), %rsp
 
-    # Disable x2APIC if supported & enabled
+	# Disable x2APIC if supported & enabled on APs
 
-    movl	$0x1, %eax
-    cpuid
-    notl	%ecx
-    andl	$0x200000, %ecx
-    jnz	nox2apic				# x2APIC supported?
-    movq	$0x1b, %rcx
-    rdmsr
-    notl	%eax
-    andl	$0xc00, %eax		# x2APIC enabled?
-    jnz	nox2apic
-    rdmsr
-    andl	$0xfffff3ff, %eax	# Disable xAPIC & x2APIC
-    wrmsr
-    orl	$0x800, %eax			# Enable xAPIC only
-    wrmsr
+	movl	$0x1, %eax
+	cpuid
+	notl	%ecx
+	andl	$0x200000, %ecx
+	jnz	nox2apic				# x2APIC supported?
+	movq	$0x1b, %rcx
+	rdmsr
+	notl	%eax
+	andl	$0xc00, %eax		# x2APIC enabled?
+	jnz	nox2apic
+	rdmsr
+	andl	$0xfffff3ff, %eax	# Disable xAPIC & x2APIC
+	wrmsr
+	orl	$0x800, %eax			# Enable xAPIC only
+	wrmsr
 nox2apic:
 
 	# Pick the correct stack.
--- a/system/smp.c
+++ b/system/smp.c
@@ -515,6 +515,38 @@ static bool start_cpu(int cpu_num)
     return true;
 }
 
+bool check_x2apic(void)
+{
+    uint32_t reg[4];
+
+    if (cpuid_info.flags.x2apic) {
+        rdmsr(MSR_IA32_APIC_BASE, reg[0], reg[1]);
+#ifdef __x86_64__
+        if ((reg[0] & IA32_APIC_ENABLED) && (reg[0] & IA32_APIC_EXTENDED)) {
+            // x2APIC is pre-enabled by BIOS.
+            cpuid(0xb, 0, &reg[0], &reg[1], &reg[2], &reg[3]);
+
+            // if BSP APIC ID is > 255 (should not happen outside a VM), give up.
+            if (reg[3] > 255) {
+                return false;
+            }
+
+            // Try to disable x2APIC and re-enable xAPIC.
+            rdmsr(MSR_IA32_APIC_BASE, reg[0], reg[1]);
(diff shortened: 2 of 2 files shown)


boot/startup64.S
```diff
@@ -152,6 +152,25 @@ startup:
...
+
+	movl	$0x1, %eax
+	cpuid
+	notl	%ecx
+	andl	$0x200000, %ecx
+	jnz	nox2apic				# x2APIC supported?
+	movq	$0x1b, %rcx
+	rdmsr
+	notl	%eax
+	andl	$0xc00, %eax		# x2APIC enabled?
```
debrouxl, line 165: Here and below, word operations like I did as part of my NX changes ( #241 ) ?
andw $0xc00,%ax
andw $0xf3ff,%ax
orw $0x800,%ax

debrouxl:
I had recently downloaded the x2APIC spec, but didn't read it as far / thoroughly as noticing that an x2APIC can be reverted to xAPIC mode. Maybe you have information sources more readable / directly practical than the x2APIC spec or the Linux kernel sources, too ?

Since an x2APIC -> xAPIC revert would make it possible to run memtest86+ in SMP mode on a range of computers with < 256 HW threads where x2APIC is enabled by default nowadays (even a 2S/16C/16T DL380 Gen9, per my latest test in #12 ), yes, I'd say that it's a good idea to integrate this, until a full implementation of x2APIC replaces it, providing the ability to test the reachability of all memory by all cores on computers with 256+ threads (e.g. 2S EPYC 9654 & 9634 servers), which, per our previous discussions, is more interesting than 256+ simultaneous threads on the same relatively small memory range of 1 GB.

Whether to run a round of tests for this patch with your usual beta-tester contacts is up to you - it depends on the effort it takes and the timeline :)

x86fr:
> I had recently downloaded the x2APIC spec, but didn't read it as far / thoroughly as noticing that an x2APIC can be reverted to xAPIC mode. Maybe you have information sources more readable / directly practical than the x2APIC spec or the Linux kernel sources, too ?

Nope, there is state transition diagram in the Intel arch64 manual showing the allowed state change in APIC mode. You can switch directly from x2APIC to APIC, but you can disable APIC entirely then re-enable xAPIC only.

This function is present in the linux kernel : 

https://github.com/torvalds/linux/blob/7c3dc440b1f5c75f45e24430f913e561dc82a419/arch/x86/kernel/apic/apic.c#L1775-L1789

But I don't get the point of the second wrmsr. It seems useless (it works without it) and the comment is also about re-enabling (not done there). Maybe a typo when they split the code.

Hypoon:
I tested this PR on one of my systems (2x Xeon E5-2683 v4), and it recognized all threads with x2APIC still enabled (awesome!). The second test was disabled by default, though... is that concerning?

debrouxl:
Good to have a confirmation that the x2APIC disabling patch works on your computer as well :)

Test 1 has been disabled by default on all platforms since early January (commit 186ef6e91333badfe315c9b2616991bd238da3f6 ), as test 2 provides the same coverage, especially on 64-bit platforms.

system/smp.c
```diff
@@ -515,6 +515,38 @@ static bool start_cpu(int cpu_num)
     return true;
 }
 
+bool check_x2apic(void)
```
debrouxl, line 518: +static, given that this is a function with a single caller, which hasn't (yet) been exported to smp.h.

dorrellmw:
I'm curious about a few things.
- First, why does it say "SMP: 231T" in the first photo? I don't see any logical relationship between 231 and the number of cores (68), threads (272), or the thread limit (255/256).
- Second, why do CPUs with more than 192 threads need a fix? 160 threads is fine, but 200 threads is not?
- Third, with HT disabled, why do you still need to reduce the thread count?

I feel like I'm missing a piece, and there's a good chance that one concept might answer all three questions.

debrouxl:
Do non-temporal writes help on that amusing platform ? The corresponding patch is on both the nontemporal (currently a373e971752d4739656670a74f381bb49f55554b ) and numa branches. It helped on all of the computers supporting non-temporal writes on which I tested them so far.

x86fr:
> I'm curious about a few things.
> 
>     * First, why does it say "SMP: 231T" in the first photo? I don't see any logical relationship between 231 and the number of cores (68), threads (272), or the thread limit (255/256).
> 
>     * Second, why do CPUs with more than 192 threads need a fix? 160 threads is fine, but 200 threads is not?
> 
>     * Third, with HT disabled, why do you still need to reduce the thread count?
> 
> 
> I feel like I'm missing a piece, and there's a good chance that one concept might answer all three questions.

All is related. Without x2APIC, the APIC ID is limited to 255, so IDs >255 are not reachable. IDs are not attributed in a linear way (from 0 to 255) for each logical core, so on this 272T platform, you have many logical and physical cores with IDs outside a 8-bit boundary (ie: more than 272-256). On this platform, only 231 logical cores (out of 272) and 58 physical cores (out of 68) have IDs in the range 0-255.

Anyway, the limit of 192 is arbitrary, based on real-world APIC ID distribution on some high-core count platforms, The real solution is to fully support x2APIC, but until this is achieved, limiting to 192T helps avoiding cases where APs APIC ID is > 255. Technically, we can handle this, but I prefer spending time on the real thing (supporting x2APIC) than on a temporary fix.

x86fr:
> Do non-temporal writes help on that amusing platform ? The corresponding patch is on both the nontemporal (currently [a373e97](https://github.com/memtest86plus/memtest86plus/commit/a373e971752d4739656670a74f381bb49f55554b) ) and numa branches. It helped on all of the computers supporting non-temporal writes on which I tested them so far.

I'll try to test it, but I'm not sure the result will be relevant: this platform doesn't have any DRAM module on the motherboard (the 16GB are LLC).

system/smp.c
x86fr: Yes, but this function is not executed (because the asm code in startup.S has already been executed on the BSP, so x2apic is already disabled when the code goes here). I'm doing some tests to find the right way to call check_x2apic() externally from startup.S, but it's more complex than expected.

Hypoon:
@x86fr, any news on the test result? Anything else blocking this PR?

x86fr:
Yes. Tests failed. There is still a lot of work to do on this patch and unfortunately, servers with x2apic are not common along betatesters.

Hypoon:
Is there anything I can do to assist? Would one extra betatester (who has x2apic hardware) make a difference here?

x86fr:
I'm looking for servers where this PR fails. That's usually with > 128 cores.

Hypoon:
Would they need > 128 physical cores, or would > 128 logical cores be sufficient?

x86fr:
> 128 threads

Hypoon:
Will you need direct access to the server, or could I run the tests? They're not mine, but I may have access to some 2x 64-core AMD EPYC machines, which should be 256 threads total.

x86fr:
Definitely, this fix works fine on some hardware, but requires too much work to debug and betatest for general use.

I'd rather spend time adding x2APIC support than a complex workaround to disable it.

PR #280, opened 2023-03-06
Add Memory Controller Registers Read to get current DRAM Timings/Frequency

x86fr:
This is a first skeleton for Memory controller registers reading. I've added Intel HSW as a first implementation example. 
It will also be used later for ECC polling. 

Before adding many more controllers, I would like to hear your thoughts (especially @debrouxl) about the structure.

Also, I would like to find a way to use a wildcard in Makefile to avoid adding each new file in system/mch individually. I've tried some tricks, but they failed. I'm sure some Makefile gurus can help here :)


debrouxl:
Looks reasonable, if skeletal / incomplete for now: `memctrl_init()` contains the only reference to `imc_type`, `IMC_HSW` and `IMC_SKL`.
The code supports a single IMC, but I think I understand why: in the worst case, maintaining cache coherency between portions of the computer whose respective IMCs run at different frequencies would align all portions on the slowest portion anyway.

I would simply eliminate the wildcard issue in the Makefile, which I don't know how to fix without looking for examples, by #include'ing every `system/mch/*.c` file into `system/memctrl.c`, above `memctrl_init()`, and making the functions static. The functions have a single caller, so the compiler should be given a chance to inline (some of) them into their caller, like in `system/smbus.c`, which contains all functions in a single source file.

I'll have other comments inline.

system/memctrl.h
```diff
@@ -0,0 +1,56 @@
...
+ * (running DRAM configuration, ECC, ...) and other
+ * platform-specific data
+ *
+ *//*
+ * Copyright (C) 2004-2023 Sam Demeulemeester.
+ */
+
+typedef struct imc_infos {
+    uint16_t    freq;
+    char        *type;
```
debrouxl, line 17: Boo, padding :'(
Swap `type` and `freq` to almost eliminate it in this struct.

system/memctrl.h
```diff
@@ -0,0 +1,56 @@
...
+    uint16_t    tRCD;
+    uint16_t    tRP;
+    uint16_t    tRAS;
+} imc_info_t;
+
+typedef enum {
+    ECC_ERR_NONE,
+    ECC_ERR_CORRECTED,
+    ECC_ERR_UNCORRECTED
+} ecc_error_typo_t;
```
debrouxl, line 30: I think there's a _typo_ here :)

system/memctrl.h
```diff
@@ -0,0 +1,56 @@
...
+
+typedef struct ecc_status {
+    bool                ecc_enabled;
+    ecc_error_typo_t    err_type;
+    uint64_t            err_adr;
```
debrouxl, line 35: More padding happens in this struct, to align err_type on a 32-bit boundary and err_addr on a 64-bit boundary.

system/memctrl.c
```diff
@@ -0,0 +1,53 @@
...
+
+#include <stdbool.h>
+
+#include "config.h"
+#include "cpuinfo.h"
+
+#include "memctrl.h"
+#include "mch/mch.h"
+
+imc_info_t imc;
```
martinwhitaker, line 17: I think you need to statically initialise this structure, otherwise you won't get a relocation record for the imc.type field. I needed to do that for the static structures in usbhcd.c.

x86fr:
I added another line in Makefile for `IMC_SRC` because the one-line is a bit too obfuscated

x86fr:
IMC reading for Core 4th -> 10th Gen (Desktop) now added. Betatesting welcome 👌

Target for merge is IMC reading Core 1st -> 13th implementation + Zen 1/2/3/4.
Then I'll work on ECC polling.

x86fr:
I need some help here because I'm loosing my mind. 😵

I'm trying to add support for Core 2nd/3rd Gen. There is nothing complex here, just some small similar code addition to the previous Core 4th -> 10th gen code in this branch.

In the newly added [system/mch/intel_snb.c](https://github.com/memtest86plus/memtest86plus/pull/280/files#diff-322829cbe358ddd2793a459c826976e2532c9e071e6836a6bdb35cdc01e0c021) L84-93, you will find 3 lines of commented code.

As soon as I uncomment these 3 lines (and only these 3 lines), the whole system hard-lock **way after** this code, much later in the init process (usually in `initial_config();` ). No exception, no crash, no reboot, just hard-lock (even reset doesn't work, only 5-sec press on power can power down the PC).

This issue only occurs in 64-bit mode. In 32-bit mode, everything is fine. I've tried numerous things without success. The exact place where the hard-lock occur is erratic.

To me, it looks like an overflow somewhere, leading to a hardware MMIO register being overwritten, but I can't understand the process exactly (hard-coding a random value instead of the pointer read solves the issue, adding a ton of dummy code is fine, etc.)

Any guess greatly appreciated.

Motherboard is Asus P8B WS. Various CPU tested. Will try on another motherboard asap.

martinwhitaker:
I notice you haven't called `map_region()`, so are you sure you have access to the right memory-mapped locations?

Looking at the code (and disassembling to confirm), with the 3 lines commented out the compiler will optimise the code to only read the second byte pointed to by `ptr`, whereas with the 3 lines included it will optimise the code to only perform a single 32-bit read. That's the only difference I can see that would cause the different behaviour.

x86fr:
Yes, I've tried with and without map_region but that doesn't make any difference. The expected bytes are read correctly in all cases (checked with the printf just after that). 

I don't understand why the hard lock occurs so much later in the process, and at a random location. It occurs at the exact same location later, but this location depends of where the memctrl code is placed in the init steps.

Changing the BSP stack size (eg from 16KB to 32KB) also changes the hard lock location. 🫠

debrouxl:
Weird behaviour on SNB related to memory below 1 MB reminds me of the older SNB workaround mentioned in https://www.phoronix.com/news/Linux-Always-Reserve-1MB . It may well not be spot on, but nothing else comes to my mind about this matter for now :)

x86fr:
The ptr is usually located at 0xFED00000 :-/ But I don't think these lines are directly related, as the hard lock can be provoked before them, just by changing the code size.

The fact that the 32-bit exact same code is fine is maybe related to the smaller code size or stack size.

x86fr:
This thing is definitely haunted.

If anyone has any idea how the hell is this possible... ⬇️

```
    // RAS-To-CAS (tRCD)
    imc.tRCD = *ptr & 0xF;
    prints(20,0,"1");

    // RAS Precharge (tRP)
    imc.tRP = (*ptr >> 4) & 0xF;
    prints(20,2,"2");

    // RAS Active to precharge (tRAS)
    // imc.tRAS = (*ptr >> 16) & 0xFF;
```

Display "1" then hard lock immediately...


```
    // RAS-To-CAS (tRCD)
    imc.tRCD = *ptr & 0xF;
    // prints(20,0,"1");

    // RAS Precharge (tRP)
    imc.tRP = (*ptr >> 4) & 0xF;
    // prints(20,2,"2");

    // RAS Active to precharge (tRAS)
    // imc.tRAS = (*ptr >> 16) & 0xFF;
```

Displays "1 2" and hardlock way later in the init process...

```
    // RAS-To-CAS (tRCD)
    // imc.tRCD = *ptr & 0xF;
    prints(20,0,"1");

    // RAS Precharge (tRP)
    // imc.tRP = (*ptr >> 4) & 0xF;
    prints(20,2,"2");

    // RAS Active to precharge (tRAS)
    // imc.tRAS = (*ptr >> 16) & 0xFF;
```

Displays "1 2" and run fine.

debrouxl:
Two other ideas which combine: use 16-bit or 32-bit reads (the hardware might be is disturbed by oversized reads), and do not re-read *ptr multiple times (for efficiency anyway, plus it might also disturb the hardware) ? This goes for the other functions as well, BTW.
```
    uint16_t mch_cfg, reg0, reg1;
    uint32_t tempvar;

    // Get DRAM Ratio
    mch_cfg = *(uint16_t*)((uintptr_t)mmio_reg + SNB_REG_MCH_CFG);
    if ((mch_cfg >> 8) & 1) {
        dram_ratio = (float)(mch_cfg & 0x1F) * (100.0f / 100.0f);
    } else {
        dram_ratio = (float)(mch_cfg & 0x1F) * (133.34f / 100.0f);
    }

    reg0 = *(uint16_t*)((uintptr_t)mmio_reg + SNB_REG_MAIN_CHAN0);

    reg1 = *(uint16_t*)((uintptr_t)mmio_reg + SNB_REG_MAIN_CHAN1);

    tempvar = *(uint32_t*)((uintptr_t)mmio_reg + SNB_REG_TIMING);

    // CAS Latency (tCAS)
    imc.tCL = (tempvar >> 8) & 0xF;

    // RAS-To-CAS (tRCD)
    imc.tRCD = tempvar & 0xF;

    // RAS Precharge (tRP)
    imc.tRP = (tempvar >> 4) & 0xF;

    // RAS Active to precharge (tRAS)
    imc.tRAS = (tempvar >> 16) & 0xFF;
```

martinwhitaker:
Correction to my previous comment. The non-working code only optimises to a 32-bit read in a 32-bit build. It optimises to a 64-bit read in a 64-bit build (because `uintptr_t` is 64 bits). This also applies to your latest trials. So it looks like doing a 64-bit read to those hardware locations is causing the fault. Why that should have a delayed effect is anyone's guess.

debrouxl:
Yeah, it would seem that the hardware does not like the 64-bit access in the latter.
And these repeated reads of the same memory location are needlessly inefficient, please try to use single reads of whichever width the hardware sees fit :)

x86fr:
This is -O1 just for debugging. These reads ARE optimized out from -O2 and up.

Some more crazy tests :

This works:

```
    // CAS Latency (tCAS)
    uint32_t *ptr2 = (uint32_t*)((uintptr_t)mmio_reg + SNB_REG_TIMING);
    imc.tCL = (*ptr2 >> 8) & 0xF;
 107:	0f b6 8e 01 40 00 00 	movzbl 0x4001(%rsi),%ecx
 10e:	83 e1 0f             	and    $0xf,%ecx
 111:	66 89 48 0c          	mov    %cx,0xc(%rax)

    // RAS-To-CAS (tRCD)
    imc.tRCD = *ptr2 & 0xF;
 115:	44 0f b7 86 00 40 00 	movzwl 0x4000(%rsi),%r8d
 11c:	00 
 11d:	41 83 e0 0f          	and    $0xf,%r8d
 121:	66 44 89 40 10       	mov    %r8w,0x10(%rax)

    // RAS Precharge (tRP)
    imc.tRP = (*ptr2 >> 4) & 0xF;
 126:	44 8b 8e 00 40 00 00 	mov    0x4000(%rsi),%r9d
 12d:	41 c1 e9 04          	shr    $0x4,%r9d
 131:	41 83 e1 0f          	and    $0xf,%r9d
 135:	66 44 89 48 12       	mov    %r9w,0x12(%rax)

    // RAS Active to precharge (tRAS)
    imc.tRAS = (*ptr2 >> 16) & 0xFF;
 13a:	0f b6 96 02 40 00 00 	movzbl 0x4002(%rsi),%edx
 141:	66 89 50 14          	mov    %dx,0x14(%rax)
```

This hard-lock :

```
    // CAS Latency (tCAS)
    uintptr_t __attribute__((aligned(4))) *ptr2 = (uintptr_t*)((uintptr_t)mmio_reg + SNB_REG_TIMING);
    imc.tCL = (*ptr2 >> 8) & 0xF;
 107:	0f b6 8e 01 40 00 00 	movzbl 0x4001(%rsi),%ecx
 10e:	83 e1 0f             	and    $0xf,%ecx
 111:	66 89 48 0c          	mov    %cx,0xc(%rax)

    // RAS-To-CAS (tRCD)
    imc.tRCD = *ptr2 & 0xF;
 115:	44 0f b7 86 00 40 00 	movzwl 0x4000(%rsi),%r8d
 11c:	00 
 11d:	41 83 e0 0f          	and    $0xf,%r8d
 121:	66 44 89 40 10       	mov    %r8w,0x10(%rax)

    // RAS Precharge (tRP)
    imc.tRP = (*ptr2 >> 4) & 0xF;
 126:	4c 8b 8e 00 40 00 00 	mov    0x4000(%rsi),%r9
 12d:	49 c1 e9 04          	shr    $0x4,%r9
 131:	41 83 e1 0f          	and    $0xf,%r9d
 135:	66 44 89 48 12       	mov    %r9w,0x12(%rax)

    // RAS Active to precharge (tRAS)
    imc.tRAS = (*ptr2 >> 16) & 0xFF;
 13a:	0f b6 96 02 40 00 00 	movzbl 0x4002(%rsi),%edx
 141:	66 89 50 14          	mov    %dx,0x14(%rax)
```

What's crazy is MCHBAR address is > 32-bit (but always > 4 GB in real world)

image

x86fr:
Haha, I missed it! 😵‍💫 Definitely that. It works now. 

Thanks for your help to both of you, I was about to go crazy...

x86fr:
Zen3 should work but deserves testing. Zen4 to go.

x86fr:
Remaining : Core 11th Gen / Zen3 / Zen4

debrouxl:
What about 1st-gen Core i*, which you mentioned above ? Sure, Nehalem & Westmere became available about 14-12 years ago, so initial ECC corrected error reporting framework + support for several newer generations is probably higher priority :)

There's a significant chronological gap between Sandy Bridge and Zen 1; implementing Fam15h (Bulldozer, Piledriver, etc.) IMC support would help close it. However, if your target for Intel IMC support really is Nehalem & Westmere, reaching chronological parity with those on the AMD side would require implementing AMD IMC support all the way back to the late K10 (Lisbon, Magny-Cours, etc.)...

Several other notes:
* I still think that the `system/mch/*.c` files should be #include'd by `system/memctrl.c: simpler Makefiles, faster compilation, possibly smaller binary;
* in `system/cpuinfo.c::determine_imc()`'s `switch` statements, many `case` whose bodies contain `imc_type = IMC_*` assignments are effectively dead code, since the assigned values are never handled in `system/memctrl.c` or elsewhere. As such, they could / should be commented out until they're actually used;
* unrelated to the IMC, but in the same file: few of the code paths in `system/cpuinfo.c::determine_cpu_model()` aren't dead code in a 64-bit build, since the vast majority of processors listed there aren't 64-bit-capable. It would be great if the brand string were available in each and every 64-bit-capable processor model, but is it ?

x86fr:
> What about 1st-gen Core i*, which you mentioned above ? Sure, Nehalem & Westmere became available about 14-12 years ago, so initial ECC corrected error reporting framework + support for several newer generations is probably higher priority :)
> 
> There's a significant chronological gap between Sandy Bridge and Zen 1; implementing Fam15h (Bulldozer, Piledriver, etc.) IMC support would help close it. However, if your target for Intel IMC support really is Nehalem & Westmere, reaching chronological parity with those on the AMD side would require implementing AMD IMC support all the way back to the late K10 (Lisbon, Magny-Cours, etc.)...

Unfortunately, I don't have the spare time needed to support all µarch released since 20 years at once. I'll do that later, one by one. This is a big PR and it already requires a lot of additional betatesting.

> Several other notes:
> 
> * I still think that the `system/mch/*.c` files should be #include'd by `system/memctrl.c: simpler Makefiles, faster compilation, possibly smaller binary;

I still think that the increased readability worth the few bytes and few microseconds saved at compile ;) Also, don't forget that many code will soon be added in /mch/, especially ECC code.

> * in `system/cpuinfo.c::determine_imc()`'s `switch` statements, many `case` whose bodies contain `imc_type = IMC_*` assignments are effectively dead code, since the assigned values are never handled in `system/memctrl.c` or elsewhere. As such, they could / should be commented out until they're actually used;

They will be at some point. Adding quirks and support for IMC/SMB or whatever future features is faster when the IMC type is already identified correctly (and tested). We're not limited in code size anymore, especially for so little bytes.

> * unrelated to the IMC, but in the same file: few of the code paths in `system/cpuinfo.c::determine_cpu_model()` aren't dead code in a 64-bit build, since the vast majority of processors listed there aren't 64-bit-capable. It would be great if the brand string were available in each and every 64-bit-capable processor model, but is it ?

We must have a more global debate about the possibility of diverging paths between the 32 and the 64 bit code. I will open a discussion about this to avoid hijacking this PR.

debrouxl:
Well, building 1 source file instead of currently 1+5, soon 1+10 or more, is quite a bit more than "few microseconds saved at compile", and the benefit scales with the number of files ;)

To be clear, I'm merely suggesting the following change, which is an improvement over what I suggested earlier (yielding an even simpler Makefile which doesn't have to deal with mch/*), and could be furthered by removing redundant #include:
```
index 30f4e16..ddcbedd 100644
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -41,9 +41,6 @@ SYS_OBJS = system/acpi.o \
            system/vmem.o \
            system/xhci.o
 
-MCH_SRCS = $(wildcard ../system/mch/*.c)
-MCH_OBJS = $(subst ../,,$(MCH_SRCS:.c=.o))
-
 LIB_OBJS = lib/barrier.o \
            lib/div64.o \
            lib/print.o \
@@ -69,13 +66,12 @@ APP_OBJS = app/badram.o \
            app/interrupt.o \
            app/main.o
 
-OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(MCH_OBJS) $(LIB_OBJS) $(TST_OBJS) $(APP_OBJS)
+OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(LIB_OBJS) $(TST_OBJS) $(APP_OBJS)
 
 all: memtest.bin memtest.efi
 
 -include boot/efisetup.d
 -include $(subst .o,.d,$(SYS_OBJS))
--include $(subst .o,.d,$(MCH_OBJS))
 -include $(subst .o,.d,$(LIB_OBJS))
 -include $(subst .o,.d,$(TST_OBJS))
 -include $(subst .o,.d,$(APP_OBJS))
@@ -102,10 +98,6 @@ system/%.o: ../system/%.c
 	@mkdir -p system
 	$(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
 
-system/mch/%.o: ../system/mch/%.c
-	@mkdir -p system/mch
-	$(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
-
 lib/%.o: ../lib/%.c
 	@mkdir -p lib
 	$(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
index 44a1ee9..8518dec 100644
--- a/build64/Makefile
+++ b/build64/Makefile
@@ -20,8 +20,8 @@ SYS_OBJS = system/acpi.o \
            system/cpulocal.o \
            system/ehci.o \
            system/font.o \
-           system/hwctrl.o \
            system/heap.o \
+           system/hwctrl.o \
            system/hwquirks.o \
            system/keyboard.o \
            system/ohci.o \
@@ -41,9 +41,6 @@ SYS_OBJS = system/acpi.o \
            system/vmem.o \
            system/xhci.o
 
-MCH_SRCS = $(wildcard ../system/mch/*.c)
-MCH_OBJS = $(subst ../,,$(MCH_SRCS:.c=.o))
-
 LIB_OBJS = lib/barrier.o \
            lib/print.o \
            lib/read.o \
@@ -68,13 +65,12 @@ APP_OBJS = app/badram.o \
            app/interrupt.o \
            app/main.o
 
-OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(MCH_OBJS) $(LIB_OBJS) $(TST_OBJS) $(APP_OBJS)
+OBJS = boot/startup.o boot/efisetup.o $(SYS_OBJS) $(LIB_OBJS) $(TST_OBJS) $(APP_OBJS)
 
 all: memtest.bin memtest.efi
 
 -include boot/efisetup.d
 -include $(subst .o,.d,$(SYS_OBJS))
--include $(subst .o,.d,$(MCH_OBJS))
 -include $(subst .o,.d,$(LIB_OBJS))
 -include $(subst .o,.d,$(TST_OBJS))
 -include $(subst .o,.d,$(APP_OBJS))
@@ -101,10 +97,6 @@ system/%.o: ../system/%.c
 	@mkdir -p system
 	$(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
 
-system/mch/%.o: ../system/mch/%.c
-	@mkdir -p system/mch
-	$(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
-
 lib/%.o: ../lib/%.c
 	@mkdir -p lib
 	$(CC) -c $(CFLAGS) -Os $(INC_DIRS) -o $@ $< -MMD -MP -MT $@ -MF $(@:.o=.d)
diff --git a/system/mch/amd_k17.c b/system/mch/amd_k17.h
similarity index 98%
rename from system/mch/amd_k17.c
rename to system/mch/amd_k17.h
index 176c9c1..5e74aef 100644
--- a/system/mch/amd_k17.c
+++ b/system/mch/amd_k17.h
@@ -20,7 +20,7 @@
 #define AMD_SMN_UMC_DRAM_TIMINGS1   AMD_SMN_UMC_BAR + 0x204
 #define AMD_SMN_UMC_DRAM_TIMINGS2   AMD_SMN_UMC_BAR + 0x208
 
-void get_imc_config_amd_k17(void)
+static void get_imc_config_amd_k17(void)
 {
     uint32_t smn_reg, offset;
     uint32_t reg_cha, reg_chb;
diff --git a/system/mch/intel_adl.c b/system/mch/intel_adl.h
similarity index 98%
rename from system/mch/intel_adl.c
rename to system/mch/intel_adl.h
index 9f72f63..356dd7e 100644
--- a/system/mch/intel_adl.c
+++ b/system/mch/intel_adl.h
@@ -33,7 +33,7 @@
 #define ADL_MMR_MC_BIOS_REG     0x5E04
 #define ADL_MMR_BLCK_REG        0x5F60
 
-void get_imc_config_intel_adl(void)
+static void get_imc_config_intel_adl(void)
 {
     uint64_t mmio_reg;
     uint32_t cha, chb, offset;
diff --git a/system/mch/intel_hsw.c b/system/mch/intel_hsw.h
similarity index 98%
rename from system/mch/intel_hsw.c
rename to system/mch/intel_hsw.h
index 6ce4c77..8f7eb26 100644
--- a/system/mch/intel_hsw.c
+++ b/system/mch/intel_hsw.h
@@ -20,7 +20,7 @@
 #define HSW_REG_TIMING_CAS  0x4014
 #define HSW_REG_TIMING_RCD  0x4000
 
-void get_imc_config_intel_hsw(void)
+static void get_imc_config_intel_hsw(void)
 {
     uint32_t mmio_reg, mch_cfg, offset;
     uint32_t reg0, reg1;
diff --git a/system/mch/intel_skl.c b/system/mch/intel_skl.h
similarity index 98%
rename from system/mch/intel_skl.c
rename to system/mch/intel_skl.h
index 975b10d..f8a66b0 100644
--- a/system/mch/intel_skl.c
+++ b/system/mch/intel_skl.h
@@ -28,7 +28,7 @@
 #define SKL_MMR_BASE_MASK       0x7FFFFF8000
 #define SKL_MMR_MAD_IN_USE_MASK 0x003F003F
 
-void get_imc_config_intel_skl(void)
+static void get_imc_config_intel_skl(void)
 {
     uint64_t mmio_reg;
     uint32_t reg0, reg1, offset;
diff --git a/system/mch/intel_snb.c b/system/mch/intel_snb.h
similarity index 98%
rename from system/mch/intel_snb.c
rename to system/mch/intel_snb.h
index 2e8d927..9ed30b5 100644
--- a/system/mch/intel_snb.c
+++ b/system/mch/intel_snb.h
@@ -19,7 +19,7 @@
 #define SNB_REG_MCH_CFG     0x5E04
 #define SNB_REG_TIMING      0x4000
 
-void get_imc_config_intel_snb(void)
+static void get_imc_config_intel_snb(void)
 {
     uint32_t mmio_reg, offset;
     uint32_t mch_cfg, reg0, reg1;
diff --git a/system/mch/mch.h b/system/mch/mch.h
index 976117e..43d4ed1 100644
--- a/system/mch/mch.h
+++ b/system/mch/mch.h
@@ -4,18 +4,18 @@
 #define _MCH_H_
 
 /* Memory configuration Detection for AMD K17 (Zen/Zen2) */
-void get_imc_config_amd_k17(void);
+static void get_imc_config_amd_k17(void);
 
 /* Memory configuration Detection for Intel Sandy Bridge */
-void get_imc_config_intel_snb(void);
+static void get_imc_config_intel_snb(void);
 
 /* Memory configuration Detection for Intel Haswell */
-void get_imc_config_intel_hsw(void);
+static void get_imc_config_intel_hsw(void);
 
 /* Memory configuration Detection for Intel Skylake */
-void get_imc_config_intel_skl(void);
+static void get_imc_config_intel_skl(void);
 
 /* Memory configuration Detection for Intel Alder Lake */
-void get_imc_config_intel_adl(void);
+static void get_imc_config_intel_adl(void);
 
 #endif /* _MCH_H_ */
diff --git a/system/memctrl.c b/system/memctrl.c
index f85a441..f49de92 100644
--- a/system/memctrl.c
+++ b/system/memctrl.c
@@ -12,7 +12,11 @@
 #include "cpuinfo.h"
 
 #include "memctrl.h"
-#include "mch/mch.h"
+#include "mch/amd_k17.h"
+#include "mch/intel_snb.h"
+#include "mch/intel_hsw.h"
+#include "mch/intel_skl.h"
+#include "mch/intel_adl.h"
 
 imc_info_t imc = {"UNDEF", 0, 0, 0, 0, 0, 0, 0};
 
```
which has no impact on the readability of the code in `memctrl.*` and `mch/*.h` ;)

FTR, unsurprisingly:
```
Before this patch:
 111082   26752  290816  428650   68a6a build32/memtest_shared
 115473   28392  294688  438553   6b119 build64/memtest_shared

After this patch:
 110703   26752  290816  428271   688ef build32/memtest_shared
 115118   28392  294688  438198   6afb6 build64/memtest_shared
```

x86fr:
I don't want to sound rude, but I don't understand why you're focusing on this while your skills in optimization may be much more useful (IMO) on other part of the code (eg: converting quirks to array, new tests, ...). On my side, I don't want to spend time on the same debate over and over again. There is so much to code and so little time. 

We don't use the same coding style and it's perfectly fine (as long as we both follow the basic coding rules). Having complex code in headers makes my eyes tearing blood. That's not where I expect to find code and it makes my brain bug. I think it's much easier to debug/disassemble a separate object file than having all of them merged in a single file, especially for low-level code like here. I'm glad Martin didn't merged all the code from the different USB controllers into a same file that way, or I wouldn't have been able to understand anything. This is just my opinion and I agree that others may think different, or have different fetish regarding code.

But basically, as long as the coding style doesn't make any noticeable difference in real-world (like saving a few bytes and a few milliseconds), the submitter is free to code the way he wants.

x86fr:
Well, something is very wrong with GitHub right now... Support page shows a lot of incident and my last (big) commit seems corrupted after push. 😢 I have a working copy locally, I'll check tomorrow.

x86fr:
Totally FUBAR'd the branch with a crappy rebasing attempt. What a mess... 😒
Well, it's probably better to delete this branch and create a new one from scratch now.

debrouxl:
I'll see about fixing the branch through interactive rebase or something along those lines, but after sleep :)

x86fr:
Don't waste too much time on this mess (it will be squashed ultimately). 
I've already recreated a clean branch here: https://github.com/memtest86plus/memtest86plus/tree/x86fr/imc

debrouxl:
Don't worry, it was about 10 minutes of work, which I did now, in fact.

The diff between the commit before the botched merge and my interactively rebased version (reordered commits by pairs of pick + squash, removed redundant commit messages):
```
$ git diff 81126354b319be409694510d433b3a199ecd04ad origin/memctrl_rebase_fixed
diff --git a/system/cpuinfo.c b/system/cpuinfo.c
index 13bbc51..c01546c 100644
--- a/system/cpuinfo.c
+++ b/system/cpuinfo.c
@@ -492,7 +492,7 @@ static void determine_imc(void)
                 imc.family = IMC_KBL;         // Core 7/8/9th Gen (Kaby/Coffee/Comet Lake)
                 break;
               case 0xB:
-                imc_type = IMC_ADL_N;       // Core 12th Gen (Alder Lake-N - Gracemont E-Cores only)
+                imc.family = IMC_ADL_N;       // Core 12th Gen (Alder Lake-N - Gracemont E-Cores only)
                 break;
               default:
                 break;
```
as well as the diff between my branch and your branch:
```
$ git diff origin/x86fr/memctrl origin/memctrl_rebase_fixed
(empty output)
```
are the expected ones, AFAICS.

In order to fix this PR, you could have used
```
git reset --hard x86fr/memctrl [or however you called your local branch tracking origin/x86fr/memctrl] origin/memctrl_rebase_fixed
git push -f origin x86fr/memctrl [or however you called your local branch tracking origin/x86fr/memctrl]
```

x86fr:
I don't understand everything, but it looks easier than expected. Trying these commands now...

x86fr:
Fixed clone #306

PR #281, opened 2023-03-09
Minor Makefile tidying

turboencabulator:
- Remove `AS` definition since it's no longer used.
- Use `$@` in each rule that produces a file.
- Fix a trivial difference between the 32/64-bit Makefiles in how the object list is sorted.

### DIFF
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -1,4 +1,3 @@
-AS = as -32
 CC = gcc
 
 GIT = git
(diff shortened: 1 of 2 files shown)


debrouxl:
That's reasonable. On my debrouxl/experiments branch, I had also fixed the order of system/* files in build64/Makefile, but never integrated that change to the main branch.

x86fr:
Fine with AS removal and object reordering, but I would avoid using `$@` for single target. 
Makefile syntax is already obfuscated enough to me. 

eg: you changed
`dd if=memtest.bin of=floppy.img bs=1474560 conv=notrunc` 
to 
`dd if=memtest.bin of=$@ bs=1474560 conv=notrunc`
but this can be shrinked again to
`dd if=$< of=$@ bs=1474560 conv=notrunc`
at the expense of even more impact on immediate readability. 
Finally, many targets can also be shrinked even more by using replacement

PS: Using automatic shortcut variables like `$@` for wildcard targets is perfectly fine.

turboencabulator:
I was considering using `$<` and friends as well to cut down on some of the repetitiveness, but:

- `$<` has different meaning on some flavors of make, I think it's BSD make that only recognizes it in pattern rules.  But then again these two files are already using a lot of GNU make syntax, so we can assume GNU make behavior.
- `$@` is pretty much universal, and considered good practice to ensure that a file with the same name as the target gets generated.

x86fr:
Time to make a decision about this one. On one side, I prefer the original code, which I find easier to read directly (Makefile syntax is unnatural to me with these macro/alias.). On the other side, yeah, this may be considered as "good practice", but in that case, this will open the door to other PRs with even more tidying with $<, substitutions, etc. This is endless and will ultimately result in a highly obfuscated Makefile... 

In the end, I will close it unless another maintainer wants to merge it.

PR #290, opened 2023-03-30
cpuinfo: Add support for Vortex86

jonathan-teh:
Hardcode cache for family 5, use CPUID cache info for family 6.

### DIFF
--- a/system/cpuinfo.c
+++ b/system/cpuinfo.c
@@ -92,7 +92,7 @@ static void determine_cache_size()
         /* fall through */
       case 'V':
         // Vortex86
-        if (cpuid_info.version.family < 6) {
+        if (cpuid_info.vendor_id.str[0] == 'V' && cpuid_info.version.family < 6) {
           // Only family 6 have cache info
           break;
         }
--- a/system/cpuinfo.c
+++ b/system/cpuinfo.c
@@ -867,9 +867,13 @@ static void determine_cpu_model(void)
                 break;
             }
             break;
-
+          case 6:
+            // Other family 6 models have brand string
+            cpu_model = "Vortex86EX";
+            l1_cache = 16;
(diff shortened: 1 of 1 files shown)


jonathan-teh:
Tested on eBOX 3350DX2 with a Vortex86DX2. The 5/8/6 CPUID seems to be shared with the Vortex86MX (and is identified that way in Linux).
memtest-ebox3350dx2

system/cpuinfo.c
```diff
@@ -90,6 +90,13 @@ static void determine_cache_size()
         }
         // Zhaoxin CPU only
         /* fall through */
+      case 'V':
+        // Vortex86
+        if (cpuid_info.version.family < 6) {
```
x86fr, line 95: 'fall through' are dangerous and shouldn't be stacked if possible. What if Zhaoxin release a CPU with family < 6 ?

jonathan-teh: I've added an additional check to ensure that the fall through only handles Vortex86 CPUs.
Actually, looking at the CPUID dumps from Zhaoxin family 7 CPUs, it has cache info in both Intel-style and AMD-style extended information so that could be simplified though perhaps for another PR (some might have a different vendor ID string as well that starts with a space ` `).

x86fr: There is also a potential issue with VIA Nano CPUs without padlock (which returns "VIA VIA VIA " instead of the usual "CentaurHauls"). I will check the cpuz database if this may fail somewhere.

CPU detection is always more complex than initially expected 😵‍💫

x86fr: Finally, I didn't find any of these CPU in the CPU-Z database (with cpuid "VIA VIA VIA ", so I think it's safe to merge as-is.

PR #291, opened 2023-03-31
Enable using custom objcopy during build.

KimonHoffmann:
When cross compiling the unqualified `objcopy` found on the host might not be able to deal with x86 binaries.

### DIFF
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -1,5 +1,6 @@
 AS = as -32
 CC = gcc
+OBJCOPY = objcopy
 
 GIT = git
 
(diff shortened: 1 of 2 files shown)


KimonHoffmann:
Exactly.
This is a common customization point for packages that support cross compilation.

The way the Makefile is written `LD` is already customizable and TBH I question the rationale behind explicitly setting `AS` and `CC` at the beginning of the Makefiles as those belong to the set of (just like `LD`) [implicitly defined Variables](https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html).

martinwhitaker:
> The way the Makefile is written `LD` is already customizable and TBH I question the rationale behind explicitly setting `AS` and `CC` at the beginning of the Makefiles

Because they weren't intended to support cross-compiling, they were used to set the preferred tools and command line options, e.g. `as -32` for 32-bit builds and `as -64` for 64-bit builds.

PR #292, opened 2023-03-31
Fix compiler used by "Build and tests" action

KimonHoffmann:
The "Build and tests" workflow claimed to build with GCC as well as clang, but was not set up in a way to actually use the specified compiler during the build.

* The environment variable was not named correctly (`C` vs. `CC`)
* Even if the environment variable was named correctly its value would not have been used, as the `Makefile`s currently explicitly assign the value `gcc` to the `CC` variable, and environment variables don't override explicitly assigned variables (only variables specified as part of the arguments do).

After fixing this it turned out that the current state of source and/or `Makefile`s is not compatible with clang, and thus the second commit in this PR disables it for now.

### DIFF
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        compiler: [gcc, clang]
+        compiler: [gcc]
         os: [ubuntu-20.04, ubuntu-22.04]
         wordsize: [32, 64]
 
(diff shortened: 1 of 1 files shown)


x86fr:
Nice for clang (never noticed it), but I'm not sure if a specific iso path is useful.

KimonHoffmann:
> [...] but I'm not sure if a specific iso path is useful.

I don't think so either, but kept it because that was the way it was done before. If you want I can remove that axis of the matrix and just make iso the default target!?

x86fr:
Yes. Probably better and easier.

KimonHoffmann:
* Made the change and reworded the commit message accordingly.
* Added `Signed-off-by:` to the second commit, which I previously forgot.

Issue #302, opened 2023-05-01
RFE: Display DDR5 memory modules temperature (per SPD5118 / SPD5 Hub standard)


rutsky:
IIUC, DDR5 memory modules have standartised built-in temperature sensors (https://www.jedec.org/standards-documents/docs/jesd300-5b). It would nice to display memory modules as part of the memtest86+ test, similarly how CPU temperature is displayed.

PassMark MemTest86 displays "RAM Temp":

![image](https://user-images.githubusercontent.com/46573/235441947-5b67e8da-675f-494c-b013-4ee1e32d2398.png)

Per JESD300-5B standard, temperature of each module can be obtained from the same I2C addresses as SPD from the following registers (© JEDEC, sharing parts per https://en.wikipedia.org/wiki/Fair_use):

![image](https://user-images.githubusercontent.com/46573/235442948-d017433a-af19-4e54-9242-f5de09adac80.png)

![image](https://user-images.githubusercontent.com/46573/235443301-a8da7642-ab8e-428d-bab2-14c6e5953cfc.png)

rutsky:
A reference for DDR3 modules temperature:

`jc42` Linux kernel module [description](https://www.kernel.org/doc/html/latest/hwmon/jc42.html) says:

> This driver implements support for JEDEC JC 42.4 compliant temperature sensors, which are used on many DDR3 memory modules for mobile devices and servers.

And points to http://www.jedec.org/sites/default/files/docs/4_01_04R19.pdf for the datasheet.

x86fr:
DDR3 temperature reporting is vastly different and not really standardized. DDR5 temperature is, but not all modules have the required sensor (it's linked to the version of the embedded PMIC).

Anyway, I will add it at some point.

rutsky:
Compared to PassMark MemTest86, It would be nice if the temperature would be shown for each dimm separately. E.g. in my setup of 4 DIMMs I discovered that during the test modules in the middle are up to 10C hotter, than modules at the edge, which may be an issue.

Also PassMark MemTest86 after the test completion displays overall temperature stats for min/max/average temperature - may be worth to have this behavior in MemTest86+ too:

![PXL_20230501_121942039~2](https://user-images.githubusercontent.com/46573/235548262-8dd5970a-6130-4568-a5a4-a1ad5e6aac5b.jpg)

rutsky:
[...]

PR #304, opened 2023-05-07
Display DDR5 SPD5 Hub memory modules temperature

rutsky:
This is a PoC for reading and displaying DDR5 SPD5 Hub temperature for https://github.com/memtest86plus/memtest86plus/issues/302

Example of how it looks on my system:
image

Marking as PoC as there are few unresolved items (there are also TODOs in the code):

1. Currently displaying the temperature to the right of SPD information about memory modules. This is a bad location, as it overlaps with error strings that Memtest will display on memory errors, plus potentially overlaps with long description lines from SPD. What would be the best place to display this information and how much of information should be displayed?

2. Temperature readings require I2C/SMBus IO operations, which are generally slow (e.g. `usleep()` in [`ich5_process()`](https://github.com/memtest86plus/memtest86plus/blob/5dcd424ea7afb857c1171e747ef064d98d26afeb/system/smbus.c#L1587)). Currently the updates are hacked into the same place where CPU temperature updates are written, which happen only once per second. But I'm unsure if sleeping during I2C/SMBus IO operation would negatively affect other stuff (actual test runs?) done in the same thread.

3. Temperature reads always using `ich5`/`i801` code path. Not sure if other SMBus implementations need to be supported.

4. SPD5 Hub without temperature sensor. Can be easily checked in the code by reading the relevant register.

Tested on ASUS Z690 with 4x32 GB Kingston Fury DDR5 (listed on the screenshot above).

Not really tested:

* Running with less than 4 sticks (e.g. when used slots list has gaps, like only 1 and 3 slots used).
* How errors will looks like (but should overlap with the displayed temperature strings).
* Output to TTY (only tested with output to actual screen).

### DIFF
--- a/system/smbus.c
+++ b/system/smbus.c
@@ -28,6 +28,14 @@ static uint16_t extra_initial_sleep_for_smb_transaction = 0;
 static int8_t spd_page = -1;
 static int8_t last_adr = -1;
 
+struct spd5_temp {
+    // Is current slot has DDR5 memory with a temperature sensor.
+    bool found;
+    // Maximum observed DIMM temperature.
+    int16_t max_temp;
+};
+static struct spd5_temp spd5_temps[MAX_SPD_SLOT] = {0};
+
 // Functions Prototypes
 static void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len);
 
@@ -52,6 +60,7 @@ static bool piix4_get_smb(uint8_t address);
 static bool ich5_get_smb(void);
 static bool ali_get_smb(uint8_t address);
 static uint8_t ich5_process(void);
+static uint8_t ich5_read_byte(uint8_t smbus_adr, uint16_t adr);
 static uint8_t ich5_read_spd_byte(uint8_t adr, uint16_t cmd);
 static uint8_t nf_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
 static uint8_t ali_m1563_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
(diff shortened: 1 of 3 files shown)


rutsky:
Pre-built 64-bit ISO for the last commit.

[memtest_2023-05-07_fdd0f893.zip](https://github.com/memtest86plus/memtest86plus/files/11415933/memtest_2023-05-07_fdd0f893.zip)

x86fr:
Thanks for your effort, but the code to read the temperature sensor is very simple and doesn't really require a PoC.

The hardest part here is to target the right module correctly and address the I2C controller the right way. 

You really can't just consider that all platforms with DDR5 are ICH5-based and run these commands without any sanity checks. 

This is extremely dangerous, especially on DDR5 platforms where it's very easy to overwrite the SPD (I did that myself while debugging the DDR5 code). 

**I really advice anyone against using this code. Any code addition with the potential to physically break the hardware really deserves a proper implementation and a lot of betatesting**

rutsky:
> The hardest part here is to target the right module correctly and address the I2C controller the right way.

Temperature data obtained roughly in the same way as DDR5 SPD (at least per specification). Provided code reuses the same approach used for DDR5 SPD reads and, assuming it's correct, I see the problem of targeting and addressing the I2C controller as already solved problem.

> You really can't just consider that all platforms with DDR5 are ICH5-based and run these commands without any sanity checks.

Yes, this is correct. The code can be changed to explicitly check for the used controller, the same way as `get_spd()` does. I didn't invest time into it in the initial implementation as it's not required on my platform and I could have followed up on this during the PR review process.

> This is extremely dangerous, especially on DDR5 platforms where it's very easy to overwrite the SPD (I did that myself while debugging the DDR5 code).
>
> I really advice anyone against using this code. Any code addition with the potential to physically break the hardware really deserves a proper implementation and a lot of betatesting

While I agree on the sentiment that such changes may be dangerous, I want to note that, compared to SPD reading on DDR5, **reading of temperature data does not require I2C writes**. I'm quite confident that if Memtest86+ successfully read SPD data on a system (something that is already implemented), it will also be able to read temperature data on the same system.

Could you share more details on the dangers of only using I2C reads (assuming the check for ICH5 would be in place or we run in a system with ICH5)?

system/smbus.c
```diff
@@ -62,6 +71,73 @@ static inline uint8_t bcd_to_ui8(uint8_t bcd)
...
+    //   high=0b00000000 low=0b00001100 is  0.75 C
+    //   high=0b00000000 low=0b00000000 is  0.00 C
+    //   high=0b00011111 low=0b11111100 is -0.25 C
+    //   high=0b00011111 low=0b11111000 is -0.50 C
+
+    // Read 11-bit of two-complement value from high/low bytes into 16-bit
+    // variable.
+    int16_t temp4 = ((high & 0b11111) << 6) | ((low >> 2) & 0b111111);
+    // Extend sign from bit 11 to make it 16-bit signed integer.
+    temp4 = (temp4 << 4) >> 4;
```
rutsky, line 98: Offsets there should be 5, not 4. Well, I stated that the negative temperatures are not tested :)

PR #310, opened 2023-05-18
Carry on after a unsafe read on a non-existent MSR

x86fr:
Catch GPF following a RDMSR instruction (usually from a non-existent msr) and allow the program to continue. 
Set a return value of 0 (as eax/edx return for the higher level function) and leave a small mark on top-right corner to indicate something went wrong at some point.

A cleaner way to do this would be to use an exception table similar to the linux kernel, but it's probably overkill for Memtest86+.

### DIFF
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -25,8 +25,10 @@
 // Constants
 //------------------------------------------------------------------------------
 
-#define HLT_OPCODE  0xf4
-#define JE_OPCODE   0x74
+#define HLT_OPCODE      0xf4
+#define JE_OPCODE       0x74
+#define RDMSR_OPCODE    0x320f
+#define WRMSR_OPCODE    0x300f
 
 #ifdef __x86_64__
 #define REG_PREFIX  "r"
@@ -149,6 +151,28 @@ void interrupt(struct trap_regs *trap_regs)
         parity_error();
         return;
 #endif
+    }
+
+    // Catch GPF following a RDMSR instruction (usually from a non-existent msr)
+    // and allow the program to continue. A cleaner way to do this would be to
+    // use an exception table similar to the linux kernel, but it's probably
+    // overkill for Memtest86+. Set a return value of 0 and leave a small mark
+    // on top-right corner to indicate something went wrong at some point.
+    if (trap_regs->vect == 13) {
+        uint16_t *pc = (uint16_t *)trap_regs->ip;
+        if (pc[0] == RDMSR_OPCODE) {
+            uintptr_t *return_addr;
+            if (cpuid_info.flags.lm == 1) {
+                return_addr = (uintptr_t *)(trap_regs->sp - 40);
+            } else {
+                return_addr = (uintptr_t *)(trap_regs->sp - 12);
+            }
+            *return_addr += 2;
+            trap_regs->ax = 0;
+            trap_regs->dx = 0;
+            printc(0, SCREEN_WIDTH - 1, '*');
+            return;
+        }
     }
 
     spin_lock(error_mutex);
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -25,10 +25,26 @@
 // Constants
 //------------------------------------------------------------------------------
 
-#define HLT_OPCODE      0xf4
+#define INT_DIVBY0     0
+#define INT_RSV        1
+#define INT_NMI        2
+#define INT_BRKPOINT   3
+#define INT_OVERFLOW   4
+#define INT_BOUND      5
+#define INT_UNDEFOP    6
+#define INT_DEVNA      7
+#define INT_DOUBLEFLT  8
+#define INT_FPUSEGOVR  9
+#define INT_INVDTSS    10
+#define INT_SEGFLT     11
+#define INT_STKSEGFLT  12
+#define INT_GPF        13
+#define INT_PAGEFLT    14
+
+#define HLT_OPCODE      0xF4
 #define JE_OPCODE       0x74
-#define RDMSR_OPCODE    0x320f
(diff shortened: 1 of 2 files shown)


x86fr:
For single core, I would just have added 2 to ip, but I copy/pasted the 40 & 12 offsets on stack address return from martin's code.
I don't fully understand how these offsets are calculated. Anyway, both (inclusion simple ip+2) seem to work

debrouxl:
The 0x320f and 0x300f constants had me briefly confused, but then I recognized the little-endian encoding for a 16-bit read :)
Nothing else comes to my mind at that point.

FWIW, in order to avoid copying the return address modification code around, I'd have written it the following way (compile-tested only):
```
diff --git a/app/interrupt.c b/app/interrupt.c
index b2da527..9368b0c 100644
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -25,8 +25,10 @@
 // Constants
 
-#define HLT_OPCODE  0xf4
-#define JE_OPCODE   0x74
+#define HLT_OPCODE      0xf4
+#define JE_OPCODE       0x74
+#define RDMSR_OPCODE    0x320f
+#define WRMSR_OPCODE    0x300f
 
 #ifdef __x86_64__
 #define REG_PREFIX  "r"
@@ -105,6 +107,7 @@ void interrupt(struct trap_regs *trap_regs)
 {
     // Get the page fault address.
     uintptr_t address = 0;
+    uintptr_t return_offset = 0;
     if (trap_regs->vect == 14) {
 #ifdef __x86_64__
         __asm__(
@@ -119,7 +122,7 @@ void interrupt(struct trap_regs *trap_regs)
 #endif
     }
 
-    if (trap_regs->vect == 2) {
+    else if (trap_regs->vect == 2) {
         uint8_t *pc = (uint8_t *)trap_regs->ip;
         if (pc[-1] == HLT_OPCODE) {
             // Assume this is a barrier wakeup signal sent via IPI.
@@ -132,17 +135,19 @@ void interrupt(struct trap_regs *trap_regs)
         // signals, so we should find the HLT opcode either at pc[0] or at pc[2]. If we find
         // it, adjust the interrupt return address to point to the following instruction.
         if (pc[0] == HLT_OPCODE || (pc[0] == JE_OPCODE && pc[2] == HLT_OPCODE)) {
+            if (pc[2] == HLT_OPCODE) {
+                return_offset = 3;
+            } else {
+                return_offset = 1;
+            }
+do_return:
             uintptr_t *return_addr;
             if (cpuid_info.flags.lm == 1) {
                 return_addr = (uintptr_t *)(trap_regs->sp - 40);
             } else {
                 return_addr = (uintptr_t *)(trap_regs->sp - 12);
             }
-            if (pc[2] == HLT_OPCODE) {
-                *return_addr += 3;
-            } else {
-                *return_addr += 1;
-            }
+            *return_addr += return_offset;
             return;
         }
 #if REPORT_PARITY_ERRORS
@@ -151,6 +156,22 @@ void interrupt(struct trap_regs *trap_regs)
 #endif
     }
 
+    // Catch GPF following a RDMSR instruction (usually from a non-existent msr)
+    // and allow the program to continue. A cleaner way to do this would be to
+    // use an exception table similar to the linux kernel, but it's probably
+    // overkill for Memtest86+. Set a return value of 0 and leave a small mark
+    // on top-right corner to indicate something went wrong at some point.
+    else if (trap_regs->vect == 13) {
+        uint16_t *pc = (uint16_t *)trap_regs->ip;
+        if (pc[0] == RDMSR_OPCODE) {
+            trap_regs->ax = 0;
+            trap_regs->dx = 0;
+            printc(0, SCREEN_WIDTH - 1, '*');
+            return_offset = 2;
+            goto do_return;
+        }
+    }
+
     spin_lock(error_mutex);
 
     clear_message_area();
```

x86fr:
Replacing all the return code offset computation with a simple `trap_regs->ip += 2;` also seem to work fine. :-/

x86fr:
TODO: Generate GPF from APs and check how it returns.

martinwhitaker:
> Replacing all the return code offset computation with a simple `trap_regs->ip += 2;` also seem to work fine. :-/

Have you tried that with a 32-bit build running on a 64-bit CPU? See the comments around int_handler64 in startup32.S for why it won't work.

I should have added a comment about that in interrupt.c too.

app/interrupt.c
```diff
@@ -151,6 +153,28 @@ void interrupt(struct trap_regs *trap_regs)
...
+            uintptr_t *return_addr;
+            if (cpuid_info.flags.lm == 1) {
+                return_addr = (uintptr_t *)(trap_regs->sp - 40);
+            } else {
+                return_addr = (uintptr_t *)(trap_regs->sp - 12);
+            }
+            *return_addr += 2;
+            trap_regs->ax = 0;
+            trap_regs->dx = 0;
+            printc(0, SCREEN_WIDTH - 1, '*');
```
martinwhitaker, line 173: Directly hacking the screen makes it hard for someone else to know where this spurious character comes from. Better to add a macro in display.h.

x86fr:
> > Replacing all the return code offset computation with a simple `trap_regs->ip += 2;` also seem to work fine. :-/
> 
> Have you tried that with a 32-bit build running on a 64-bit CPU?

Ah yes, obviously!

app/interrupt.c
```diff
@@ -121,7 +137,7 @@ void interrupt(struct trap_regs *trap_regs)
 #endif
     }
 
-    if (trap_regs->vect == 2) {
+    if (trap_regs->vect == INT_NMI) {
```
debrouxl, line 140: +else, here and below ? `trap_regs->vect` isn't modified by that code.

PR #312, opened 2023-05-20
Fix GitHub Actions 404 Errors due to lack of apt-get update

x86fr:


x86fr:
404 solved

debrouxl:
Right, that (or `apt update`, on modern versions of `apt`) is the appropriate fix.

PR #313, opened 2023-05-21
Fix memory error addresses displayed by badram reporting mode

debrouxl:
Per the discussion in #308.

I have tested this fix on top of the `numa` branch, since that's where I'm running the debugger using a modified version of  #177. I used the following patch to interfere with the contents of the memory at an address in the third 1 GB directory page, in order to generate multiple errors when there's enough memory:
```
diff --git a/tests/own_addr.c b/tests/own_addr.c
index 23db5ec..f20a777 100644
--- a/tests/own_addr.c
+++ b/tests/own_addr.c
@@ -125,7 +125,7 @@ static int pattern_check(int my_cpu, testword_t offset)
             if (!offset) {
                 do {
                     testword_t expect = (testword_t)p;
-                    testword_t actual = read_word(p);
+                    testword_t actual = (p == (testword_t *)2345678912) ? (testword_t)0 : read_word(p);
                     if (unlikely(actual != expect)) {
                         data_error(p, expect, actual, true);
                     }
@@ -134,7 +134,7 @@ static int pattern_check(int my_cpu, testword_t offset)
             else {
                 do {
                     testword_t expect = (testword_t)p + offset;
-                    testword_t actual = read_word(p);
+                    testword_t actual = (p == (testword_t *)2345678912) ? (testword_t)0 : read_word(p);
                     if (unlikely(actual != expect)) {
                         data_error(p, expect, actual, true);
                     }
```

When executing `debug_memtest.sh` with the `MACHINE=2S12CWestmere MEMSIZE=8192` environment variables:
* the individual error reporting mode keeps reporting correct errors (2345678912 = 0x8bd03840, and QEMU has a memory hole between 2 GB and 4 GB):
memtest86plus_badram_fix_test_20230521_04
* without this fix, the badram error reporting mode incorrectly merges error reports to a single error at a physical address inside the memory hole range:
memtest86plus_badram_fix_test_20230521_02
* with this fix, the badram error reporting mode reports the same set of 6 errors as the individual error reporting mode:
memtest86plus_badram_fix_test_20230521_03

### DIFF
--- a/app/badram.c
+++ b/app/badram.c
@@ -28,6 +28,7 @@
 #include "display.h"
 
 #include "badram.h"
+#include "memsize.h"
 
 //------------------------------------------------------------------------------
 // Constants
@@ -38,9 +39,9 @@
 
 // DEFAULT_MASK covers a uintptr_t, since that is the testing granularity.
 #ifdef __x86_64__
-#define DEFAULT_MASK (UINTPTR_MAX << 3)
+#define DEFAULT_MASK (UINT64_MAX << 3)
 #else
-#define DEFAULT_MASK (UINTPTR_MAX << 2)
+#define DEFAULT_MASK (UINT64_MAX << 2)
 #endif
 
 //------------------------------------------------------------------------------
@@ -48,8 +49,8 @@
 //------------------------------------------------------------------------------
 
 typedef struct {
-    uintptr_t   addr;
-    uintptr_t   mask;
+    uint64_t   addr;
+    uint64_t   mask;
 } pattern_t;
 
 //------------------------------------------------------------------------------
@@ -70,7 +71,7 @@ static int          num_patterns = 0;
 /*
  * Combine two addr/mask pairs to one addr/mask pair.
  */
-static void combine(uintptr_t addr1, uintptr_t mask1, uintptr_t addr2, uintptr_t mask2, uintptr_t *addr, uintptr_t *mask)
+static void combine(uint64_t addr1, uint64_t mask1, uint64_t addr2, uint64_t mask2, uint64_t *addr, uint64_t *mask)
 {
     *mask = COMBINE_MASK(addr1, mask1, addr2, mask2);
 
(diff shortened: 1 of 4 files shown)


x86fr:
Looks good! I'll check soon with betatesters what's the best strategy for merging this (testing alone or merge everything and test as a whole).

debrouxl:
I should have removed the `addr` parameter of `badram_insert()`, since it's unused, which I missed by not looking properly at the output of `make -j`. I'll do it, it's not like an unused function parameter will impact beta tests.

x86fr:
You can push it (I will not merge this before next week-end)

app/badram.c
```diff
@@ -227,11 +228,11 @@ void badram_init(void)
     }
 }
 
-bool badram_insert(uintptr_t addr)
+bool badram_insert(testword_t page, testword_t offset, uintptr_t addr)
 {
     pattern_t pattern = {
-            .addr = addr,
-            .mask = DEFAULT_MASK
+        .addr = (page << PAGE_SHIFT) + offset,
```
martinwhitaker, line 234: This will overflow for addresses > 4G on a 32-bit build. I think we need to change the `addr` field of `pattern_t` to be a `uint64_t`. That may require other changes...

debrouxl:
In the previous iteration, the badram reporting mode was still broken, due to the usage of an `uintptr_t`-sized mask and improper address printing code. I switched the mask to `uint64_t`, which is probably more readable than keeping `uintptr_t` and using a bunch of casts to `uint64_t` where needed.
The current implementation of `vprintf` only supports `uintptr_t` width, which is 32-bit on x86, so I switched to two pairs of 32-bit prints in the last hunk of `app/badram.c`. There _could_ be different code paths for x86 and x86_64 here.

Tested in badram error reporting mode with emulated 4S 4C PIII, 6 GB RAM, without KVM acceleration, and 2S 12C Westmere, 6 GB RAM, with KVM acceleration. The former has a 1 GB memory hole, the latter has a 2 GB memory hole.
memtest86plus_badram_fix_test_20230524_01
memtest86plus_badram_fix_test_20230524_02

andersk (approved):
My laptop developed a bad bit at `0x14f8884b8`. This patch corrects the badram output from `0x000000008f8884b8,0xfffffffffffffff8` to `0x000000014f8884b8,0xfffffffffffffff8`.

Issue #339, opened 2023-09-07
VIA C3 Samuel 2 wrong L2 cache size detection


nickos222:
Hello guys,
I believe there is a bug in the L2 cache size detection logic.
Memtest86 v6.20 detects 16MB when in reality the L2 cache size is 64kb.

Older (ancient) memtest versions detect the l2 size correctly:
Memtest-86 v3.5b
Memtest86+ v4.20

I had some old floppy disks with these earlier versions and decided to see if they still worked.

x86fr:
Can you please confirm the CPUID of your CPU?

nickos222:
Here you are:

vendor_id	: CentaurHauls
cpu family	: 6
model		: 7
model name	: VIA Samuel 2
stepping	: 1

Almost no linux 586 image can boot with this cpu anymore!
It is very difficult to get additional info for now, nothing really works decently or at all with this cpu and with 512mb ram.
I have tried many distros and most failed with oops or ileggal instruction messages.
Something is broken in recent linux kernel for this cpu.

debrouxl:
For years, both the kernel and the user-space in most 32-bit x86 Linux distros have been built for 686-class computers, which assumes support for instructions such as the CMOV family. When running such software combinations on hardware which does not support CMOV (e.g. AMD K6-1, I tried that in the past), many illegal instructions occur, indeed.

martinwhitaker:
Mageia still supports 586-class machines.

nickos222:
According to VIA documentation the Extended CPUID L2 Cache Configuration encoding found on Samuel 2 differs from the Intel/AMD ones.

Intel/AMD use ECX bits 31:16 to report L2 Cache size in 1K units
VIA Samuel 2 uses ECX bits 31:24 to report L2 Cache size in 1K units and ECX bits 23:16 to report L2 associativity 

Memtest cpuid code erroneously takes into account the additional bits 23:16 for the l2 size and the size from 64kb becomes 16mb.

Memtest shall only use ECX bits 31:24 to report L2 Cache size on VIA C3 Samuel 2 processors (Family 6, Model 7, Stepping 0~7)

VIA C3 Ezra (Family 6, Model 7, Stepping 8+) and C3 Ezra-T (Family 6, Model 8, Stepping 0~7) processors may also use samuel2 l2 cache config encoding,
but I have neither the hardware nor any documentation to verify it.

[...]

PR #341, opened 2023-09-15
cpuinfo: Fix VIA Samuel 2 L2 cache

jonathan-teh:
Centaur family 6 models 7/8 only use the MSB to report L2 cache size. Fix it by right-shifting by 8.

### DIFF
--- a/system/cpuinfo.c
+++ b/system/cpuinfo.c
@@ -80,6 +80,10 @@ static void determine_cache_size()
         if (cpuid_info.version.family == 5 || cpuid_info.version.family == 6) {
             l1_cache = cpuid_info.cache_info.l1_d_size;
             l2_cache = cpuid_info.cache_info.l2_size;
+            if (cpuid_info.version.family == 6 && (cpuid_info.version.model == 7 || cpuid_info.version.model == 8)) {
+                // Samuel 2/Ezra/Ezra-T
+                l2_cache >>= 8;
+            }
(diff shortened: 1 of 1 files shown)


jonathan-teh:
Fixes #339.
@nickos222 please test.

x86fr:
This must be hardcoded for 2 reasons : 1/ these VIA have writable cpuid registers and some BIOS (like MSI-A6309) do the shift at boot time and re-write cpuid to ensure compatibility. 2/ this doesn't work on many Engineering Sample because VIA messed with cpuid a lot in this era.

nickos222:
> This must be hardcoded for 2 reasons : 1/ these VIA have writable cpuid registers and some BIOS (like MSI-A6309) do the shift at boot time and re-write cpuid to ensure compatibility. 2/ this doesn't work on many Engineering Sample because VIA messed with cpuid a lot in this era.

What is the FMS of these eng. samples ? It cannot be 671 like my c3, right ?
FMS 671 works on MS-6309 V2.1 , I see no reason to set l2 size it directly if the eng. samples do not have FMS 671.

x86fr:
6.7.0/1/2/8. A6309MS V2 had better BIOS than V1. Some older BIOS >>= 8 with C5B but not C5C. Some other do the opposite. VIA platforms are quite messy. That's an issue we also had with cpu-z decades ago.

PR #343, opened 2023-09-23
Add "singlepass" option

gnubie2:
Normally memtest86+ runs forever. Using this command line option causes it to exit/reboot after one pass of the selected tests is performed.

Also adds the feature as a menu option and if enabled from the menu during testing it will cause the memtest86+ to exit/reboot after the current pass is completed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,8 @@ recognised:
       * mmio16 = 16-bit MMIO
       * mmio32 = 32-bit MMIO
     * and *y* is the MMIO address in hex. with `0x` prefix (eg: 0xFEDC9000)
+  * singlepass
+    * performs one pass of selected tests then reboots rather than testing endlessly
 
 ## Keyboard Selection
 
--- a/app/config.c
+++ b/app/config.c
@@ -113,6 +113,8 @@ int             tty_mmio_stride    = 4;                 // Stride for MMIO (regi
 
 bool            err_banner_redraw  = false;             // Redraw banner on new errors
 
+bool            single_pass        = false;
+
 //------------------------------------------------------------------------------
 // Private Functions
 //------------------------------------------------------------------------------
@@ -263,6 +265,8 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "3", 2) == 0) {
             usb_init_options |= USB_2_STEP_INIT|USB_EXTRA_RESET;
         }
+    } else if (strncmp(option, "singlepass", 11) == 0) {
+        single_pass = true;
     }
 }
 
@@ -879,10 +883,16 @@ void config_menu(bool initial)
             printf(POP_R+8,  POP_LI, "<F6>  Temperature %s", enable_temperature ? "disable" : "enable ");
             //if (no_temperature) set_foreground_colour(WHITE);
             printf(POP_R+9,  POP_LI, "<F7>  Boot trace %s",  enable_trace  ? "disable" : "enable ");
-            prints(POP_R+10, POP_LI, "<F10> Exit menu");
+            printf(POP_R+10, POP_LI, "<F8>  Single pass %s",  single_pass  ? "disable" : "enable ");
+            prints(POP_R+11, POP_LI, "<F10> Exit menu");
         } else {
             prints(POP_R+7,  POP_LI, "<F5>  Skip current test");
-            prints(POP_R+8 , POP_LI, "<F10> Exit menu");
+            if (single_pass) {
+                prints(POP_R+8,  POP_LI, "<F6>  Run tests forever     ");
+            } else {
+                prints(POP_R+8,  POP_LI, "<F6>  End after current pass");
+            }
+            prints(POP_R+9,  POP_LI, "<F10> Exit menu");
         }
 
         if (tty_update) {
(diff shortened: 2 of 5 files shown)


debrouxl:
I think that limiting the number of passes run by memtest86+ can be considered desirable functionality. However:
* functionally, a `maxpasses=<n>` option is more powerful and more flexible than a `singlepass` option :)
* more importantly: the addition of such a change alone breaks the ability to distinguish between "the machine rebooted because the maximum number of passes was reached" and "the machine rebooted because there was a serious error, possibly the effect of overheating". Halting testing after the maximum number of passes was reached, until a human hopefully looks at the result, is possible; making tests run indefinitely can help detecting some long-term thermal warm-up effects, though.

See also #207 / #138.

x86fr:
I agree with @debrouxl. This is a particular case that can't be pushed to main as-is. 

Can you please describe how you use Memtest86+ exactly, as it's probably chain-loaded in a test suite? We need to understand how this can be helpful for general use.

As said previously, the main issue is to distinguish between a genuine "test complete without errors" reboot and an unexpected reboot due to a failure somewhere. Especially, I'm working in ECC support right now, and many modern memory controllers with ECC enabled (eg: Zen 2/3) are configured by default to automatically reboot when an uncorrectable error is found.

gnubie2:
Yes, my use case is generally for this to be used with test automation although it can and may be run outside of that environment.  In all my use cases the only output device is the serial console.  There are three outcomes: 1) A pass message is displayed before rebooting.  2) A failure message is displayed before rebooting, 3) As was mentioned, the system will spontaneously reboot due to an uncorrectable ECC error.  Automation can detect which occurred due to the differences in output.

It should be trivial to add another option to wait for a key press in the event of a pass.  That will make a clear distinction between pass/fail as both types of failures will result in a reboot while the pass would not.  I'll see what my schedule is like tomorrow and knock it out as soon as I can.

As for one pass versus a programmable number, my experience from memory testing is that people either run one pass, since that will almost always catch issues caused by a faulty DIMM, or they run the test for at least 24 hours perhaps even while performing margin testing (varying voltage and temperatures to the extremes.)  There really isn't a strong use case for running it, say, 4 times.  At least in my experience.

debrouxl:
On my side, I usually run 2 passes, sometimes 3, when I'm testing RAM for real (as opposed to benchmarking memtest86+, I mean) :)

oxedions:
Dear @x86fr, dear @debrouxl,

I am also very interested by such feature, and since you requested it, here is a use case.

Basically, we deliver huge HPC clusters to customers and testing blades either at factory, or when delivering, or even during cluster life (new blade in production, cluster cold start after a power cut, etc) is mandatory. Available tools that can run without a Linux kernel is limited, and we currently have to boot a whole diskless image to run some benchmarks and stress tests (Linpack, Stream, etc). But this is an heavy task, that takes a lot of time and energy, and difficult to maintain. Being able to manipulate memtest would be much simpler (this does not prevent deeper/intensive tests, but that would allow to filter 80% of bad nodes/servers). Also, even if we are no more at the age of DDR2 with >10% break after a cold restart, we still have some faulty hardware each time a cluster is restarted after a long power off (power cut, whatever).

Currently, all blades boot via a standard ISC dhcp server, that serves via tftp/http a custom iPXE rom. This iPXE rom uses a chain load mechanism to get instructions from an http server (which can be apache, flask, whatever) and finally load via http a tool (memtest, other), or deploy operating system via the network, or start a diskless boot, or continue boot on local disk.
(see https://github.com/bluebanquise/bluebanquise/blob/master/collections/infrastructure/roles/pxe_stack/images/iPXE_chain.png)


I would like to do the following:

1. Via cli tool, instruct http server to redirect via iPXE a range of nodes to a memtest, with for example 3 pass, and ask that if nodes passed (rebooted), then they should boot let's say on local disk. So I would ask something like that:

```
bluebanquise-bootset -n c[001-500] -b memtest:disk
```

This would set a counter to `0` for all these nodes on the http server that manages iPXE chain.

Then I start all nodes, using IPMI or RedFish.

2. When node c001 for example boot, it will load the iPXE rom, then via iPXE chain mechanism be redirected by http server to memtest. Http server would have some logical that would increment a counter for this node to `1`.

3. c001 does a memtest, 3 pass, and lest say it succeed.

4. c001 reboots, and this time, http server knows its counter is `1` so node is redirected to `disk` instead of memtest.

If at step 3 node fails the memtest, then it would stay at this stage, admin can then connect on the console and check what happened, and use BMC logs to investigate.

We could also launch a 'night' test, so I ask for memtest:poweroff, meaning all nodes that do with success the 3 pass would poweroff at next boot. When I come in the morning, all nodes powered on are to investigate.

In production, when I come to customer with a new blade, I would have a simple mechanism to recheck (because it was already checked at factory) the blade before putting it in production.

I would be ready to code a flask based example, along with documentation, to be stored in this repository for all the server side if it can help others. But I am not sure I have the knowledge to develop the `do X pass and reboot if 0 or stay if 1` feature on the memtest side :blush: 

Note: this does not solves the unexpected reboot stated by @debrouxl, and I see no simple way to cover that. In theory, admin also have to investigate BMC events to check that they are "clear" after the test.

Small note to @x86fr : never had the opportunity to thank you for CPCH articles I could read all these years, so thanks! :wink:

debrouxl:
Thanks for spending time writing this detailed piece of input :)
Besides this post, the first one received three upvotes. This confirms that there's demand for such a feature, among users who know what they're doing and are willing to deal with the consequences.

A possible solution for the halt vs. reboot issue may be to have the user select either, through something like `maxpasses=n,reboot|halt` (or `maxpasses=reboot|halt,n`, if that's easier to parse for memtest86+).

justinclift:
Extra potential option(s) to control the hang vs reboot behaviour on pass/fail sound like a good idea for automation purposes.

x86fr:
merged to #640

PR #351, opened 2023-11-13
Optimize test_mov_inv_walk1 for size

debrouxl:
I noticed that compiling this file with -O3 yields poor code with significant quasi-duplication, and one of the specialized copies of the first hot loop has a `not` which can be avoided by hoisting the computation outside the loop.
AFAICT, this duplication-specialization can be avoided with several simple changes to the code of `test_mov_inv_walk1()`. If these changes had only saved 100 bytes, maybe I wouldn't have bothered, but surprisingly, they saved well over 1 KB with GCC 13 on both x86 and x86-64 (see commit message)... hence the PR targeting `main`.
Gotta offset the cost of added features (growing IMC support, hopefully NUMA, etc.), after all ;)


martinwhitaker:
Looks good to me.

PR #353, opened 2023-11-19
Preliminary ECC support for AMD Zen CPUs

x86fr:
Currently only available for selected AMD Zen CPUs (K17, K19 Rembrandt, K19 Raphael, K19 Vermeer).

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -99,6 +99,8 @@ bool            enable_sm          = true;
 bool            enable_bench       = true;
 bool            enable_mch_read    = true;
 
+bool            enable_ecc_polling = false;
+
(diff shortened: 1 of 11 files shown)


debrouxl:
Good to see this feature being introduced :)

Nitpick: in `poll_ecc_amd_zen()`, for consistency with `get_imc_config_amd_zen()`, the scope of `regl` and `regh` could be reduced, if they're not going to be reused for other purposes later.

x86fr:
I usually declare them function-wide, but I had to restrict the scope in `get_imc_config_amd_zen()` to avoid 'unused' warnings on 32-bit builds.

PR #355, opened 2023-11-19
Reduce padding and relocations

debrouxl:
No functional changes, just reducing / avoiding padding and avoiding costly relocations. The pointers + relocation records caused by the current version of `system/usbhcd.c` and `tests/tests.h` -> `tests/tests.c` are especially bad, considering that the strings are already manually padded to the same length. Just let them be fixed-width strings :)

```
Before/after:
 112654   26752  290816  430222   6908e build32/memtest_shared
 112036   26708  290848  429592   68e18 build32/memtest_shared

 116473   28424  294656  439553   6b501 build64/memtest_shared
 115290   27696  294624  437610   6ad6a build64/memtest_shared
```

Same motivation as #351 and #354 , higher benefit on x86-64 than on x86 this time: nearly 2 KB. Combined with these two other PRs, ~4 KB.
Tested for months on a variety of platforms through the `numa` and `simd` branches.

### DIFF
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -56,7 +56,7 @@
 #define ADR_DIGITS  "8"
 #endif
 
-static const char *codes[] = {
+static const char codes[][13] = {
     "Divide by 0",
     "Debug",
     "NMI",
--- a/system/jedec_id.h
+++ b/system/jedec_id.h
@@ -10,12 +10,12 @@
 #define JEP106_CNT \
     sizeof(jep106)/sizeof(jep106[0])
 
-struct spd_jedec_manufacturer {
+struct __attribute__((packed)) spd_jedec_manufacturer {
     uint16_t jedec_code;
     char *name;
 };
 
-static const struct spd_jedec_manufacturer jep106[] = {
+static const struct __attribute__((packed)) spd_jedec_manufacturer jep106[] = {
     { 0x0001, "AMD" },
 //  { 0x0002, "AMI" },
 //  { 0x0003, "Fairchild" },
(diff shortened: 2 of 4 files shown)


x86fr:
Packing JEP106 is fine. 

About switching to fixed-width array, I usually hate doing that because it's so easy to add overflowing bugs later when you add/change array members. Fortunately, definitions are embedded for hci & interrupts names, so that's acceptable. Potential issues are higher for tests names, but that's still acceptable as we don't add or change tests

debrouxl:
I had started writing a reply about compilers complaining when one overflows a fixed-width char array - I checked for that, with the current set of compiler flags - but it looks like I didn't send it...
With 1) compilers complaining when they should, and 2) the current release tests being successful on that front (otherwise beta-testers would have reported it by now), there's little reason to revert that patch targeting these fixed-width strings later, isn't it ?

In the short term, why not revert the type change for the enum. However, C23 supports strongly typed enums, and current GCC 13 in Debian sid amd64 successfully builds
```
diff --git a/app/config.h b/app/config.h
index b23cb33..9be6f70 100644
--- a/app/config.h
+++ b/app/config.h
@@ -16,7 +16,7 @@
 #include "smp.h"
 #include "cpuid.h"
 
-typedef enum {
+typedef enum : uint8_t {
     PAR,
     SEQ,
     ONE
```
even without passing `-std=c23` to the compiler, which would indicate that this behaviour was backported to `-std=c11` currently used by the memtest86+ Makefiles.
GCC 12 and older refuse to understand that construct, even when passed `-std=c2x`, so while the strongly typed enum construct could be used for the likes of Ubuntu 24.04, the current stable Debian version's default toolchain won't understand it until the next release circa 2025.

PR #356, opened 2023-11-20
hwquirks: Detect cache on VIA VP3/MVP3

jonathan-teh:
Add quirk to detect motherboard cache on VIA VP3/MVP3.

### DIFF
--- a/system/hwquirks.c
+++ b/system/hwquirks.c
@@ -48,6 +48,16 @@ static void asus_tusl2_configure_mux(void)
     outb(0xAA, 0x2E);
 }
 
+static int *get_motherboard_cache(void)
+{
+    if (l2_cache == 0) {
+        return &l2_cache;
+    } else if (l3_cache == 0) {
+        return &l3_cache;
+    }
+    return NULL;
+}
+
 static void get_m1541_l2_cache_size(void)
 {
     if (l2_cache != 0) {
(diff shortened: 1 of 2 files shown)


jonathan-teh:
Tested on FIC VA-503+ (VIA Apollo MVP3) with AMD K6:
cache-k6

and K6-2+:
cache-k6-2+

While the motherboard has an MVP3 (0598h), the BIOS set the host bridge to use the VP3 device ID (0597h). The cache enable and size bits in the cache control registers are identical on both. In fact, they're also the same on the VP1/VPX (0585h) and VP2 (0595h) but I don't have hardware to test on those.

Linux applies a PCI quirk in `drivers/pci/quirks.c` to expose the real device ID:
```c
/*
 * VIA VT82C598 has its device ID settable and many BIOSes set it to the ID
 * of VT82C597 for backward compatibility.  We need to switch it off to be
 * able to recognize the real type of the chip.
 */
static void quirk_vt82c598_id(struct pci_dev *dev)
{
	pci_write_config_byte(dev, 0xfc, 0);
	pci_read_config_word(dev, PCI_DEVICE_ID, &dev->device);
}
DECLARE_PCI_FIXUP_HEADER(PCI_VENDOR_ID_VIA,     PCI_DEVICE_ID_VIA_82C597_0,     quirk_vt82c598_id);
```

As an aside, `get_m1541_l2_cache_size()` should call `get_motherboard_cache()` to support L3 cache when a K6-III/2+/III+ is present but I don't have that hardware to test.

PR #357, opened 2023-11-30
Add error mode for displaying bad pages

0xAdk:
Main motivation for this patch is to be able to block bad pages on windows with it's bcdedit utility.

Since windows wants page numbers and not rages I opted to try and fit as many on screen as I could.

So it's just a list of bad pages:
screenshot of memtest86plus in list bad pages error mode

Max is 70 since each of the 14 lines can hold 5 max length addresses.

### DIFF
--- a/app/bad_pages_list.c
+++ b/app/bad_pages_list.c
@@ -1,0 +1,139 @@
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "display.h"
+
+#include "bad_pages_list.h"
+#include "test.h"
+
+//------------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


erdman:
Came here to see if there had been any discussion of this very feature.  Thanks to 0xAdk for the ready implementation.  Merged this PR into my cloned repo without issue.  Followed project's straightforward docs to build and make an iso.  Dumped to flash drive, and it's working for me as expected (ie perfectly).  Recommend maintainers consider merging this nice feature into main.

x86fr:
Added in #454

Issue #381, opened 2024-02-20
UEFI Incorrect video mode with non normal oriented panel.


howl:
The video output appears rotated with a non normal oriented panel, the device is a Lenovo idepad D330. It's a little special because it has the panel mounted in the non normal way.

I file the bug because some UEFI binaries achieve to display in the correct way, for example the comercial with free to use version counterpart.

If the correct col x row video output mode is selected is possible to have the display as it should but I don't know how memtest86+ check for the mode to use.

If it's possible, could be added a parameter like the usb ones to pass to memtest to force video modes? If not to be in production at least to help in develop a way to auto set the correct mode.

debrouxl:
Thanks for the report, and for gathering information about the available modes :)

I see that there's some debug code inside https://github.com/memtest86plus/memtest86plus/blob/main/boot/efisetup.c . Are you able to change the `#define DEBUG 0` to `#define DEBUG 1`, recompile the memtest86+ binary, and test it ?

howl:
@debrouxl I have done it with 7.00 src, not nightly.

Just booting memtest86+, the first line is echoed by GRUB2:
![photo_2024-02-20_23-04-48](https://github.com/memtest86plus/memtest86plus/assets/758086/5aa5681e-a25c-4ed3-ac18-b1bb0f0ca315)

Hitting to continue:
![photo_2024-02-20_23-04-51](https://github.com/memtest86plus/memtest86plus/assets/758086/117b5a27-b53d-4d1e-b2d1-010286a3ea45)

And then hitting to continue one more time starts memtest86+.

howl:
Is very probably that the other mode shows 480 x 640, but just guessing because of the normal orientation of the panel. This kind of panel mount is annoying, for example using X11 instead of Wayland where you can show the refresh rate moving windows, etc.. the refresh is indeed vertical instead horizontal.

martinwhitaker:
The character-based console modes are not available once you exit the UEFI boot services, so Memtest86+ uses a graphics mode for display. It picks the mode that has the minimum width (in pixels), provided it is at least 640 pixels wide and 400 pixels high. If there is more than one mode with the same width, it picks the first one in the list supplied by the BIOS. To find out what graphics modes are being advertised by the BIOS, try applying this patch
````
diff --git a/boot/efisetup.c b/boot/efisetup.c
index 7e0a510..54bb4b8 100644
--- a/boot/efisetup.c
+++ b/boot/efisetup.c
[...]

PR #383, opened 2024-02-29
Add boot options to perform display rotation and set preferred screen resolution

martinwhitaker:
This addresses issue #381. It's not a perfect fix for that machine, due to a buggy BIOS, but hopefully it will be useful for other 2-in-1 laptops too.

As I needed to add boot command line parsing in the efisetup code, I've taken the opportunity to make the EFI debug code a run-time option instead of a compile-time option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ recognised:
       * usb
       * both
   * screen.mode=*w*x*h* (EFI framebuffer only)
-    * where *w*x*h* is the preferred screen reolution (e.g. 1024x768)
+    * where *w*x*h* is the preferred screen resolution (e.g. 1024x768)
   * screen.mode=bios (EFI framebuffer only)
     * uses the default screen resolution set by the UEFI BIOS
   * screen.rhs-up (graphics mode only)
--- a/README.md
+++ b/README.md
@@ -142,6 +142,16 @@ recognised:
       * legacy
       * usb
       * both
+  * screen.mode=*w*x*h* (EFI framebuffer only)
+    * where *w*x*h* is the preferred screen reolution (e.g. 1024x768)
+  * screen.mode=bios (EFI framebuffer only)
+    * uses the default screen resolution set by the UEFI BIOS
+  * screen.rhs-up (graphics mode only)
+    * rotates the display clockwise by 90 degrees
+  * screen.lhs-up (graphics mode only)
+    * rotates the display anti-clockwiseby 90 degrees
+  * efidebug
+    * displays information about the EFI framebuffer
   * usbdebug
     * pauses after probing for USB keyboards
   * usbinit=*mode*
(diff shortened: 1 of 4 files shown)


x86fr:
I don't think I have any concerned laptop to check in real-world unfortunately, but code looks fine. 
Only potential concern is about the serial output, as it uses the framebuffer.

martinwhitaker:
Serial output uses the text shadow buffer, which is unaffected by these changes.

debrouxl:
It looks reasonable indeed. It's certainly safe to consider that screens which would need 180° rotation are such a cursed thing that it doesn't need to be seriously taken into account :)
I'm not entirely a fan of the partial code duplication in the two `parse_option()`, but I think I understand why it's being done that way.

README.md
```diff
@@ -142,6 +142,16 @@ recognised:
...
       * usb
       * both
+  * screen.mode=*w*x*h* (EFI framebuffer only)
+    * where *w*x*h* is the preferred screen resolution (e.g. 1024x768)
+  * screen.mode=bios (EFI framebuffer only)
+    * uses the default screen resolution set by the UEFI BIOS
+  * screen.rhs-up (graphics mode only)
+    * rotates the display clockwise by 90 degrees
+  * screen.lhs-up (graphics mode only)
+    * rotates the display anti-clockwiseby 90 degrees
```
Nable80, line 152: ```s/anti-clockwiseby/counter-clockwise by/```

system/screen.c
```diff
@@ -10,8 +10,20 @@
...
 
+#include "string.h"
+
 #include "screen.h"
 
+//------------------------------------------------------------------------------
+// Private Types
+//------------------------------------------------------------------------------
+
+typedef enum  __attribute__ ((packed)) {
```
Nable80, line 21: ``-fshort-enums``? The ``packed`` attribute is quiet unusual for anything except structures.

README.md
martinwhitaker: anti-clockwise is correct English.

system/screen.c
martinwhitaker: -fshort-enums would affect all enums, which we don't necessarily want.

README.md
Nable80: My bad, I didn't come across this variant before. It looks like "anti-clockwise" is more common in en_GB, while I'm really spoiled by en_US… I'm sorry I didn't check it before commenting, thank you for clarification.

A missing space before ``by`` should be fixed anyway, shouldn't it?

system/screen.c
Nable80: This is true but it shouldn't break anything for bare-metal code.
OK, I got it, I should check it myself first.

README.md
martinwhitaker: Yes, missing space fixed, thanks for spotting that.

PR #387, opened 2024-03-10
USB-serial support proof of concept

tormodvolden:
This kind of works with a CH340 dongle and an XHCI controller. I have tested it by connecting the CH340 dongle to a PL2303 dongle and running "screen /dev/ttyUSB 19200" on the latter. The CH340 was accessed by memtest86+ running in QEMU.

I can also see it works on plain hardware from the flashing LED on the CH340 <s>but I haven't hooked up the serial line to another computer yet.</s> EDIT: See below screenshots on bare metal.

<s>After a while it stops working and there is an unknown interrupt on CPU0. Quite possibly my XHCI code can need some more work, and there can still be conflicts between keyboard polling and bulk output transfer handling.</s>

Text screenshot copied from "screen":
```
      Memtest86+ v7.00      | QEMU Virtual CPU version 2.5+                     
CLK/Temp: 2810MH206/206*C   | Pass   %                                          
L1 Cache:   64KB            | Test   %                                          
L2 Cache:  512KB            | Test #                                            
L3 Cache:   16MB            | Testing:                                          
Memory  :  122MB            | Pattern:                                          
--------------------------------------------------------------------------------
CPU: 1 Cores 1 Threads    SMP: Disabled   | Time:           Status: Testing  -  
Using:                                    | Pass:  0        Errors: 0           
--------------------------------------------------------------------------------
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
 <ESC> Exit  <F1> Configuration  <Space> Scroll Lock            7.00.ccaa218.x64
```

### DIFF
--- a/system/ehci.c
+++ b/system/ehci.c
@@ -437,6 +437,24 @@ static bool get_data_request(const usb_hcd_t *hcd, const usb_ep_t *ep, const usb
     return do_async_transfer(ws, 3);
 }
 
+// FIXME
+static bool out_data_request(const usb_hcd_t *hcd, const usb_ep_t *ep, const usb_setup_pkt_t *setup_pkt,
+                             const void *buffer, size_t length)
+{
+    workspace_t *ws = (workspace_t *)hcd->ws;
+
+    if (setup_pkt) {
+        build_ehci_qtd(&ws->qtd[0], &ws->qtd[2], EHCI_QTD_PID_SETUP, EHCI_QTD_DT(0), setup_pkt, sizeof(usb_setup_pkt_t));
+        build_ehci_qtd(&ws->qtd[1], &ws->qtd[2], EHCI_QTD_PID_IN,    EHCI_QTD_DT(1), NULL, 0);
+        build_ehci_qtd(&ws->qtd[2], &ws->qtd[2], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+    } else {
+        // what DT ?
+        build_ehci_qtd(&ws->qtd[0], &ws->qtd[0], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+    }
+    build_ehci_qhd(&ws->qhd[0], &ws->qtd[0], ep, false);
+    return do_async_transfer(ws, 3);
+}
+
 static void poll_keyboards(const usb_hcd_t *hcd)
 {
     workspace_t *ws = (workspace_t *)hcd->ws;
@@ -479,6 +497,7 @@ static const hcd_methods_t methods = {
     .configure_kbd_ep    = NULL,
     .setup_request       = setup_request,
     .get_data_request    = get_data_request,
+    .out_data_request    = out_data_request,
     .poll_keyboards      = poll_keyboards
 };
 
--- a/system/ehci.c
+++ b/system/ehci.c
@@ -684,6 +684,9 @@ bool ehci_probe(uintptr_t base_addr, usb_hcd_t *hcd)
     for (int kbd_idx = 0; kbd_idx < num_keyboards; kbd_idx++) {
         usb_ep_t *kbd = &keyboards[kbd_idx];
 
+        if (!IS_EP_INT(kbd))
+            continue;
+
         ehci_qhd_t *kbd_qhd = &ws->qhd[1 + kbd_idx];
         ehci_qtd_t *kbd_qtd = &ws->qtd[3 + kbd_idx];
 
--- a/system/ehci.c
+++ b/system/ehci.c
@@ -437,7 +437,6 @@ static bool get_data_request(const usb_hcd_t *hcd, const usb_ep_t *ep, const usb
     return do_async_transfer(ws, 3);
 }
 
-// FIXME
 static bool out_data_request(const usb_hcd_t *hcd, const usb_ep_t *ep, const usb_setup_pkt_t *setup_pkt,
                              const void *buffer, size_t length)
 {
@@ -445,14 +444,17 @@ static bool out_data_request(const usb_hcd_t *hcd, const usb_ep_t *ep, const usb
 
     if (setup_pkt) {
         build_ehci_qtd(&ws->qtd[0], &ws->qtd[2], EHCI_QTD_PID_SETUP, EHCI_QTD_DT(0), setup_pkt, sizeof(usb_setup_pkt_t));
-        build_ehci_qtd(&ws->qtd[1], &ws->qtd[2], EHCI_QTD_PID_IN,    EHCI_QTD_DT(1), NULL, 0);
-        build_ehci_qtd(&ws->qtd[2], &ws->qtd[2], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+        build_ehci_qtd(&ws->qtd[1], &ws->qtd[2], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+        build_ehci_qtd(&ws->qtd[2], &ws->qtd[2], EHCI_QTD_PID_IN,    EHCI_QTD_DT(1), NULL, 0);
+        build_ehci_qhd(&ws->qhd[0], &ws->qtd[0], ep, false);
+        return do_async_transfer(ws, 3);
     } else {
-        // what DT ?
-        build_ehci_qtd(&ws->qtd[0], &ws->qtd[0], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+        // TODO toggle per endpoint, reset on clear stall
+        static int toggle;
+        build_ehci_qtd(&ws->qtd[0], &ws->qtd[0], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(toggle++ & 1), buffer, length);
+        build_ehci_qhd(&ws->qhd[0], &ws->qtd[0], ep, false);
+        return do_async_transfer(ws, 1);
     }
-    build_ehci_qhd(&ws->qhd[0], &ws->qtd[0], ep, false);
-    return do_async_transfer(ws, 3);
 }
 
 static void poll_keyboards(const usb_hcd_t *hcd)
--- a/system/ehci.c
+++ b/system/ehci.c
@@ -684,6 +684,8 @@ bool ehci_probe(uintptr_t base_addr, usb_hcd_t *hcd)
     for (int kbd_idx = 0; kbd_idx < num_keyboards; kbd_idx++) {
         usb_ep_t *kbd = &keyboards[kbd_idx];
 
+        save_ep(kbd_idx, kbd);
+
         if (!IS_EP_INT(kbd))
             continue;
 
--- a/system/usbhcd.c
+++ b/system/usbhcd.c
@@ -68,6 +68,7 @@ static const hcd_methods_t methods = {
     .configure_kbd_ep    = NULL,
     .setup_request       = NULL,
     .get_data_request    = NULL,
+    .out_data_request    = NULL,
     .poll_keyboards      = NULL
 };
 
@@ -87,6 +88,9 @@ static int num_hcd = 0;
 
 static int print_row = 0;
 static int print_col = 0;
+
+static const usb_hcd_t *print_hcd = NULL;
+static usb_ep_t *print_ep = NULL;
 
 //------------------------------------------------------------------------------
 // Public Variables
@@ -945,4 +949,25 @@ uint8_t get_usb_keycode(void)
         }
     }
     return 0;
+}
+
+bool usb_serial_print(const char *str)
+{
+    const char *packet = str;
+
+    // OUT data method not implemented for all controller types yet
+    if (!print_hcd || !print_hcd->methods->out_data_request)
+        return false;
+
+    while (*packet != '\0') {
+        int i;
+
+        for (i = 0; packet[i] != '\0' && i < print_ep->max_packet_size; i++)
+            ;
+        if (!print_hcd->methods->out_data_request(print_hcd, print_ep, NULL, packet, i)) {
+            return false;
+        }
+        packet = &packet[i];
+    }
+    return true;
 }
--- a/system/usbhcd.c
+++ b/system/usbhcd.c
@@ -89,8 +89,8 @@ static int num_hcd = 0;
 static int print_row = 0;
 static int print_col = 0;
 
-static usb_ep_t uninit_ep;
-static usb_ep_t *print_ep = &uninit_ep;
+static int print_idx = -1;
+static usb_ep_t print_ep = {0};
 static const usb_hcd_t *print_hcd = &hcd_list[0];
 
 //------------------------------------------------------------------------------
@@ -944,7 +944,7 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
             } else if (kbd->reserved == (uint8_t) DEV_SERIAL_CH341) {
                 if (!configure_ch341(hcd, &ep0)) break;
                 print_hcd = hcd;
-                print_ep = kbd;
+                print_idx = kbd_idx;
 
                 print_usb_info(" CH341 serial adapter found on port %i interface %i endpoint %i",
                                port_num, kbd->interface_num, kbd->endpoint_num);
@@ -952,7 +952,7 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
             } else if (kbd->reserved == (uint8_t) DEV_SERIAL_CP210X) {
                 if (!configure_cp210x(hcd, &ep0, kbd->interface_num)) break;
                 print_hcd = hcd;
-                print_ep = kbd;
+                print_idx = kbd_idx;
 
                 print_usb_info(" CP210x serial adapter found on port %i interface %i endpoint %i",
                                port_num, kbd->interface_num, kbd->endpoint_num);
@@ -960,7 +960,7 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
             } else if (kbd->reserved == (uint8_t) DEV_SERIAL_PL2303) {
                 if (!configure_pl2303(hcd, &ep0)) break;
                 print_hcd = hcd;
-                print_ep = kbd;
+                print_idx = kbd_idx;
 
                 print_usb_info(" PL2303 serial adapter found on port %i interface %i endpoint %i",
                                port_num, kbd->interface_num, kbd->endpoint_num);
@@ -972,6 +972,13 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
     }
 
     return keyboard_found;
+}
+
+void save_ep(int kbd_idx, usb_ep_t *ep)
+{
+    if (kbd_idx == print_idx) {
+        print_ep = *ep;
+    }
 }
 
 bool process_usb_keyboard_report(const usb_hcd_t *hcd, const hid_kbd_rpt_t *report, const hid_kbd_rpt_t *prev_report)
@@ -1080,7 +1087,10 @@ bool usb_serial_print(const char *str)
 {
     const char *packet = str;
 
-    if (print_ep == &uninit_ep)
+    if (print_ep.max_packet_size == 0)
+        return false;
+
+    if (print_ep.reserved == (uint8_t) DEV_UNKNOWN)
         return false;
 
     // OUT data method not implemented for all controller types yet
@@ -1090,9 +1100,9 @@ bool usb_serial_print(const char *str)
     while (*packet != '\0') {
         int i;
 
-        for (i = 0; packet[i] != '\0' && i < print_ep->max_packet_size; i++)
+        for (i = 0; packet[i] != '\0' && i < print_ep.max_packet_size; i++)
             ;
-        if (!print_hcd->methods->out_data_request(print_hcd, print_ep, NULL, packet, i)) {
+        if (!print_hcd->methods->out_data_request(print_hcd, &print_ep, NULL, packet, i)) {
             return false;
         }
         packet = &packet[i];
--- a/system/usbhcd.c
+++ b/system/usbhcd.c
@@ -313,6 +313,28 @@ static bool configure_keyboard(const usb_hcd_t *hcd, const usb_ep_t *ep0, int in
     return true;
 }
 
+#define CH341_REQ_SERIAL_INIT 0xA1
+#define CH341_REQ_WRITE_REG   0x9A
+
+static bool configure_ch341(const usb_hcd_t *hcd, const usb_ep_t *ep0)
+{
+    usb_setup_pkt_t setup_pkt;
+
+    // Set up serial line
+    build_setup_packet(&setup_pkt, USB_REQ_TO_DEVICE | USB_REQ_VENDOR, CH341_REQ_SERIAL_INIT, 0, 0, 0);
+    if (!hcd->methods->setup_request(hcd, ep0, &setup_pkt)) {
+        return false;
+    }
+
+    // div/prescale F3 07 gives 923077 baud, set bit 7 for non-buffered
+    build_setup_packet(&setup_pkt, USB_REQ_TO_DEVICE | USB_REQ_VENDOR, CH341_REQ_WRITE_REG, 0x1312, 0xf387, 0);
+    if (!hcd->methods->setup_request(hcd, ep0, &setup_pkt)) {
+        return false;
+    }
+
+    return true;
+}
+
 static bool scan_hub_ports(const usb_hcd_t *hcd, const usb_hub_t *hub, int *num_devices,
                            usb_ep_t keyboards[], int max_keyboards, int *num_keyboards)
 {
@@ -779,7 +801,11 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
                    device->vendor_id, device->product_id, device->device_major, device->device_minor);
 #endif
 
-    device_type = DEV_UNKNOWN;
+    if (device->vendor_id == 0x1a86 && device->product_id == 0x7523) {
+        device_type = DEV_SERIAL_CH341;
+    } else {
+        device_type = DEV_UNKNOWN;
+    }
 
     // Fetch the descriptors for the first configuration into the data transfer buffer. In theory a keyboard device
     // may have more than one configuration and may only support the boot protocol in another configuration, but
@@ -838,6 +864,14 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
                 if (!configure_keyboard(hcd, &ep0, kbd->interface_num)) break;
 
                 print_usb_info(" Keyboard found on port %i interface %i endpoint %i",
+                               port_num, kbd->interface_num, kbd->endpoint_num);
+
+            } else if (kbd->reserved == (uint8_t) DEV_SERIAL_CH341) {
+                if (!configure_ch341(hcd, &ep0)) break;
+                print_hcd = hcd;
+                print_ep = kbd;
+
+                print_usb_info(" CH341 serial adapter found on port %i interface %i endpoint %i",
                                port_num, kbd->interface_num, kbd->endpoint_num);
             }
 
--- a/system/usbhcd.c
+++ b/system/usbhcd.c
@@ -216,7 +216,8 @@ static int get_configuration_descriptors(const usb_hcd_t *hcd, const usb_ep_t *e
 }
 
 static void get_keyboard_info_from_descriptors(const uint8_t *desc_buffer, int desc_length, usb_ep_t keyboards[],
-                                               int max_keyboards, int *num_keyboards)
+                                               int max_keyboards, int *num_keyboards,
+                                               usb_device_type_t *device_type)
 {
     usb_ep_t *kbd = NULL;
     const uint8_t *curr_ptr = desc_buffer + sizeof(usb_config_desc_t);
@@ -239,6 +240,10 @@ static void get_keyboard_info_from_descriptors(const uint8_t *desc_buffer, int d
                 sleep(1);
             }
             if (ifc->class == 3 && ifc->subclass == 1 && ifc->protocol == 1) {
+                *device_type = DEV_KEYBOARD;
+            }
+
+            if (*device_type != DEV_UNKNOWN) {
                 kbd = &keyboards[*num_keyboards];
                 kbd->interface_num = ifc->interface_num;
             } else {
@@ -251,12 +256,25 @@ static void get_keyboard_info_from_descriptors(const uint8_t *desc_buffer, int d
                                (uintptr_t)endpoint->address, (uintptr_t)endpoint->attributes);
                 sleep(1);
             }
-            if (kbd && (endpoint->address & 0x80) && (endpoint->attributes & 0x3) == 0x3) {
+            if (!kbd) {
+                // next
+            } else if (*device_type == DEV_KEYBOARD && (endpoint->address & 0x80) && (endpoint->attributes & 0x3) == 0x3) {
+                // register first endpoint that is IN and interrupt
                 kbd->endpoint_num    = endpoint->address & 0xf;
                 kbd->max_packet_size = endpoint->max_packet_size;
                 kbd->interval        = endpoint->interval;
+                kbd->reserved        = (uint8_t) *device_type;
                 kbd = NULL;
 
+                *num_keyboards += 1;
+            } else if (*device_type >= DEV_SERIAL && !(endpoint->address & 0x80) && (endpoint->attributes & 0x3) == 0x2) {
+                // register first endpoint that is OUT and bulk
+                kbd->endpoint_num    = endpoint->address & 0xf;
+                kbd->max_packet_size = endpoint->max_packet_size;
+                kbd->interval        = 0; // not interrupt endpoint
+                kbd->reserved        = (uint8_t) *device_type;
+
+                kbd = NULL;
                 *num_keyboards += 1;
             }
         }
@@ -741,6 +759,7 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
                                  usb_ep_t keyboards[], int max_keyboards, int *num_keyboards)
 {
     bool keyboard_found = false;
+    usb_device_type_t device_type;
 
     // Set the USB device address. If successful, this also fills in the descriptor for the default control endpoint
     // (ep0) and leaves the device descriptor in the data transfer buffer.
@@ -750,6 +769,13 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
     }
     usb_device_desc_t *device = (usb_device_desc_t *)hcd->ws->data_buffer;
     bool is_hub = (device->class == USB_CLASS_HUB);
+
+#if 0
+    print_usb_info("device %04x:%04x bcdDevice %02x.%02x",
+                   device->vendor_id, device->product_id, device->device_major, device->device_minor);
+#endif
+
+    device_type = DEV_UNKNOWN;
 
     // Fetch the descriptors for the first configuration into the data transfer buffer. In theory a keyboard device
     // may have more than one configuration and may only support the boot protocol in another configuration, but
@@ -784,7 +810,8 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
         int old_num_keyboards = *num_keyboards;
         int new_num_keyboards = *num_keyboards;
         get_keyboard_info_from_descriptors(hcd->ws->data_buffer, hcd->ws->data_length,
-                                           keyboards, max_keyboards, &new_num_keyboards);
+                                           keyboards, max_keyboards, &new_num_keyboards,
+                                           &device_type);
         if (new_num_keyboards == old_num_keyboards) {
             return false;
         }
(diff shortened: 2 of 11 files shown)


tormodvolden:
memtest-qemu-ch340
memtest-qemu-realserial2

tormodvolden:
Now I tested memtest86+ on bare metal with the CH340 linked to a PL2303 on another computer. Works the same.

tormodvolden:
I started with CH340 because it needs very little setup (just a control transfer) to come up, and with a reasonable baud rate by default. The CP2104 I have been testing came up with 625 baud by default :) But it is very easy to set baud rates on these, so it has now been added. My workflow consists of picking the minimally needed steps from the Linux kernel drivers, getting a "hello world" to work in pyusb, and then reimplement it in memtest86plus. I couldn't get my PL2303HX working in pyusb so far. They require a series of magic reads and writes to initialize but I might be missing something else. Finally the ftdixxx ones I am not planning to look at, it is apparently an evil company deliberately destroying people's hardware with their drivers so not something I'd like to promote. They don't seem to be opensource friendly either, leading to a history of a mess of proprietary drivers and free drivers on Linux. I really can recommend CP2104 chipsets from Silicon Labs, they have well working, non-fuzz drivers on all platforms and work in my experience very reliably.

tormodvolden:
Added PL2303 HX, with 921600 baud (like for the CP2104). Maybe it works with some other variants too, but not the "legacy" H variant. Many cheap USB UART cables use this chipset and should work already if they use VID 0x067b. And a lot of other VID/PIDs which use the same chipset can be added.

debrouxl:
Thanks. For the PL2303, I see what you meant by "a series of magic reads and writes" :)

I had forgotten about FTDI's hardware bricking practices. Adding a full-custom driver in a code base like memtest86+ shouldn't make any difference on the well-being of existing end user hardware, but yeah, this kind of manufacturer behaviour isn't good.

system/usbhcd.c
```diff
@@ -338,6 +338,53 @@ static bool configure_cp210x(const usb_hcd_t *hcd, const usb_ep_t *ep0, int inte
...
+    } \
+  } while (0)
+
+static bool configure_pl2303(const usb_hcd_t *hcd, const usb_ep_t *ep0)
+{
+    usb_setup_pkt_t setup_pkt;
+    uint8_t buf1;
+
+    // Magic sequence
+    pl2303_vendor_read(0x8484);
```
debrouxl, line 361: At least the setup packets containing 0x8484 and 0x8383 should be extracted under `static const` form, so as to avoid repeated on-the-fly generation of the packet data. But AFAICT, all of these setup packets containing constant data could be made `static const` as well.

tormodvolden: Thanks for the hint. For now I am deliberately using the same function names and constants as in the kernel source while we pick the parts that we want to use and how much we want to support, and any optimization can follow. I have also refrained from any reorganization and renaming of the memtest86plus code base. If you decide to go for this we'll may want to organize the USB code, like a new folder for USB or for USB-serial and separate out real keyboard handling etc. I don't know if you would prefer to merge then reorganize or the other way around.

tormodvolden:
With the last CH341 fixup, they now all run at ~921600 baud, and if you have a handful of various adapters there is a good chance one of them will work.

What remains the largest annoyance is that it crashes after a while. Maybe because of how the code is moved around in memory and access to certain data structures is not obvious all the time? From below crash dump screenshot you can see rbx and rdi having values suspiciously looking like memory test patterns, while the IP is in issue_normal_trb(), and I believe there should have been data pointers in them:
```
IP is 0x20f578
rbx is 0xfefefefefefefefe
rdi is 0xfefefefefefefefe

(gdb) info sym 0x20f578
issue_normal_trb + 18 in section .text of build64/memtest.debug
(gdb) disass issue_normal_trb
Dump of assembler code for function issue_normal_trb:
   0x000000000020f566 <+0>:     push   %rbx
   0x000000000020f567 <+1>:     mov    %rdi,%rbx
   0x000000000020f56a <+4>:     mov    %rsi,%r8
   0x000000000020f56d <+7>:     mov    %rcx,%r9
   0x000000000020f570 <+10>:    mov    %edx,%ecx
   0x000000000020f572 <+12>:    or     $0x420,%ecx
   0x000000000020f578 <+18>:    mov    0x80(%rdi),%edx
   0x000000000020f57e <+24>:    mov    $0x8,%esi
   0x000000000020f583 <+29>:    call   0x20f37f <enqueue_trb>
   0x000000000020f588 <+34>:    mov    %eax,0x80(%rbx)
   0x000000000020f58e <+40>:    pop    %rbx
   0x000000000020f58f <+41>:    ret
```
crash

debrouxl:
In order to make it possible to test nearly all memory - pinned MMIO ranges are the main exception - memtest86+ relocates its memory image between a memory location within 0-4 MB, and a memory location above 4 MB. The code is position-independent, and the remainder, e.g. function pointers in global structs, needs to be initialized, so that the (expensive) generated relocation records can be used to relocate the program before jumping to the new address.

The crash looks like it could be caused by a straggling pointer to the old memory range before the latest relocation - maybe the result of an asynchronous operation which crossed a relocation taking place, or non-relocated pointer or uintptr_t value.
Not that it's the cause of your current issue, since you're using XHCI and EHCI controllers, but the hcd_methods_t instances in ohci.c and uhci.c need to be updated as well, for the aforementioned relocatability reasons.

tormodvolden:
Thanks! How are the hcd_methods_t instances initialized differently in xhci.c than in ehci.c? And what about the XHCI workspaces, and pointers to them, do they need special care? I was thinking since the keyboarding polling works across relocation, the other USB transfer stuff would as well. Is there any asynchronous code at all?

system/usbhcd.c
```diff
@@ -87,6 +88,9 @@ static int num_hcd = 0;
 static int print_row = 0;
 static int print_col = 0;
 
+static const usb_hcd_t *print_hcd = NULL;
+static usb_ep_t *print_ep = NULL;
```
martinwhitaker, line 92: You need to initialise these two variables with non-null values in order to generate the necessary relocation records, as is done for `hcd_list` above. This is probably the cause of your crashes.

system/ehci.c
```diff
@@ -437,6 +437,24 @@ static bool get_data_request(const usb_hcd_t *hcd, const usb_ep_t *ep, const usb
...
+    if (setup_pkt) {
+        build_ehci_qtd(&ws->qtd[0], &ws->qtd[2], EHCI_QTD_PID_SETUP, EHCI_QTD_DT(0), setup_pkt, sizeof(usb_setup_pkt_t));
+        build_ehci_qtd(&ws->qtd[1], &ws->qtd[2], EHCI_QTD_PID_IN,    EHCI_QTD_DT(1), NULL, 0);
+        build_ehci_qtd(&ws->qtd[2], &ws->qtd[2], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+    } else {
+        // what DT ?
+        build_ehci_qtd(&ws->qtd[0], &ws->qtd[0], EHCI_QTD_PID_OUT,   EHCI_QTD_DT(1), buffer, length);
+    }
+    build_ehci_qhd(&ws->qhd[0], &ws->qtd[0], ep, false);
+    return do_async_transfer(ws, 3);
```
debrouxl, line 455: `do_async_transfer(ws, 3)` even after a single call to `build_ehci_qtd()` in the `else` path raised my eyebrows. Now, I'm not knowledgeable about USB controller usage, and this can be entirely justified, but in that case, I'd say that this warrants a comment :)

x86fr:
> Finally the ftdixxx ones I am not planning to look at, it is apparently an evil company deliberately destroying people's hardware with their drivers so not something I'd like to promote. They don't seem to be opensource friendly either, leading to a history of a mess of proprietary drivers and free drivers on Linux.

I strongly disagree. FTDI is by far the most reliable USB-Serial chipset available on the market. They're used in the majority of industrial/embedded professional products where reliability is crucial. I design such cables, and I would not use anything but FTDI.

Unfortunately, their biggest problem is the quality of their product. As they became the industry-standard solution in the 2000s, a ton of counterfeit chipset emerged, using the VID/PID of original FTDI products. Many quirks in the Linux kernel drivers for FTDI VID were added over the years to support these low-quality counterfeit products. The (in)famous FTDI drivers that bricked many fake cables was a side effect of using the same faked ID. 

Supporting FTDI seems mandatory to me.

martinwhitaker:
> Supporting FTDI seems mandatory to me.

If only because if we don't, we'll get endless requests to do so...

As a professional electronic engineer, I don't consider FTDI evil for trying to clamp down on counterfeit devices. I would also say that they didn't really "brick" the devices - they just changed the USB product ID in those devices so they no longer identified as genuine parts. Given a driver that tolerated fake parts, they would still work.

system/ehci.c
martinwhitaker: The EHCI controller provides separate schedulers for asynchronous (control and bulk) and periodic (isochronous and interrupt) transfers. So this is something inherent in how the EHCI controller operates.

system/usbhcd.c
tormodvolden: I tried, using intptr_t initially set to -1 and casting accordingly in usb_serial_print() but it is the same crash. Why do print_row and print_col get away with not being null-initialized?

system/ehci.c
tormodvolden: Note the ehci changes, as per the commit message, hasn't been tested. For instance I am not sure what the DT value is meant to be, also as commented in the code. The point is that for bulk transfers there is only one packet going out (with ACK from device), whereas for control transfers there is the setup and in/out stages. The rest is copied (cargo-culted) from the existing get_data_request().

system/usbhcd.c
martinwhitaker: Because they contain integer values, not pointer values. The relocation operation needs to be performed on the pointer value stored in the variable, not on the address of the variable.

tormodvolden:
> I strongly disagree

What exactly do you strongly disagree on? I didn't comment on the hardware's reliability or adoption.

> The (in)famous FTDI drivers that bricked many fake cables was a side effect of using the same faked ID.

No, it was not a side-effect but deliberate damage by the company against people's hardware. This is intolerable and not to be easily forgiven or forgotten. I am glad if I refreshed the awareness on this. They should take their battles to court, and not fight them on innocent people's property.

> Supporting FTDI seems mandatory to me.

I am not telling you to not support the hardware, but I am not spending my time encouraging people to buy products from them.

system/usbhcd.c
```diff
@@ -278,6 +300,98 @@ static bool configure_keyboard(const usb_hcd_t *hcd, const usb_ep_t *ep0, int in
...
+    const uint32_t baud_le = 921600; // really 923076 on CP2104, CP2105, CP2110
+    build_setup_packet(&setup_pkt, USB_REQ_TO_INTERFACE | USB_REQ_VENDOR, CP210X_SET_BAUDRATE, 0, interface_num, 4);
+    if (!hcd->methods->out_data_request(hcd, ep0, &setup_pkt, &baud_le, 4)) {
+        return false;
+    }
+
+    return true;
+}
+
+#define pl2303_vendor_read(value) do { \
```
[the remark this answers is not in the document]
tormodvolden: See my comment in https://github.com/memtest86plus/memtest86plus/pull/387#discussion_r1527166694

x86fr:
The initial brick of counterfeit FT232 devices was done by unaligned 32bit write to the EEPROM interface, leading to the upper 16bit (corresponding to the PID) being write as 0s on fake chip. Back in these days (2014), they also reprogrammed the PID of some of their own device to support alternate mode. No matter if the original issue was really intentional or not, it's true that they assumed it and even added more counter-measure a couple years after by sending garbage in TX line on fake chips.

There was an endless debate at the time about whatever if you "innocently" buy a stolen car, it's normal to have it seized and destroyed or not. I won't discuss this as I agree that the end user should not be "punished" for purchasing fake FTDIs. Likewise, I don't think Memtest86+ users should be punished for purchasing real FTDI chips. I added support for countless devices based on questionable marketing or technical strategies. I don't spend my time supporting hardware vendors or not, I spend my time supporting users.

tormodvolden:
You're sounding like if there is a discussion about whether Memtest86+ should support the FTDI (fake and real :) ) or not. I was simply explaining that I am providing a proof of concept for the excellent CP2104 devices, PL2303HX and CP340. Of course I don't mind you spending your time on providing support for other chipsets if you like.

debrouxl:
Eh, it's alright. Due to their popularity, we certainly all agree that functionally, for maximizing the usefulness of USB serial output functionality to end users, memtest86+ needs some level of support for FTDI 232 chips, at the very least a reasonable subset of the official ones.
However, especially in the non-commercial, cooperative setting that we're in for this GPL'ed software, we have some freedom to choose what we spend our free time and money on :)

system/usbhcd.c
tormodvolden: I pushed a commit where I had initialized with a dummy (but valid) non-null pointer. With this I don't see the same crash, but it gets stuck anyway. When I look at it in gdb, it looks like the assignments of print_ep and print_hcd (e.g. in find_attached_usb_keyboard) don't take effect, but it is maybe gdb that looks in the wrong place.

debrouxl:
After resolving the trivial rebase conflict by myself, I've started taking this for a test drive, on an Asus N53SV (Sandy Bridge) laptop; the USB serial adapters are plugged into one of the three USB <= 2 (black) ports, there's also an USB 3 (blue) port.
Test results:
* the CH341 and CP210x adapters are detected and initialized, but no initialization message is printed for the PL2303 adapter, though it's listed in the USB debug output;
* so far, I haven't been able to get any form of console output with either recognized and initialized adapter, though. Not even by forcing `enable_tty` and `tty_usb` to `true` in `app/config.c` and leaving default `console=` parameters. The receiving terminal is standard `screen /dev/ttyUSB0 115200` connected to a FTDI TTL-232R-3V3 cable; the TX -> RX and RX -> TX cross cable setup is correct, since from Linux, two `screen /dev/ttyUSB0 115200` print on the other side the characters entered on one side.

The USB descriptors of my cables are as follows:
```
Bus 002 Device 005: ID 1a86:7523 QinHeng Electronics CH340 serial converter
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass          255 Vendor Specific Class
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0         8
  idVendor           0x1a86 QinHeng Electronics
  idProduct          0x7523 CH340 serial converter
  bcdDevice            2.62
  iManufacturer           0 
  iProduct                2 USB2.0-Serial
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0027
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower               98mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           3
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass      1 [unknown]
      bInterfaceProtocol      2 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0020  1x 32 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x02  EP 2 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0020  1x 32 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval               1
Device Status:     0x0000
  (Bus Powered)

Bus 002 Device 006: ID 067b:2303 Prolific Technology, Inc. PL2303 Serial Port / Mobile Phone Data Cable
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x067b Prolific Technology, Inc.
  idProduct          0x2303 PL2303 Serial Port / Mobile Phone Data Cable
  bcdDevice            3.00
  iManufacturer           1 Prolific Technology Inc.
  iProduct                2 USB-Serial Controller
  iSerial                 0 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0027
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           3
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x000a  1x 10 bytes
        bInterval               1
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x02  EP 2 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x83  EP 3 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
Device Status:     0x0000
  (Bus Powered)

Bus 002 Device 007: ID 10c4:ea60 Silicon Labs CP210x UART Bridge
Negotiated speed: Full Speed (12Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               1.10
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x10c4 Silicon Labs
  idProduct          0xea60 CP210x UART Bridge
  bcdDevice            1.00
  iManufacturer           1 Silicon Labs
  iProduct                2 CP2102 USB to UART Bridge Controller
  iSerial                 3 0001
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0020
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              100mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              2 CP2102 USB to UART Bridge Controller
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
Device Status:     0x0000
  (Bus Powered)
``` 

BTW: the BIOS emulates the legacy keyboard, so I have to pass `keyboard=usb` to memtest86+ in order to perform USB initialization and detection.

tormodvolden:
Note that all the included drivers configure ~921600 baud so no need to try other speeds. The PL2303 driver supports the HX chip variant, but not the H variant, so please look at the chip marking to see if it is one of those (they all share the same VID:PID so more code would be needed to identify the less common variants).

debrouxl:
Well... while the markings for the CH340G, CP2102 and CH9102X chips are easily readable, I can't read the marking on the PL2303 chip, under the same bright light source. Sure, the plastic sheet and the bits of dust trapped underneath probably get in the way, but if there are any markings on that chip, they're very low-contrast.

tormodvolden:
The Linux driver should also debug-print the variant name. Anyway, I am surprised none of your dongles work. Does an external keyboard work well with memtest on this laptop?

debrouxl:
Yes, the external USB keyboard works with both memtest86+ and Linux. I shot that particular laptop's internal keyboard years ago, by involuntarily applying a powerful clothing-induced ESD across the left Ctrl key with my finger. Since then, most of the keys have required considerable mechanical pressure to sometimes produce a keypress which can be recorded by the OS...

tormodvolden:
I refreshed the patch set since there had been a few conflicting changes on main since last time. There is nothing new, the issue with corrupted pointers (likely due to the relocation) remains. I have tested it again with real CP2104 and PL2303 dongles accessed by memtest86+ running in QEMU. Thanks again to @koenigr for the debug_memtest.sh script which makes QEMU/gdb testing so easy!

tormodvolden:
With the new commit it runs reliably (on XHCI). Screenshot (copied from "screen") after completing 2 passes:
```
      Memtest86+ v8.00      | QEMU Virtual CPU version 2.5+                     
CLK/Temp: 2810MH227/227*C   | Pass  7% ##                                       
L1 Cache:   64KB            | Test 22% ########                                 
L2 Cache:  512KB            | Test #5  [Moving inversions, random pattern]      
L3 Cache:   16MB            | Testing: 4MB - 127MB [123MB of 122MB]             
Memory  :  122MB            | Pattern: 0xc797676ae0dc0b01                       
--------------------------------------------------------------------------------
CPU: 1 Cores 1 Threads    SMP: Disabled   | Time:  0:17:08  Status: Pass     |  
Using: All Cores                          | Pass:  2        Errors: 0           
--------------------------------------------------------------------------------
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
 <ESC> Exit  <F1> Configuration  <Space> Scroll Lock            8.00.82bec.x64  
```

tormodvolden:
More good news, I fixed the EHCI control/bulk out code. Now it works on EHCI also. Here is a screenshot from testing on real computers connected via serial, where the PL2303 dongle is detected on the EHCI controller:
```
      Memtest86+ v8.00      | Intel(R) Core(TM) i7-4785T CPU @ 2.20GHz          
CLK/Temp: 2194MHz   62/62*C | Pass 17% ######                                   
L1 Cache:   32KB   124 GB/s | Test 12% ####                                     
L2 Cache:  256KB  22.3 GB/s | Test #6  [Moving inversions, 64 bit pattern]      
L3 Cache:    8MB  24.3 GB/s | Testing: 4MB - 1GB [1020MB of 15.9GB]             
Memory  : 15.9GB  13.4 GB/s | Pattern: 0x0000000000000001                       
--------------------------------------------------------------------------------
CPU: 4 Cores 8 Threads    SMP: 8T (PAR)   | Time:  0:02:42  Status: Testing  /  
IMC: 797MHz (DDR3-1595) CAS 11-11-11-28   | Pass:  0        Errors: 0           
--------------------------------------------------------------------------------
                                                                                
Memory SPD Information                                                          
----------------------                                                          
 - Slot 0: 8GB DDR3-1600 - Patriot PSD38G1600L2S (2023-W39)                     
 - Slot 2: 8GB DDR3-1600 - Patriot PSD38G1600L2S (2023-W39)                     
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
                                                                                
 <ESC> Exit  <F1> Configuration  <Space> Scroll Lock            8.00.3d0e4.x64  
```

 @debrouxl If you get to retest this, please check in the usbdebug output what kind of controller that the dongles are detected on. If it is on XHCI or EHCI it might work, if it is on UHCI or OHCI it won't.

debrouxl:
Alright, good.

I did a single test attempt with a CH340 adapter plugged into a USB 3 port of my DL560 Gen9, passing just `console=usb` to memtest86+ - no `usbdebug`. The adapter was correctly wired, in that on Linux, between two `screen /dev/ttyUSB0 115200` sessions running against that CH340 adapter and a cable based on FTDI-232-R, I could type keys and read the appropriate ones in the other session. However, on memtest86+ with this patch applied, there was no output in the `screen` session on the other computer.

Next attempt on Friday or Saturday at the earliest. I'll try harder: XHCI+EHCI, and another adapter model :)

tormodvolden:
Just in case someone can take a look (@martinwhitaker ?) I also shared the WIP UHCI and OHCI code. I have been testing UHCI in QEMU and it seems to almost work, simply it hangs in wait_for_uhci_done() because the controller never signals the completed transfer. Maybe just some parameters or flags are wrong. The OHCI code is untested.

tormodvolden:
All the drivers are hard-coded to 921600-ish baud, so please use this baud rate also for sanity checking.

tormodvolden:
One challenge with UHCI is the explicit management of the frame list. After reserving some frame list entries for control/bulk, it now sends out data on the serial line. There are timeouts and transfer errors and lost data though (also increasingly over time).

If anyone wants to test UHCI they should preferably also add this timeout hack, otherwise it all freezes instead:
```
diff --git a/system/uhci.c b/system/uhci.c
index 1a189a0..09f593b 100644
--- a/system/uhci.c
+++ b/system/uhci.c
@@ -313,11 +313,17 @@ static bool wait_for_uhci_done(workspace_t *ws)
 {
     // Rely on the controller to timeout if the device doesn't respond.
     uint16_t status = 0;
-    while (true) {
+    uint32_t retries = 10000; // 100 ms
+    while (--retries != 0) {
         status = get_uhci_done(ws);
         if (status != 0) break;
         usleep(10);
     }
+    if (status == 0) {
+       static int timed_out_cnt = 0;
+       if (++timed_out_cnt % 100 == 0)
+           print_usb_info("timed out 100x");
+    }
 
     return ~status & UHCI_USBSTS_ERR;
 }
```

tormodvolden:
> Not that it's the cause of your current issue, since you're using XHCI and EHCI controllers, but the hcd_methods_t instances in ohci.c and uhci.c need to be updated as well, for the aforementioned relocatability reasons.

Reminder about this, now that I have attacked UHCI.

debrouxl:
I'm testing on the DL380 Gen9 this time, where the same CH340G and FTDI-232-R cables are connected as before, and typing text into two `screen /dev/ttyUSB0 921600` sessions make it appear on the other side.
`usbdebug` shows the CH341: mt86plus_387_DL380Gen9_CH341_20251116
However, `console=usb` still doesn't cause output to be printed into the `screen` session on the other side. More tests tomorrow, with other adapters.

debrouxl:
The CP2102 and PL2303 adapters are plugged in such a way that the `screen` session on the receiver side receives characters from another screen session at 921600 baud, and are both detected by memtest86+:
mt86plus_387_DL380Gen9_PL2303_20251117
mt86plus_387_DL380Gen9_CP2102_20251117
However, while memtest86+ is running, nothing gets printed to the screen session.
I tried to add `usbinit=` to the mix of CLI options: no change with `usbinit=1` (`USB_2_STEP_INIT`), and both `usbinit=2` and `usbinit=3` (`USB_EXTRA_RESET`) trigger a hang (for several dozen seconds, at least) during the UHCI probing.

The CH9102X adapter isn't detected, but that's expected.

I don't have a logic analyzer with protocol analysis support. Do you have direct or indirect experience with the Sipeed SLogic16U3, which I saw mentioned this week at https://www.cnx-software.com/2025/11/12/69-sipeed-slogic16u3-low-cost-logic-analyzer-supports-3-2-gbps-bandwidth-150-protocols/ , and its usage from Linux (I see a mention of sigrok, but still) ? Not that I'm keen to buy it, but at least, it's significantly more capable and cheaper than the competitors I had seen before, which are mentioned in the comparison.

tormodvolden:
Thanks, I haven't seen this logic analyzer. If it works fine with Sigrok, I know what I want for Christmas.

Some USB-serial dongles have LEDs that light up for RX and TX activity. Does any of yours have this, so that you can confirm there is nothing sent out? My CH340 dongle have such LEDs.

Do you have different USB ports on this machine that you can try out?

When the "serial adapter found" messages appear, it means at least the control out transfers were reported as successful, but there could still be issues for bulk out transfers. XHCI works fine for me on several machines though.

Can you try without any keyboards plugged in? I guess you would need autostart, or adding a startup delay so that you can launch mt86plus but unplug the keyboard before USB is initialized...

tormodvolden:
It can also be worthwhile to try your USB serial adapters with QEMU.
1. Blacklist the corresponding OS driver with e.g. an /etc/modprobe.d/usb_serial.conf file with contents `blacklist cp210x` . Unplug and `modprobe -r cp210x`, replug. Of course, do not blacklist the other adapter's driver, that you will use with `screen`, if you run it on the same machine.
3. Find bus/port with lsusb, chgrp $USER /dev/bus/usb/xxx/yyy (or use udev rules to grant access)
4. Add QEMU flags to debug_memtest.sh to pass the USB device on to the emulated machine, adapting the IDs accordingly:
```
QEMU_FLAGS+=" -device qemu-xhci"
QEMU_FLAGS+=" -device usb-host,vendorid=0x10c4,productid=0xea60,pcap=cp210x.pcap"
```
The pcap option creates a packet dump that you can analyze with Wireshark.
Or for "plugging" the device into a virtual UHCI controller:
```
QEMU_FLAGS+=" -usb"
QEMU_FLAGS+=" -device usb-host,vendorid=0x10c4,productid=0xea60,pcap=cp210x.pcap,bus=usb-bus.0"
```
You may want to pass a USB keyboard through as well:
```
QEMU_FLAGS+=" -device usb-kbd"
```
5. Run ./debug_memtest.sh following the instructions in [doc/HOW_TO_DEBUG_WITH_GDB.md](https://github.com/memtest86plus/memtest86plus/blob/main/doc/HOW_TO_DEBUG_WITH_GDB.md)

debrouxl:
> Thanks, I haven't seen this logic analyzer. If it works fine with Sigrok, I know what I want for Christmas.

Happy to pass useful information around, then :)
The end of the CNX Software article mentions that Sipeed has started work on a higher-end, more expensive model, but the current one is already quite an improvement over other models.

> Some USB-serial dongles have LEDs that light up for RX and TX activity. Does any of yours have this, so that you can confirm there is nothing sent out? My CH340 dongle have such LEDs.

AFAICS, all 4 dongles have Power + TX + RX LEDs, and I had totally forgotten about the latter two...

> Do you have different USB ports on this machine that you can try out?

Yup, although all three externally accessible ports are blue USB 3 A female connectors, and plugged to the same XHCI controller.

> When the "serial adapter found" messages appear, it means at least the control out transfers were reported as successful, but there could still be issues for bulk out transfers. XHCI works fine for me on several machines though.

Sure.

> Can you try without any keyboards plugged in? I guess you would need autostart, or adding a startup delay so that you can launch mt86plus but unplug the keyboard before USB is initialized...

Will do at the end of the upcoming week. No USB keyboard means that I won't be able to use iLO in order to take perfect quality  screenshots, but you may be onto something, considering that merely moving the mouse in the HTML 5 iLO console is enough to bypass memtest86+'s wait for a keypress at the end of the usbdebug screen, which I'd tend to consider spurious behaviour.
Besides, the usage of a physical screen will make it easier for me to move around the server and view TX / RX LED activity.

debrouxl:
I got partial success on a Skylake generation laptop, in that with all three adapter families handled by this PR, connected to the sole XHCI controller in this computer (usbdebug indicates two USB keyboards, which is incorrect, as I didn't plug an external USB keyboard), proper output was produced in the `screen` session, and the connection with the PL2303 remained stable for about half an hour, more than one pass:
mt86plus_387_GL553VW_PL2303_20251117
before I powered the computer down to switch to the CH340:
mt86plus_387_GL553VW_CH340_20251117
then the CP2102:
mt86plus_387_GL553VW_CP2102_20251117

However:
* sending ESC, F1, 1 or Space didn't make the TX LEDs blink and trigger a reboot / the opening of the configuration menu / the toggle of the scroll lock feature. ESC and F1 are special keypresses and may have been caught by screen and/or the Konsole & xterm sessions screen was running in, but 1 and Space wouldn't have been caught;
* on the CP2102, I had to press Enter in the initial memtest86+ screen for the test to begin, or maybe I happened to press Enter right around the time the test would have started on its own anyway. The timing base is wacky, as evidenced by the ludicrous CPU speed, insane memory bandwidths, and the fact that the bit fade test was reached in 00:00:48 according to memtest86+, which still sent terminal updates every "second" - more like every ~23 wall clock seconds, which is consistent with a ~2600 MHz CPU being reported as 60100 MHz (2600 x 23 = 59800). At this rate, the bit fade test, which waits 80 seconds before reading memory back, is going to take 80 x ~23 x 2 wall clock seconds, nearly an hour.

tormodvolden:
Great, that is progress! Yes, there is no support for reading data from the serial line. It is a one-way stream of screen data only.

The slow performance with CP2102 might indicate some timeouts are happening. Note that everything on the software side runs synchronously, so the print routine waits for confirmation by the USB controller that the send request has been submitted, before returning and letting the memory check continue. Apart from fixing any failing acknowledgments and timeouts, buffering of print output to use fewer, but larger, transfers would speed up things. I am not sure how this affects the displayed CPU frequency though, since I'd expect this to be established during initialization. Another cause of problems could be that we are still accessing some stale pointers, I have some ideas and WIP about that, in particular we need to keep a static list of endpoint properties.

debrouxl:
* unlike the DL380 Gen9, the DL560 Gen9 has 2 black USB 2 ports; when the CH340 adapter is plugged into one of these, it's shown in the `usbdebug` screen, but the TX led doesn't blink, and data isn't printed onto the `screen` session on the other side. This server is now testing RAM for real, so next tests tomorrow at best;
* on a HP Z210 SFF with a PS/2 keyboard and no other USB devices, the `usbdebug` screen doesn't wait for a keypress (I don't see it at all, in fact) and no data is printed onto the `screen` session, whichever the adapter among CH340, PL2303 and CP2102, and whichever the USB port: I've tried most of the 6 ports on the rear, and some of the 4 ports on the front as well.

tormodvolden:
> * unlike the DL380 Gen9, the DL560 Gen9 has 2 black USB 2 ports; when the CH340 adapter is plugged into one of these, it's shown in the `usbdebug` screen, but the TX led doesn't blink, and data isn't printed onto the `screen` session on the other side. This server is now testing RAM for real, so next tests tomorrow at best;

Do you remember under which type of *HCI controller it was listed? OHCI will probably not work.

> * on a HP Z210 SFF with a PS/2 keyboard and no other USB devices, the `usbdebug` screen doesn't wait for a keypress (I don't see it at all, in fact) and no data is printed onto the `screen` session, whichever the adapter among CH340, PL2303 and CP2102, and whichever the USB port: I've tried most of the 6 ports on the rear, and some of the 4 ports on the front as well.

There is some application logic so that if you specify "keyboard=usb" it will pause 10 seconds if no USB keyboards were found. "usbdebug" should always make it wait for a keypress, but maybe the code somehow detects a phantom key press. Maybe also test with main branch or last release to check. This shouldn't be affected by the usb-serial code.

debrouxl:
FWIW:
1) I've attempted to rebase these changes to current main. It was almost pain-free up until the USB MSD support and AArch64 support commits; when rebasing, as I don't have much experience with USB stuff, I might have broken the code further... The result is the `usbserial` branch in this repo.
2) I pointed the `usbserial` branch to a no-fee variant of one of the mid-range, open weight LLMs. In plan mode, it reviewed the code, pointed what it considers as issues and detailed explanations for them, and produced the detailed, bite-sized implementation plan I requested. In build mode, it followed the plan and produced a couple dozen commits containing what it considers as fixes. The result is the `usbserial_with_potential_fixes` branch.

Both are untested for now.

x86fr:
Quite cool! USB code is by far the most complex part, as it requires a lot of quirks everywhere. What LLM did you use for this? From my experience working with them, you should drop OHCI and UHCI support and only keep EHCI/XHCI. The codebases that LLM have been trained on for very old controllers are too limited, even at the Fable/Sol level, and they tend to bring EHCI-related concepts back into older controllers. The key is really to test everything extensively on real hardware. I can send you a bunch of USB-to-serial adapters if you have time to test them.

debrouxl:
* here, I used a free hosted DS V4 Flash with context reduced to 256K, which proved to be enough for this review + planning + fixing task. The hosting provider is using user data for feedback and training... but the task of reviewing and suggesting fixes for publicly available FLOSS code, with additional input from this public PR, didn't qualify as remotely confidential, even before I posted about the results here ^^
* implementing USB serial output for XHCI and EHCI is indeed higher priority than implementing it for UHCI and OHCI. However, the one computer I can carry around and most easily reboot for testing purposes, an old Core 2-based laptop I found in an electronics equipment trash bin years ago, has mostly UHCI controllers. I'm not even sure how to reach its EHCI controllers, since under Linux, when I plug the USB serial adapters into any of the three accessible USB ports, `lsusb -t` indicates that the adapters end up under an USB 1.1 root hub handled by the uhci_hcd driver.
* well, having more hardware is always interesting. As I already have the 5 USB serial adapters listed above, covering the main USB serial adapter chip model families - CP2102, (CH9102X which is different), CH340, PL2303, plus a 15+-year-old FTDI-232R-3V3 as reference input cable - I'd say that I'm not in immediate need for more USB serial adapters.
The class of USB-related hardware I still don't have around, and which would be useful for my other projects as well when fine USB analysis is necessary, is one of the USB analyzer | generic logic analyzer models supported by Sigrok, with several corresponding adapters if needed (e.g. mini-A male to A female). I know that it's not very expensive on e.g. AliExpress.

Issue #172, opened 2022-09-19
Add PC Speaker beeps for errors and pass


psycodeoday:
It could be great to implement PC-Speaker simple sounds.
.
Natural short tone when test starts.
Low heavy tone when errors detected
Happy tones when test is passed.

Sometimes whole test can take around 1 hour, and without any sounds we have to watching the screen from time to time.
With Speaker sounds implemented, I would be focus on something else, and only wait for sound.
And those who don't like sounds, would have posibility to turn it off in settings.

Ver 5.01 /5.31 has a speaker sound on the beginning of the test.
Now 6.0x doesn't have sounds at all.
Is any important reason for that decision?

x86fr:
Speaker is deprecated on modern PC architecture. But this can be re-added for older computers.

psycodeoday:
> Done. Please keep it as simple as possible and put it as an option (default disabled)
Does that kind of instruction (to pc speaker) can create any issues on modern platform where there is no pc speaker? And that is why you you restricted it, to be disabled as default?

(Most of the people doesn't have pc speaker connected to motheboard in these days, and someone who doesn't like pc speaker sounds at all,  removed it before anyway - so why disables as default?)

x86fr:
Beeps are extremely annoying for a lot a testers. Memtest86+ 5.31 disables all beeps by default :

```
/* BEEP_MODE - Beep on error. Default off, Change to 1 to enable */
#define BEEP_MODE 0

/* BEEP_END_NO_ERROR - Beep at end of each pass without error. Default off, Change to 1 to enable */
#define BEEP_END_NO_ERROR 0
```
The starting beep (only one that's enabled by default) should have been disabled by BEEP_MODE.

And yes, this requires careful testing on modern platform: using the PC Speaking on UEFI is **MUCH** more complex than just sending some bytes with oub() like on old BIOS/CSM.

01e3:
Sure, how about the following plan:
 - Disabled by default, controlled via a "#define" allowing custom binary to have it enabled
 - Controlled by a boot parameter, for example beep=on/off
 - Perhaps: using different code depending if booted via BIOS or UEFI

Any preference what to do with Boot Sector /  MBR loaders, where specifying boot parameters is not possible? Does it make sense to add a menu option?

BTW: if we are worried about PC Speaker not working in UEFI, should we be also concerned about TSC calibration, which uses PIT Channel 2, which is exactly how PC Speaker is controlled?

[...]

PR #404, opened 2024-05-15
[Issue #172] add beep sound

demiurg-spb:
Added a sound alarm when the test is completed: 2 different tones for success and failure.

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -27,6 +27,8 @@
 
 #include "tests.h"
 
+#include "sound.h"
+
 #include "display.h"
 
 //------------------------------------------------------------------------------
@@ -393,6 +395,8 @@ void display_big_status(bool pass)
         return;
     }
 
+    sound_beep(pass);
+
     save_screen_region(POP_STATUS_REGION, popup_status_save_buffer);
 
     set_background_colour(BLACK);
@@ -572,6 +576,7 @@ void do_tick(int my_cpu)
     if (update_spinner) {
         spin_idx = (spin_idx + 1) % NUM_SPIN_STATES;
         display_spinner(spin_state[spin_idx]);
+        sound_tick_task();
     }
 
     // This only tick one time per second
--- a/app/sound.c
+++ b/app/sound.c
@@ -1,0 +1,43 @@
+#include <stdbool.h>
+#include "pit.h"
+#include "sound.h"  // self to check prototypes
+
+static bool enabled;
+static unsigned duration;
+static const unsigned DURATION = 7;
+
+static void beep_off()
+{
+	if (enabled)
+	{
+		enabled = 0;
+		duration = 0;
+		pit_off();
+	}
+}
+
+static void beep_on(unsigned f_hz)
+{
+	if (!enabled)
+	{
+		enabled = 1;
+		duration = DURATION;
+		pit_init_square_wave_generator(2, f_hz);  // pit-ch2 is connected to buzzer
+	}
+}
+
+void sound_beep(bool ok)
+{
+	static const unsigned beep_freq_ok = 1100;
+	static const unsigned beep_freq_er = 380;
+
+	beep_on(ok?  beep_freq_ok : beep_freq_er);
+}
+
+void sound_tick_task(void)
+{
+	if (duration)
+		duration--;
+	else
+		beep_off();
+}
--- a/app/sound.h
+++ b/app/sound.h
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: GPL-2.0
+#ifndef SOUND_H
+#define SOUND_H
+/**
+ * \file
+ *
+ * Provides types and variables used when performing the memory tests.
+ *
+ *//*
+ * Copyright (C) 2024 Anton Ivanov (aka demiurg_spb+rigler).
+ */
+
+#include <stdbool.h>
+
+void sound_beep(bool ok);
+void sound_tick_task(void);
+
+
+#endif // SOUN_H
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -76,6 +76,7 @@ TST_OBJS = tests/addr_walk1.o \
 
 APP_OBJS = app/badram.o \
            app/config.o \
+           app/sound.o \
            app/display.o \
            app/error.o \
            app/interrupt.o \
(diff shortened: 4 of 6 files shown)


demiurg-spb:
I just proposed a solution to this problem: pull request #404 ( [Issue #172] add beep sound)

debrouxl:
Thanks for this PR.
As requested by Sam in #172 , please make the beeping configurable and disabled by default :)
You'll need to modify app/config.* for this.

demiurg-spb:
It seems to me that this should be divided into two patches (functionality and configuration menu). In my opinion, the sound signal should be turned on by default, as is done in the commercial project memtest86pro from Passmark. The main task of this tool is to quickly check memory. The signal contributes to this, because You can quickly change the module to the next one in the test bench. If this signal bothers some boy who is afraid to wake up his parents at night... Then this is not the target of this tool and it can either turn off the speaker, or use the previous version of the program, or offer a patch itself. So please just accept this patch as is. We'll deal with the menu later. Either me or someone else. There is no need to drag out this decision for years. Thanks for understanding.

x86fr:
As already explained, this will not be merged as-is, especially enabled by default.

PC speaker using PIT Timer is deprecated since a very long time. Vast majority of modern computers since a decade doesn't have a speaker. Intel deprecated PIT Timer in Core 10th Gen, disabled it by default starting with 12th Gen and they will probably remove it completely starting with Arrow Lake. Accessing a disabled PIT Timer has been proven to cause hangs and freeze on some UEFI platforms (especially Mobile). 

When this feature was used in the 00s, the way you added it was only good for small business or power user, only testing one PC at time. On factories testing many PCs in a row, having them beep all at once was confusing. They only wanted a "beep on error" features, hence the need for additional configuration.

Anyway, as it's a trivial feature to add when using the legacy PIT timer, it's perfectly fine to have forks tailored for your specific need. Maybe we can add this in the "legacy" Memtest86+ when we will split the code base for pre- and post-Haswell (unrelated, but I agree with @debrouxl that Haswell could be a good boundary).

demiurg-spb:
Fine. Wait indefinitely for manna from heaven... Instead of ALREADY getting the chip people need. If you don't like it, let them turn off the beeper yourself. 99.9% of users need the sound in the format in which I suggested it. 
Do you think the people who make the commercial product memtest86pro are idiots? Good luck...

x86fr:
Ok. Time to stop here.

debrouxl:
Well, we maintainers would be complete idiots to *enable by default* an *auxiliary* function which can demonstrably cause broken timer base on user computers ( #40 ) and freezes or complete hangs of said computers, in a way which might not even be easy to programmatically detect reliably.
Put another way: potentially preventing users (and a growing number thereof, as time passes by and platforms without a working PIT become more numerous) from being able to enjoy the *essential* function of our memory testing program (namely testing memory), wasting both user time and maintainer time (since at least some users will spend time reporting bugs caused by that broken feature), just because we enable a broken *auxiliary* function by default, does little good :)

Being an exclusively UEFI application, Passmark's finest can use the UEFI methods to access the speaker, which probably shields them from a number of issues, assuming the BIOS implementations and the platform hardware aren't buggy either... However, in the process, by not exiting the UEFI boot services, unlike memtest86+, Passmark's finest can't test the megabytes (to put it lightly) of memory permanently reserved by the UEFI boot services, which isn't ideal either.

Since beep sound production is a recurring user feature request, and one where users seem to have *particularly strong* opinions, we'll probably save maintainer and user time by integrating a derivative of this patch, augmented by configurability - but definitely a default disabled value. It's up to users to flip the default and recompile themselves, if they so wish.

A first reasonable cutoff for x86 micro-architectures would be x86-64-v2, so Nehalem+ and Bulldozer+. But yeah, x86-64-v3 is the next step, though more distant in time, IMO.

debrouxl:
FTR, on a laptop equipped with a Core 2 Duo, the code in this PR, augmented by the patch below, produced sound at the end of the first pass.

```
diff --git a/app/config.c b/app/config.c
index a182b75..5e9a5fe 100644
--- a/app/config.c
+++ b/app/config.c
@@ -99,6 +99,7 @@ bool            enable_sm          = true;
 bool            enable_bench       = true;
 bool            enable_mch_read    = true;
 bool            enable_numa        = false;
+bool            enable_sound       = false;
 
 bool            enable_ecc_polling = false;
 
@@ -258,6 +259,8 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "high", 5) == 0) {
             power_save = POWER_SAVE_HIGH;
         }
+    } else if (strncmp(option, "sound", 6) == 0) {
+        enable_sound = true;
     } else if (strncmp(option, "trace", 6) == 0) {
         enable_trace = true;
     } else if (strncmp(option, "usbdebug", 9) == 0) {
diff --git a/app/config.h b/app/config.h
index 755b4cd..6ca77aa 100644
--- a/app/config.h
+++ b/app/config.h
@@ -61,6 +61,7 @@ extern bool         enable_bench;
 extern bool         enable_mch_read;
 extern bool         enable_ecc_polling;
 extern bool         enable_numa;
+extern bool         enable_sound;
 
 extern bool         pause_at_start;
 
diff --git a/app/display.c b/app/display.c
index b97f8de..f02de54 100644
--- a/app/display.c
+++ b/app/display.c
@@ -395,7 +395,9 @@ void display_big_status(bool pass)
         return;
     }
 
-    sound_beep(pass);
+    if (enable_sound) {
+        sound_beep(pass);
+    }
 
     save_screen_region(POP_STATUS_REGION, popup_status_save_buffer);
 
@@ -576,7 +578,9 @@ void do_tick(int my_cpu)
     if (update_spinner) {
         spin_idx = (spin_idx + 1) % NUM_SPIN_STATES;
         display_spinner(spin_state[spin_idx]);
-        sound_tick_task();
+        if (enable_sound) {
+            sound_tick_task();
+        }
     }
 
     // This only tick one time per second
```

At a minimum, fixing the coding style, replacing those static const constants by defines or simply hard-coding their single occurrence (we can't rely on C23 constexpr yet ^^), fixing variable types would also be in order for integrating the code in this PR.

debrouxl:
I've pushed a `sound` branch containing this code with the modifications described in the commit message. After #172 and #404 , it represents as far of a tradeoff as I'm willing to go wrt. beep sound production, if we were to integrate the commit to the main branch: both disabled by default and undocumented.

Issue #361, opened 2023-12-03
Useless comparison in system/hwquirks.c


debrouxl:
When trying to build memtest86+ with clang in order to determine the availability of strongly typed enums in C (they work with Clang 12-17 in `-std=c11` mode, FTR), an interesting warning was present in the compiler output:
```
../system/hwquirks.c:85:46: warning: overlapping comparisons always evaluate to false [-Wtautological-overlap-compare]
    if (cpuid_info.version.extendedModel < 6 && cpuid_info.version.extendedModel > 7)   // Not Rev G
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

The `&&` should certainly be a `||`.

Also, the comparison just below the aforementioned one:
```
    if (cpuid_info.version.extendedModel == 6 && cpuid_info.version.extendedModel < 9)  // Not Desktop
        return;
```
looks odd ?

PR #405, opened 2024-05-19
Clang build fixes

debrouxl:
It happens that only a couple tweaks are necessary to make it possible to *compile* and *assemble* the current memtest86+ code. Therefore, I think we should integrate such changes, so as to be able to benefit from Clang's warnings, at least one of which ( #361 ) is demonstrably useful.
Making Clang able to *link* memtest86+ requires an additional tweak, because Clang's `__builtin_memcpy()` and `__builtin_memset()` do, in this case, end up calling `memcpy()` and `memset()`, which we have been able to stop providing by default thanks to GCC's builtins, except for the GDB debugging build.

There are obvious workarounds, either by handling `__clang__` the same way as `DEBUG_GDB`, or by providing an inline version of those two functions:
```
diff --git a/lib/string.h b/lib/string.h
index fea0b3b..da1b15e 100644
--- a/lib/string.h
+++ b/lib/string.h
@@ -37,7 +37,18 @@ static inline int memcmp(const void *s1, const void *s2, size_t n)
  * void *memcpy(void *dst, const void *src, size_t n);
  */
 #ifndef DEBUG_GDB
-    #define memcpy(d, s, n) __builtin_memcpy((d), (s), (n))
+    #ifdef __clang__
+        static inline void *memcpy (void *dest, const void *src, size_t len)
+        {
+            char *d = dest;
+            const char *s = src;
+            while (len--)
+                *d++ = *s++;
+            return dest;
+        }
+    #else
+        #define memcpy(d, s, n) __builtin_memcpy((d), (s), (n))
+    #endif
 #else
     void *memcpy (void *dest, const void *src, size_t len);
 #endif
@@ -55,7 +66,17 @@ void *memmove(void *dest, const void *src, size_t n);
  * void *memset(void *s, int c, size_t n);
  */
 #ifndef DEBUG_GDB
-    #define memset(s, c, n) __builtin_memset((s), (c), (n))
+    #ifdef __clang__
+        static inline void *memset (void *dest, int val, size_t len)
+        {
+            unsigned char *ptr = dest;
+            while (len-- > 0)
+                *ptr++ = val;
+            return dest;
+        }
+    #else
+        #define memset(s, c, n) __builtin_memset((s), (c), (n))
+    #endif
 #else
     void *memset (void *dest, int val, size_t len);
 #endif
```

### DIFF
--- a/app/main.c
+++ b/app/main.c
@@ -171,7 +171,8 @@ static void run_at(uintptr_t addr, int my_cpu)
     __asm__ __volatile__("movl %0, %%edi" : : "r" (new_start_addr));
 #endif
 
-    goto *new_start_addr;
+    // Jump to new_start_addr.
+    ((void (*)(void))new_start_addr)();
 }
 
 static bool set_load_addr(uintptr_t *load_addr, size_t program_size, uintptr_t lower_limit, uintptr_t upper_limit)
--- a/boot/x86/startup32.S
+++ b/boot/x86/startup32.S
@@ -648,18 +648,37 @@ gdt_end:
 
 	.data
 
-	.macro	ptes64 start, count=64
-	.quad	\start + 0x0000000 + 0x83
-	.quad	\start + 0x0200000 + 0x83
-	.quad	\start + 0x0400000 + 0x83
-	.quad	\start + 0x0600000 + 0x83
-	.quad	\start + 0x0800000 + 0x83
-	.quad	\start + 0x0A00000 + 0x83
-	.quad	\start + 0x0C00000 + 0x83
-	.quad	\start + 0x0E00000 + 0x83
-	.if \count-1
-	ptes64	"(\start+0x01000000)",\count-1
-	.endif
+	.macro	ptes8 start, flags
+	.quad	\start + 0x00000000 + \flags
+	.quad	\start + 0x00200000 + \flags
+	.quad	\start + 0x00400000 + \flags
+	.quad	\start + 0x00600000 + \flags
+	.quad	\start + 0x00800000 + \flags
+	.quad	\start + 0x00A00000 + \flags
+	.quad	\start + 0x00C00000 + \flags
+	.quad	\start + 0x00E00000 + \flags
+	.endm
+
+	.macro	ptes64 start, flags
+	ptes8	\start + 0x00000000, \flags
+	ptes8	\start + 0x01000000, \flags
+	ptes8	\start + 0x02000000, \flags
+	ptes8	\start + 0x03000000, \flags
+	ptes8	\start + 0x04000000, \flags
+	ptes8	\start + 0x05000000, \flags
+	ptes8	\start + 0x06000000, \flags
+	ptes8	\start + 0x07000000, \flags
+	.endm
+
+	.macro	ptes512 start, flags
+	ptes64	\start + 0x00000000, \flags
+	ptes64	\start + 0x08000000, \flags
+	ptes64	\start + 0x10000000, \flags
+	ptes64	\start + 0x18000000, \flags
+	ptes64	\start + 0x20000000, \flags
+	ptes64	\start + 0x28000000, \flags
+	ptes64	\start + 0x30000000, \flags
+	ptes64	\start + 0x38000000, \flags
 	.endm
 
 	.macro	maxdepth depth=1
@@ -702,22 +721,22 @@ pdp:
 	.align	4096
 	.globl	pd0
 pd0:
-	ptes64	0x0000000000000000
+	ptes512	0x0000000000000000, 0x0000000000000083
 
 	.align	4096
 	.globl	pd1
 pd1:
-	ptes64	0x0000000040000000
+	ptes512	0x0000000040000000, 0x0000000000000083
 
 	.align	4096
 	.globl	pd2
 pd2:
-	ptes64	0x0000000080000000
+	ptes512	0x0000000080000000, 0x0000000000000083
 
 	.align	4096
 	.globl	pd3
 pd3:
-	ptes64	0x00000000C0000000
+	ptes512	0x00000000C0000000, 0x0000000000000083
 
 	.previous
 
--- a/boot/x86/startup64.S
+++ b/boot/x86/startup64.S
@@ -473,18 +473,37 @@ gdt_end:
 
 	.data
 
-	.macro	ptes64 start, count=64
-	.quad	\start + 0x0000000 + 0x83
-	.quad	\start + 0x0200000 + 0x83
-	.quad	\start + 0x0400000 + 0x83
-	.quad	\start + 0x0600000 + 0x83
-	.quad	\start + 0x0800000 + 0x83
-	.quad	\start + 0x0A00000 + 0x83
-	.quad	\start + 0x0C00000 + 0x83
-	.quad	\start + 0x0E00000 + 0x83
-	.if \count-1
-	ptes64	"(\start+0x01000000)",\count-1
-	.endif
+	.macro	ptes8 start, flags
+	.quad	\start + 0x00000000 + \flags
+	.quad	\start + 0x00200000 + \flags
+	.quad	\start + 0x00400000 + \flags
+	.quad	\start + 0x00600000 + \flags
+	.quad	\start + 0x00800000 + \flags
+	.quad	\start + 0x00A00000 + \flags
+	.quad	\start + 0x00C00000 + \flags
+	.quad	\start + 0x00E00000 + \flags
+	.endm
+
+	.macro	ptes64 start, flags
+	ptes8	\start + 0x00000000, \flags
+	ptes8	\start + 0x01000000, \flags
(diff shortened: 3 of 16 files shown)


app/main.c
```diff
@@ -171,7 +171,7 @@ static void run_at(uintptr_t addr, int my_cpu)
     __asm__ __volatile__("movl %0, %%edi" : : "r" (new_start_addr));
 #endif
 
-    goto *new_start_addr;
+    ((void (*)(void))new_start_addr)();
```
martinwhitaker, line 174: This makes the intent less clear. Maybe add a comment?

martinwhitaker:
It seems bad to have both an inline and a non-inline version of memcpy and memset in the code base.

What happens if you use clang's `__builtin_memcpy_inline`?

app/main.c
debrouxl: Alright, done.

debrouxl:
I fully agree that it's bad to have both an inline and a non-inline version of these functions.
However, Clang's `__builtin_memcpy_inline()` and `__builtin_memset_inline()` (TIL that these exist) require compile-time constant size; we have three calls which use link-time or run-time size, in `system/smp.c` and `system/xhci.c`... and surprisingly, `__builtin_constant_p()` somehow isn't strong enough to yield correct behaviour:
```
diff --git a/lib/string.h b/lib/string.h
index fea0b3b..5c35b42 100644
--- a/lib/string.h
+++ b/lib/string.h
@@ -37,7 +37,11 @@ static inline int memcmp(const void *s1, const void *s2, size_t n)
  * void *memcpy(void *dst, const void *src, size_t n);
  */
 #ifndef DEBUG_GDB
-    #define memcpy(d, s, n) __builtin_memcpy((d), (s), (n))
+    #ifdef __clang__
+        #define memcpy(d, s, n) (__builtin_constant_p(n) ? __builtin_memcpy_inline((d), (s), (n)) : __builtin_memcpy((d), (s), (n)))
+    #else
+        #define memcpy(d, s, n) __builtin_memcpy((d), (s), (n))
+    #endif
 #else
     void *memcpy (void *dest, const void *src, size_t len);
 #endif
@@ -55,7 +59,11 @@ void *memmove(void *dest, const void *src, size_t n);
  * void *memset(void *s, int c, size_t n);
  */
 #ifndef DEBUG_GDB
-    #define memset(s, c, n) __builtin_memset((s), (c), (n))
+    #ifdef __clang__
+        #define memset(s, c, n) (__builtin_constant_p(n) ? __builtin_memset_inline((s), (c), (n)) : __builtin_memset((s), (c), (n)))
+    #else
+        #define memset(s, c, n) __builtin_memset((s), (c), (n))
+    #endif
 #else
     void *memset (void *dest, int val, size_t len);
 #endif
```

-> (clang 18)

```
../system/smp.c:849:5: error: argument to '__builtin_memcpy_inline' must be a constant integer
    memcpy((uint8_t *)HEAP_BASE_ADDR, ap_trampoline, ap_trampoline_size);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lib/string.h:41:60: note: expanded from macro 'memcpy'
        #define memcpy(d, s, n) (__builtin_constant_p(n) ? __builtin_memcpy_inline((d), (s), (n)) : __builtin_memcpy((d), (s), (n)))
                                                           ^                                 ~~~

../system/xhci.c:1046:5: error: argument to '__builtin_memset_inline' must be a constant integer
    memset((void *)scratchpad_vaddr, 0, scratchpad_size);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lib/string.h:63:60: note: expanded from macro 'memset'
        #define memset(s, c, n) (__builtin_constant_p(n) ? __builtin_memset_inline((s), (c), (n)) : __builtin_memset((s), (c), (n)))
                                                           ^                                 ~~~

../system/xhci.c:1062:5: error: argument to '__builtin_memset_inline' must be a constant integer
    memset((void *)device_context_index_vaddr, 0, device_context_index_size);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lib/string.h:63:60: note: expanded from macro 'memset'
        #define memset(s, c, n) (__builtin_constant_p(n) ? __builtin_memset_inline((s), (c), (n)) : __builtin_memset((s), (c), (n)))
```

debrouxl:
I was browsing through today's LLVM Weekly 551 installment, and I noticed an interesting entry in it:
```
* llvm.memcpy.inline no longer requires a constant length.
  [522fd53]( https://github.com/llvm/llvm-project/commit/522fd53838d5 ).
```
That commit references an earlier change for `llvm.memset.inline`. It looks like both changes are going to be in LLVM / Clang 19, whose release branch is going to be created soon; I'll have to see whether they unblock this PR :)

RossComputerGuy:
Just tried this PR and ran into an error:
```
memtest86+-x86_64-linux> x86_64-unknown-linux-gnu-ld -shared -Bsymbolic -T ldscripts/memtest_shared.lds -o memtest_shared boot/startup.o boot/efisetup.o system/acpi.o system/cpuid.o system/cpuinfo.o system/cpulocal.o system/ehci.o system/font.o system/hwctrl.o system/heap.o system/hwquirks.o system/keyboard.o system/ohci.o system/memctrl.o system/pci.o system/pmem.o system/reloc.o system/screen.o system/serial.o system/smbios.o system/smbus.o system/smp.o system/temperature.o system/timers.o system/uhci.o system/usbhcd.o system/vmem.o system/xhci.o system/imc/amd_zen.o system/imc/intel_adl.o system/imc/intel_hsw.o system/imc/intel_icl.o system/imc/intel_skl.o system/imc/intel_snb.o lib/barrier.o lib/print.o lib/read.o lib/string.o lib/unistd.o tests/addr_walk1.o tests/bit_fade.o tests/block_move.o tests/modulo_n.o tests/mov_inv_fixed.o tests/mov_inv_random.o tests/mov_inv_walk1.o tests/own_addr.o tests/test_helper.o tests/tests.o app/badram.o app/config.o app/display.o app/error.o app/interrupt.o app/main.o
memtest86+-x86_64-linux> x86_64-unknown-linux-gnu-ld: error: unknown argument '--warn-constructors'
memtest86+-x86_64-linux> make: *** [Makefile:160: memtest_shared] Error 1
```

debrouxl:
Yeah, variants of memset + memcpy not being found at link time, or not being compilable because of non-constant size arguments, are the remaining issue. Clang 19-rc1 did not fix it.

debrouxl:
From your message, I'm not sure whether you're working on a fork of memtest86+ or a fork of Clang. However, besides using `__builtin_memcpy_inline()` and `__builtin_memset_inline()` on Clang - when they work, that is - I'm not aware that there's anything left to change in memtest86+ ?
memtest86+ is at least the second code base in which I'm using `__builtin_memcpy()` in a freestanding environment. GCC has made it work, without making the generated code needlessly inefficient, since over a decade ago. For compatibility and usability purposes, Clang needs to to the same :)

RossComputerGuy:
> From your message, I'm not sure whether you're working on a fork of memtest86+ or a fork of Clang.

Memtest86+

debrouxl:
I've just force pushed a mix between the former state of #405 and the changes in #521 , I'll work on it again later tonight.
I'll keep this PR a draft until I've reduced the changes to a quasi no-op for gcc+ld - that is, a no-op beyond the goto -> normal function call change, which changes the contents of .text.

debrouxl:
Well, duh. When comparing final binaries, one has to make sure that the embedded git commit hash is the same, otherwise the global hashes are different, too...

This version integrates all but one changes from #521 , given that besides the goto -> function normal function call, they have zero impact on the final binary produced by gcc + ld . In its current state, this PR is much more than enough to build with Clang + ld, so I'll remove the draft status.
The change I left out for now is the last roadblock for building with lld ("working around lld's limitations", I'd say): moving `.got` next to `.rela.got` (the contributed version moves `_edata` in the process, which is likely wrong - at least, if I move `_edata` to the end of `.data`, I obtain a different final binary). That change would require wider testing, indeed.

martinwhitaker (approved):
Looks OK to me.

debrouxl:
@kilaterlee this changes the Loongarch files as well, but the changes are no-ops. I have a better reason for tagging you, though: https://github.com/memtest86plus/memtest86plus/actions/runs/18263517933 shows that in order to be able to add clang to the compiler matrix in the Github builder definitions across all architectures, we'll need a `clang` toolchain which supports LA64 :)

The other roadblock for adding clang to the compiler matrix is the i586 build failure on Ubuntu 22.04. Since there's no issue on 24.04, after memtest86+ 8.00, either we simply drop support for that old distro due to its seemingly buggy Clang toolchain, or we use a package with a fixed version of Clang.

PR #406, opened 2024-05-20
Add GPT partition support

pbatard:
People creating the media through [_File System Transposition_](https://lists.gnu.org/archive/html/grub-devel/2022-06/msg00024.html) rather than DD copy may end up with a media that uses a GPT partition table rather than MBR.

So add GPT support to GRUB, as a low cost beneficial change.

For reference, _File System Transposition_ is the default method used by Rufus and other utilities for creating bootable media from an ISO, and it allows users to select GPT instead over MBR, in which case `memtest86+` will not boot.

### DIFF
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -228,7 +228,7 @@ GRUB_LIB_DIR ?= /usr/lib/grub
 
 GRUB_MKIMAGE := $(shell command -v grub2-mkimage || command -v grub-mkimage)
 
-GRUB_MODULES = iso9660 fat part_msdos all_video font gfxterm gfxmenu \
+GRUB_MODULES = iso9660 fat part_msdos part_gpt all_video font gfxterm gfxmenu \
                boot chain configfile echo ls
 
 grub-eltorito.img:
--- a/build64/Makefile
+++ b/build64/Makefile
@@ -227,7 +227,7 @@ GRUB_LIB_DIR ?= /usr/lib/grub
 
 GRUB_MKIMAGE := $(shell command -v grub2-mkimage || command -v grub-mkimage)
 
-GRUB_MODULES = iso9660 fat part_msdos all_video font gfxterm gfxmenu \
+GRUB_MODULES = iso9660 fat part_msdos part_gpt all_video font gfxterm gfxmenu \
                boot chain configfile echo ls
 
 grub-eltorito.img:


debrouxl:
Interesting. Although I have my Windows-using colleagues exclusively use the dd mode of Rufus or similar tools, and I currently consider that memtest86+ users should do the same for the memtest86+ ISO and in fact any other ISO (unless said ISO is a broken historic relic and only the massaging performed by Rufus or friends fixes the ability to boot; for new ISOs, fixing the issue at the source may arguably be a better way)... this change is indeed low cost, `part_gpt.mod` being a bit over 3 KB on my computer :)

Sam, Martin ?

pbatard:
Thanks for considering this request.

I'll add that there's one benefit of being able to use FST with GPT, which is to ensure that the media can only ever boot in pure UEFI mode and never BIOS/Legacy, as opposed to DD copy, which usually offers both and will pick whatever the computer is set to... which may not always be what the user actually wants.

Granted, this shouldn't make much of a difference for memtest86+, but I actually ran into the issue when I wanted to make sure that I ran the EFI version and not the BIOS version, and then found that my media didn't boot... ;)

x86fr:
Never used FST, but if this can help, it seems harmless, so fine to me.

martinwhitaker:
Well, it adds another variable when users report that the ISO doesn't boot on their hardware. But I guess it solves the problem of users not reading the instructions to use the dd method.

pbatard:
> is no fun for upstreams

I hope what I'm going to say is not going to sound to antagonistic, but I suppose I need to point out that, what makes it easy for upstream, and may appear as a surefire win for maintainers in the form of ISOHybrid, [may not exactly translate in that sure a win for the end users](https://github.com/pbatard/rufus/wiki/FAQ#user-content-Why_doesnt_Rufus_recommend_DD_mode_over_ISO_mode_for_ISOHybrid_images_Surely_DD_is_better) and that, per the implicit goals of Open Source, providing **choice** to end users when it comes to media creation, even if it comes at the cost of more work for upstream Open Source maintainers, is what we should all ultimately strive for...

Or to put it more simply: Treating ISOhybrid as glorified DD images is far from the panacea that a lot of people seem to believe it is...

debrouxl:
Don't worry, I'm not feeling antagonized - you're being polite, and it takes significantly more than that, both in terms of wording and content, to ruffle my feathers. We both know that there's far worse in the FLOSS world :)

You're right, I should have written "is no fun for either users / customers or upstream". When Rufus or a similar tool breaks the bootability of an ISO, which is normally relatively infrequent, users / customers are the first to waste time; some of these users / customers will end up spending further time reporting the bugs; then maintainers will have to spend time acting on bug reports (and in the commercial world, management can be unhappy that the tight schedule is derailed). Everyone loses.

I can understand that in some circumstances, users can be better off using Rufus or similar tools. I also acknowledge that you have a much wider view than I do on that topic, and I want in no way to disparage your work on Rufus. However, I tend to still like dd mode for recent ISOs, because in my more limited experience on both sides over the past 7 years (maintainer for producing two strains of custom RHEL-based distro installer ISOs, user for using these ISOs on USB MSDs by myself on various computer models), using a hybrid ISO and weaning it from having to be massaged by Rufus has repeatedly proven to be a significant time, bore and cost saver, on both the maintainer-user and customer side.

On the maintainer-user side, being able to
(go to the appropriate place, fetch a USB MSD, transfer it to a physical Linux computer at near-GbE speeds, use `dd` without frills or fuss, and then proceed to use the USB MSD onto the nearby computers)
is much better than having to
(go to the appropriate place, fetch the USB MSD, transfer the ISO to a Windows computer across a slow network, move back to the Windows computer - which, BTW needs a periodically renewed special exemption from security restrictions for being able to 1) use Rufus and 2) leave the FAT32 filesystem on the MSD unencrypted - then wait for Rufus to produce and write a FAT32 FS out of that ISO, and then move back to the location with the computers to be installed with this USB MSD)
especially when repeated multiple times per day at times (testing on VMs doesn't cut it for all purposes), and dozens of times a year :)

On the user-customer side (colleagues or customers using the ISO on target computers), there's another major perk to using dd (mode), which has already saved us multiple times over the years: the built-in integrity checking, i.e. `implantisomd5` and the installer-side checking based on such data. Last time I checked, which was admittedly a while ago, that was broken when Rufus massaged the ISO. The integrity checking did make it possible to detect broken ISO copies / improper writes to USB MSDs... they still happen, especially when many intermediate file transfers are required for the ISO to reach the computer used to write the ISO to USB MSD.
When the data errors affect a RPM / the RPM metadata, the RPM integrity checking will pause / abort the install anyway; when a used part of the installer itself is hit, chances are that it will trigger some sort of serious misbehaviour; when the data errors hit a pre-install / post-install script, it can produce an install system broken in weird ways, which could, in extreme cases, mean an unplanned trip to a faraway country - with project schedule, administrative, commercial and legal ripple effects.

To sum up: in my day job, the Rufus massaging approach would already have failed me and my colleagues, whereas the isohybrid+dd method has been working well so far across multiple RHEL-derived distros and versions thereof, and has avoided real practical issues. However, as far as memtest86+ is concerned, I still plan to integrate your contribution, which won't harm users of the (intended) dd mode, and which will help users of the FST mode, as you described :)

martinwhitaker:
I agree with Lionel. No objection to merging this.

x86fr:
Really interesting. Thanks for this comment.

PR #410, opened 2024-06-18
Add LoongArch support

kilaterlee:
This patch set is submit the LoongArch support.
1. Added LoongArch support in lib, system, test and app
2. Added 8 and 16 bits memory operations.
3. Added MMIO methods.
4. Added PCI MMIO methods.
5. Added the mdthod to get SPD via I2C.
6. Added Loongson LoongArch CPU IMC support.


boot/macros.h
```diff
@@ -12,7 +12,7 @@
 
 #ifndef __ASSEMBLY__
 
-#ifdef __GNUC__
+#if defined(__GNUC__) && !defined(__loongarch_lp64)
```
debrouxl, line 15: Well... I took these enhanced definitions from a recent Linux kernel version, multiple releases after Loongarch support was added to the Linux kernel, so I'm surprised by this. Could you elaborate why memtest86+ shouldn't use them on LA64 ?

lib/barrier.c
```diff
@@ -45,7 +45,20 @@ void barrier_spin_wait(barrier_t *barrier)
     if (__sync_sub_and_fetch(&barrier->count, 1) != 0) {
         volatile bool *i_am_blocked = &waiting_flags[my_cpu].flag;
         while (*i_am_blocked) {
+#if defined(__x86_64) || defined(__i386__)
             __builtin_ia32_pause();
+#elif defined (__loongarch_lp64)
+            __asm__ __volatile__ (
```
debrouxl, line 51: If/when some pause-type instruction is added to a future implementation of the architecture, the code can be adjusted.

app/display.c
```diff
@@ -143,10 +144,14 @@ void display_init(void)
     clear_screen_region(ROW_FOOTER, 0, ROW_FOOTER, SCREEN_WIDTH - 1);
     prints(ROW_FOOTER, 0, " <ESC> Exit  <F1> Configuration  <Space> Scroll Lock");
     prints(ROW_FOOTER, 64, MT_VERSION "." GIT_HASH);
-#if TESTWORD_WIDTH > 32
-    prints(ROW_FOOTER, 76, ".x64");
+#ifdef CONFIG_64BIT
```
debrouxl, line 147: I think that __x86_64__ / __loongarch_lp64 / __i386__ checks would be better here, no need for CONFIG_64BIT.

lib/string.c
```diff
@@ -53,7 +53,7 @@ void *memmove(void *dest, const void *src, size_t n)
     return dest;
 }
 
-#ifdef DEBUG_GDB
+#if (defined(DEBUG_GDB) || defined(__loongarch_lp64))
```
debrouxl, line 56: Having to do this looks like a compiler deficiency, but then these don't work on Clang either ( #405 ).

system/hwquirks.c
```diff
@@ -99,6 +102,23 @@ static void amd_k8_revfg_temp(void)
     cpu_temp_offset = 21.0f;
 }
 
+static void loongson_7a00_ehci_workaround(void)
```
debrouxl, line 105: I don't expect this chipset to be available outside LoongArch computers, so this quirk could be made conditional on `__loongarch_lp64` ?

system/io.h
```diff
@@ -118,6 +119,105 @@ __OUTS(l)
     __inlc(port) : \
     __inl(port))
 
+#elif defined(__loongarch_lp64)
+
+#define LOONGSON_IO_PROT_BASE 0x0efdfc000000
+
+static __inline uint8_t __inb (int port)
```
debrouxl, line 126: These `__inb` / `__inw` / `__inl` and `__outb` / `__outw` / `__outl` are relatively similar to each other. The existing implementations are macro-based anyway, so I'm not against using macros for these new implementations.

system/memrw16.h
```diff
@@ -0,0 +1,93 @@
...
+ * Copyright (C) 2021-2022 Martin Whitaker.
+ */
+
+#include <stdint.h>
+
+/**
+ * Reads and returns the value stored in the 16-bit memory location pointed
+ * to by ptr.
+ */
+static inline uint16_t read16(const volatile uint16_t *ptr)
```
debrouxl, line 21: I think that we should take this as an opportunity to macro-ify the simple (SISD, temporal) memrw functions as well. Sam, Martin ?

system/mmio.h
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * This file is not meant to be obfuscating: it's just complicated
+ * to (a) handle it all in a way that makes gcc able to optimize it
+ * as well as possible and (b) trying to avoid writing the same thing
+ * over and over again with slight variations and possibly making a
+ * mistake somewhere.
+ */
+
+#if defined(__loongarch_lp64)
+static __inline uint8_t mmio_read8(uint8_t *addr)
```
debrouxl, line 17: I'd like macros here as well :)
Sam, Martin ?

system/reloc64.c
```diff
@@ -70,12 +72,21 @@ typedef struct
 static inline Elf64_Addr __attribute__ ((unused)) get_load_address(void)
 {
     Elf64_Addr addr;
+#if defined(__i386__) || defined(__x86_64__)
```
debrouxl, line 75: `__x86_64__` alone would suffice here, since this file is used only there.

system/pci_mmio.c
```diff
@@ -0,0 +1,127 @@
...
+                  | (dev  & 0x1f)  << 11
+                  | (func & 0x07)  << 8
+                  | (reg  & 0xff);
+    return (addr);
+}
+
+uint8_t pci_config_read8(int bus, int dev, int func, int reg)
+{
+    if (bus == 0) {
+      return mmio_read8((uint8_t *)pci_config_type0_addr(bus, dev, func, reg));
```
debrouxl, line 58: An occasional indent slip. You've globally done well, though :)

debrouxl:
AFAICT, patches 1, 34, possibly 35, and 38 could be merged ahead of time, without waiting for the integration of the full PR.
Patch 25 is also independent of the LoongArch support, but it's effectively dead code without patch 17.

boot/macros.h
kilaterlee: Sorry, I forgot to change `c11` to `gnu11` in the Makefile, this patch will be dropped in next time.

lib/barrier.c
kilaterlee: Yes, if/when the pause instruction has been added, I will adjust this code.

app/display.c
kilaterlee: OK, I will removed the `CONFIG_64BIT` next time.

lib/string.c
kilaterlee: Yes, I think it would be better to implement the string family functions ourselves? If so, we can give up the reliance on compiler builtin functions.

system/hwquirks.c
kilaterlee: But the LoongArch computers can use other chipsets, for EX: Loongson 3C6000 and later supports the LCL(Loognson coherent link, compatible PCIe4), which supports PCIe4 bridge or chipsset linking, so at this point, I think it is not unique to LoongArch.

system/io.h
kilaterlee: Thanks.

system/reloc64.c
kilaterlee: OK, I will adjust next time.

system/pci_mmio.c
kilaterlee: OK, I will add more two spaces next time.

kilaterlee:
> AFAICT, patches 1, 34, possibly 35, and 38 could be merged ahead of time, without waiting for the integration of the full PR. Patch 25 is also independent of the LoongArch support, but it's effectively dead code without patch 17.

Hi Debrouxl @debrouxl,
Thank you careful review of my patch set. I will fix the issues you point out, and I think this patch set also needs to be discussed with more people, after all, this is the first submission.

system/memrw16.h
martinwhitaker: Why do you want macros? In-line functions are just as efficient, and far more informative about parameter and return types.

system/imc/loongson/loongson_la.c
```diff
@@ -0,0 +1,131 @@
...
+void get_imc_config_loongson(void)
+{
+    uint32_t val;
+    uint16_t refc, loopc, div, div_mode, ref_clk;
+
+    imc.type  = "DDR4";
+
+    if (strstr(cpuid_info.brand_id.str, "3C") ||
+        strstr(cpuid_info.brand_id.str, "3D") ||
+        strstr(cpuid_info.brand_id.str, "3B6000")) {
```
xen0n, line 103: Consider `strncmp` to only match the prefix. Otherwise there's always the (unlikely but non-zero) chance that later Loongson processors are named according to a different scheme and a short fragment appears elsewhere, and the test will be a false-positive and cause problems.

system/memrw16.h
```diff
@@ -0,0 +1,93 @@
...
+        "movw %0, %1"
+        :
+        : "m" (*ptr),
+          "r" (val)
+        : "memory"
+    );
+#elif defined(__loongarch_lp64)
+    __asm__ __volatile__(
+        "st.h %1, %0\n"
+        "ld.h %1, %0"
```
xen0n, line 84: For the flush semantic do we want a `dbar 0b01001` (cached and uncached, all loads after the barrier must wait for all stores before the barrier to complete) instead? I'm uncertain LoongArch's weaker memory model would allow a `st + ld` sequence to behave like x86.

If this place is to be changed, be sure to sync the changes to several other locations that do similar things.

cc @heiher in case I misunderstood the `dbar` spec

system/timers.c
```diff
@@ -85,6 +91,33 @@ static void correct_tsc(void)
...
+    //
+    // Get stable count frequency
+    //
+    calc_base_freq   = __cpucfg(0x4);
+    clock_multiplier = __cpucfg(0x5) & 0xFFFF;
+    clock_divide     = (__cpucfg(0x5) >> 16) & 0xFFFF;
+
+    excepted_ticks = (((calc_base_freq * clock_multiplier) / clock_divide) * num_millisec) / millisec_div;
+
+    __asm__ __volatile__("rdtime.d %0, $zero":"=r"(current_ticks):);
```
xen0n, line 109: The `__rdtime_d` intrinsic could be more readable. But I think maybe it's that you don't want to read a node ID along with the tick counter?

xen0n (commented):
Some quick review regarding the LoongArch-specific bits. Thank you for porting and open-sourcing this!

system/memrw16.h
debrouxl: Oh, I meant macro-ifying the definitions of the static inline functions themselves :)
The differences between the 8-bit, 16-bit, 32-bit and 64-bit versions are the value type and the width of the assembly instructions. With this patch, the four system/memrw*.h files supporting 3 architectures will amount to nearly 400 lines of quasi-duplicated code and comments; should ARM or RISC-V vendors join the fun, the duplication will worsen.

system/hwquirks.c
debrouxl: ACK then.

lib/string.c
debrouxl: In x86 and x86-64 release builds, GCC's builtins usually produce good, tiny code: `rep stos[bwlq]` for `memset`, `rep movs[bwlq]` for `memcpy`. A while ago, when I tried to use these builtins to replace calls to generic, out of line functions, I'm pretty sure that the code was also better than inline versions of the current out of line implementations, and I didn't feel like adding complexity to the latter in order to yield better code, so I simply switched memtest86+ to the GCC builtins. There are now 3 (known) situations where the builtins can't be used, though... meh.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+  fst.d  $fs0, $sp, 24 * RSIZE
+  fst.d  $fs1, $sp, 25 * RSIZE
+  fst.d  $fs2, $sp, 26 * RSIZE
+  fst.d  $fs3, $sp, 27 * RSIZE
+  fst.d  $fs4, $sp, 28 * RSIZE
+  fst.d  $fs5, $sp, 29 * RSIZE
+  fst.d  $fs6, $sp, 30 * RSIZE
+  fst.d  $fs7, $sp, 31 * RSIZE
+
+  movfcsr2gr  $t3, $r0
```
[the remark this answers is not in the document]
kilaterlee: OK, I will think about it.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+  st.d        $t3, $sp, 32 * RSIZE  // Push the FCSR0 register.
+
+  //
+  // Push the fcc0-fcc7 registers.
+  //
+  movcf2gr    $t3, $fcc0
+  or          $t2, $t3, $zero
```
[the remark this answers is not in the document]
kilaterlee: Ahh, will fix that next time.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+  fld.d  $fs1, $sp, 25 * RSIZE
+  fld.d  $fs2, $sp, 26 * RSIZE
+  fld.d  $fs3, $sp, 27 * RSIZE
+  fld.d  $fs4, $sp, 28 * RSIZE
+  fld.d  $fs5, $sp, 29 * RSIZE
+  fld.d  $fs6, $sp, 30 * RSIZE
+  fld.d  $fs7, $sp, 31 * RSIZE
+
+  ld.d        $t0, $sp, 32 * RSIZE
+  movgr2fcsr  $r0, $t0             // Pop the fcsr0 register.
```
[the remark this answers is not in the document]
kilaterlee: Same as above.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+#define CSR_REG_CONTEXT_SIZE  9  * RSIZE  // CSR registers size
+
+#define NUM_INT_VEC	20
+
+	.text
+
+exception_entry:
+  csrwr   $sp, 0x30
+
+  csrrd   $sp, 0x30
```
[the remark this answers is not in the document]
kilaterlee: > It could be helpful to `#define` the CSR numbers so the usage sites are much more self-descriptive. For example here `0x30` is `SAVE0` so it's immediately obvious that we're stashing the original SP into a scratch CSR.

This code is very samplified version of EDK2 exception handler, so some definitions are not sync with EDK2 version. I think we can defined them if it is necessary.
> 
> And BTW why is it necessary to read back the SP immediately after the write above?

Because the stack layout has to do this.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+  ld.d    $s7, $sp, 30 * RSIZE
+  ld.d    $s8, $sp, 31 * RSIZE
+  ld.d    $sp, $sp, 3 * RSIZE
+
+  ertn // Returen from exception.
+
+	.globl	startup32
+
+startup32:
+    idle 0  # Nerver be here.
```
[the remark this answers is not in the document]
kilaterlee: OK

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+
+	.org	0x440
+	.globl	startup64
+startup64:
+    # Disable global intrrupts
+    li.w    $t0, (0x1 << 2)
+    csrxchg $zero, $t0, 0x0
+
+    # Init Core
+    li.w    $t0, 0xA8        # DA mode, load/store cacheable, instructions cacheable, disable intrrupts, DMWn have no effect.
```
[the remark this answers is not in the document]
kilaterlee: I think the **effect** is more intuitive. Because the DMWn and DM mode may turned on at the same time, in which case, the DMWn is not effect. so I think it is better to use **effect**.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+    csrwr   $t0, 0x0
+
+    //
+    // Be sure the legacy mode
```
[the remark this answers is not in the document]
kilaterlee: It is possible for the AP to run in legacy mode while the BSP is running in non-legacy mode, but this is rare, so when I wrote this I wanted to say: make sure legacy mode works properly too.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+Jmp:
+    invtlb  0x0, $zero, $zero
+
+    # Set the PG and DMWn
+    li.d    $t0, 0x8000000000000001
+    csrwr   $t0, 0x180       # Uncache able address window, MMIO base: 0x8000000000000000
+    li.d    $t0, 0x11        # Cache able address window, PA == VA
+    csrwr   $t0, 0x181
+
+    li.w    $t0, 0xB0        # PG mode, load/store cacheable, instructions cacheable, disable intrrupts, DMWn effected.
```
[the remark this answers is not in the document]
kilaterlee: Same as above

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+    # Init Core
+    li.w    $t0, 0xA8        # DA mode, load/store cacheable, instructions cacheable, disable intrrupts, DMWn have no effect.
+    csrwr   $t0, 0x0
+
+    //
+    // Be sure the legacy mode
+    //
+    la.pcrel $t0, Jmp
+    li.d     $t1, 0xFFFFFFFF
+    and      $t0, $t0, $t1
```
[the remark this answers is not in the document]
kilaterlee: OK

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,465 @@
...
+    # Only enable IPI intrrupt
+    li.w    $t0, 0x1FFF
+    csrxchg $zero, $t0, 0x4
+    li.w    $t0, (0x1 << 12)
+    csrxchg $t0, $t0, 0x4
+
+    # Set exception base
+    csrrd    $t0, 0x4
+    li.d     $t1, ~(0x7 << 16)
+    and      $t0, $t0, $t1
```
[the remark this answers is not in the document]
kilaterlee: OK

build64/la64/Makefile
```diff
@@ -0,0 +1,211 @@
+LOCAL_COMPILE = true
+
+ifeq ($(LOCAL_COMPILE),true)
+  CC = gcc
+  OBJCOPY = objcopy
+else
+  CC = loongarch64-linux-gnu-gcc
+  OBJCOPY = loongarch64-linux-gnu-objcopy
```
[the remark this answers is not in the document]
kilaterlee: FC 39 also uses `loongarch-linux-gnu`, are you sure other OSes might use `loongarch64-unknown-linux-gnu`?

system/cache.h
```diff
@@ -10,12 +10,23 @@
  * Copyright (C) 2020-2022 Martin Whitaker.
  */
 
+#ifdef __loongarch_lp64
+#include <larchintrin.h>
+#include "string.h"
+#define cache_op(op,addr)     \
+    __asm__ __volatile__(     \
+    "cacop %0, %1\n"          \
```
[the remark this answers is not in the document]
kilaterlee: I think uses the instruction is more intuitive, because builtin function `__cacop_d` operand does not matched to the instruction, so I don't like it.

system/cache.h
```diff
@@ -69,12 +86,47 @@ static inline void cache_on(void)
...
+    }
+    uint64_t ways = (__cpucfg(0x14) & 0xFFFF) + 1;
+    uint64_t sets = 1 << ((__cpucfg(0x14) >> 16) & 0xFF);
+    uint64_t line_size = 1 << ((__cpucfg(0x14) >> 24) & 0x7F);
+    uint64_t va, i, j;
+    uint64_t cpu_module[1];
+    va = 0;
+
+    cpu_module[0] = (uint64_t)__iocsrrd_d(0x20);
+    if (strstr((const char *)cpu_module, "A6")) {
```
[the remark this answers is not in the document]
kilaterlee: OK, we needs to check the longer string.

system/imc/loongson/loongson_la.c
kilaterlee: Let's me think about this.

system/memrw16.h
kilaterlee: I think if only ensure the write is complete, we can change it to `dbar` series.

system/timers.c
kilaterlee: I just don't want to add more code, `__rdtime_d` returned not a value but a structure, so I think it is easy to use `__rdtime_d`. BTW, this instruction is reads out the Stable Counter, not the node counter, Stable Counter is a global stable timer, anyone using this instruction can only read the same stable counter.

lib/string.c
kilaterlee: So, I suggest, this two functions can also be used by some ARCH who not complete it in GCC built-in functions. But Clang... :) .

debrouxl:
Sam, Martin: any objections to merging patches
1 (CONFIG_64BIT)
34 (addition of the LA PE machine numbers, and I'd add the RISC-V PE machine numbers while at it, since we already have the ARM ones despite not supporting this architecture yet)
35 (startup64 symbol export, we'll eventually need it for X86-S as well when this becomes a thing)
38 (i386 instead of x86_64 for a x86-specific bit)
ahead of time, BTW ?

kilaterlee:
> Sam, Martin: any objections to merging patches 1 (CONFIG_64BIT) 34 (addition of the LA PE machine numbers, and I'd add the RISC-V PE machine numbers while at it, since we already have the ARM ones despite not supporting this architecture yet) 35 (startup64 symbol export, we'll eventually need it for X86-S as well when this becomes a thing) 38 (i386 instead of x86_64 for a x86-specific bit) ahead of time, BTW ?

Debrouxl,
This version is not yet complete, I plan to submit the next version in a few days and address all the comments. I agree with your plan, let's merge some patches that look good first.

kilaterlee:
> I would prefer to add the `CONFIG_64BIT` definition as a compiler flag in the makefile, to avoid including boot.h in files that don't need it.
> 
> I would have named it `ARCH_64BIT`, but won't object if others prefer `CONFIG_64BIT`.
> 
> No objection to merging patches 34, 35, and 38.

Hi @martinwhitaker ,
Do you means definition it in makefile like this:
image
If so, we can remove the boot.h in some file that don't need it.

martinwhitaker:
Hi @kilaterlee , yes that's exactly what I meant.

kilaterlee:
OK, I will submit the next version as soon as possible, I have a bit busy in past few days.

debrouxl:
Yeah, good idea.
Maybe it should be `-DCONFIG_BITS=64` or `-DARCH_BITS=64` ? I thought about that because of the RV128 PE machine type... not that I'd expect a 128-bit architecture to become popular, but a single define ought to simplify preprocessor conditionals if someone wants to add support for a 128-bit architecture.

kilaterlee:
> Yeah, good idea. Maybe it should be `-DCONFIG_BITS=64` or `-DARCH_BITS=64` ? I thought about that because of the RV128 PE machine type... not that I'd expect a 128-bit architecture to become popular, but a single define ought to simplify preprocessor conditionals if someone wants to add support for a 128-bit architecture.

I also prefer `-DCONFIG_BITS=64` or `-DARCH_BITS=64`, but there is a question, if someone forgets or doesn't define it in makefile, what will compiler do? Considers it as 0? Maybe we needs to test that.

debrouxl:
I have just applied patches 34 (+ RISC-V machine numbers), 35 and 38. 36 is also trivial, but is dead code until non-x86 archs are supported.

IMO, not testing for undefined CONFIG_BITS / ARCH_BITS (let's say the latter) is fine: Makefiles for new architectures are modified copies of existing Makefiles, so it's going to be hard to forget defining it; missing the fact that it's undefined is going to be even harder :)

lib/assert.h
```diff
@@ -18,7 +18,11 @@
 static inline void assert(int expr)
 {
     if (!expr) {
+#if defined(__i386__) || defined(__x86_64__)
         __asm__ __volatile__ ("int $3");
+#elif defined(__loongarch_lp64)
+        __asm__ __volatile__ ("break 03");
```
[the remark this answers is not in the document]
kilaterlee: OK, will use `break 0` next time.

build64/la64/Makefile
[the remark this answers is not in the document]
kilaterlee: OK, I will adjust it to: `CC ?= loongarch64-linux-gnu-gcc`

kilaterlee:
Hi @debrouxl @martinwhitaker ,
Does memtest86 using the company's copyright? I found that all copyrights in existing code are personal.

system/imc/loongson/loongson_la.c
kilaterlee: IMO, Loongson's CPU roadmap does not define other 3B6000 series except 3B6000 and 3B6000M. So next time I will add another logic to distinguish them.

xen0n: > IMO, Loongson's CPU roadmap does not define other 3B6000 series except 3B6000 and 3B6000M. So next time I will add another logic to distinguish them.

Isn't there 3B6600 on the roadmap?

kilaterlee: Yes, it is. But it probably be 2 MC not 4, and it will be an improved generation that will be presented with DDR5, this submit only for DDR4.

system/memrw16.h
debrouxl: I've just spent time implementing what I had in mind on that front, and pushed the WIP result on the `memrw_refactor` branch. It's slightly less straightforward than I had originally envisioned, because one cannot just put #if chains inside macros, but there are other ways to achieve what I wanted :)

Adding support for another ISA becomes about a dozen lines of code, adding the 8-bit + 16-bit function becomes about 6 lines of code... and the likelihood of bugs, such as the unused `flush64()` currently using `movl`, is reduced, still without slaughtering maintainability, IMO.

kilaterlee: > I've just spent time implementing what I had in mind on that front, and pushed the WIP result on the `memrw_refactor` branch. It's slightly less straightforward than I had originally envisioned, because one cannot just put #if chains inside macros, but there are other ways to achieve what I wanted :)
> 
> Adding support for another ISA becomes about a dozen lines of code, adding the 8-bit + 16-bit function becomes about 6 lines of code... and the likelihood of bugs, such as the unused `flush64()` currently using `movl`, is reduced, still without slaughtering maintainability, IMO.

Hi @debrouxl ,
The code is very creative, and your ideas is very clever, one more question:
Is the WIP not finished yet? I don't see the 8-bit and 16-bit operations.
image

debrouxl: That's correct, it's not finished. I added the LA64 support to show how I envision things to work for multi-ISA support on top of the refactor, but I didn't add the 8-bit and 16-bit operations because they'll be added by this PR.

Should Martin and Sam not frown on my changes, I'd integrate a variant of my commit (switching to a single memrw.h file and removing the LA64 support) to the `main` branch, leaving the addition of LA64 support + the 8-bit and 16-bit operations to you :)

x86fr: I love one-liners and it's for sure creative, so fine to me. 
It's way more understandable that some obscure bit manipulation macros we have. 😅

martinwhitaker: @debrouxl, in general I'm happy with the change, but I'd like to see the Doxygen documentation of the files and functions restored.

debrouxl: I've just pushed a `memrw_refactor_v2` branch which contains the single `memrw.h` file I mentioned, and the slight fallout in other files.
The Doxygen documentation of the file was and is there, though it's applied to the definition of `__*_FUNC()` macros only... maybe you mean that the Doxygen comments should be applied to the `__*_FUNC(32/64)` instances ? It doesn't seem to help Doxygen generate documentation for the `read*()` functions, though.

debrouxl:
Any additional comments on the new version of patch 1 ?

system/hwquirks.c
debrouxl: You could make the x86-specific quirks depend on i386 or x86_64, though, if you don't want to clutter LA64 binaries with dead code for x86/x86_64 workarounds.

x86fr: This remark is also valid for A LOT of other code everywhere. I don't think adding a ton of IF is a good idea. At some point, if having unreachable code is really an issue (I don't think so), a separate build is the correct solution.

system/memrw16.h
martinwhitaker: The documentation of the `memrw32.h` / `memrw64.h` files is no longer there. This is what I get from your branch:
screenshot2a
compared to this from the main branch:
screenshot1a
screenshot1b
Yes, the `memrw.h` file has documentation:
screenshot2b
screenshot2c
but I don't think that's as helpful.

You can get Doxygen to generate documentation for the real functions by setting the `EXPAND_AS_DEFINED` variable in the `Doxygen` file to
````
EXPAND_AS_DEFINED      = __READ_FUNC __WRITE_FUNC __FLUSH_FUNC
````

Comment (debrouxl) [reply system/memrw16.h:22]: Alright. I have modified the commit on the `memrw_refactor_v2` branch to add the change to Doxyfile you suggested, and duplicate the comments to the uses of the macros, which restores documentation for the individual functions :)

Comment (martinwhitaker) [reply system/memrw16.h:22]: Thanks. Of course I don't know if anyone actually uses the Doxygen documentation...

File: system/hwquirks.c (hunk 5 above)
Comment (kilaterlee) [reply system/hwquirks.c:105]: Yes, the architecture-related code has been separated, but some codes still together. I will think about this issue.

**Review by debrouxl — CHANGES_REQUESTED** [review]
* you should make a single commit for the memrw.h changes, currently patches 12-15: the diffstat will be more palatable after the impending integration of #415 ;
* the LA64 `__inb/__inw/__inl` and `__outb/__outw/__outl` in patch 18 should be written using macros, like the x86/x86_64 functions from the same `io.h` file, even if there are more assembly instructions on LA64 than on x86/x86_64;
* likewise for the `mmio_*` functions in patch 27, though the inspiration could also be drawn from the memrw.h functions;
* I'd move the changes from patch 10 "system/cpuinfo: Add Loongson family CPU IMC support" to patch 38 "system/loongarch: Add LoongArch ARCH specific files", since they're unused until that point;
* likewise for the CPU IMC handler: squash the tiny patch 35 into patch 36;
* likewise, I'd squash patches 16 "pci: Add Loongson PCI vendor ID", 17 "system/hwquirks: Add Loongson 7A chipset EHCI workaround" and 25 "system/usbhcd: Add hwquirk support" together;
* `system/loongarch/smp.c` has too many spurious differences with `system/smp.c`, which make the diff harder to read. When those are removed, we may be able to see about using a single file, i.e. moving all LA64 changes to `system/smp.c`. The cost would be several preprocessor conditionals... but IMO, those are better in the long run than having to maintain largely duplicated files for currently 2 architecture families, maybe one day 3 or 4.

**Comment by kilaterlee** [comment]
> * you should make a single commit for the memrw.h changes, currently patches 12-15: the diffstat will be more palatable after the impending integration of [Refactor the memrw functions to reduce the redundancy #415](https://github.com/memtest86plus/memtest86plus/pull/415) ;
> * the LA64 `__inb/__inw/__inl` and `__outb/__outw/__outl` in patch 18 should be written using macros, like the x86/x86_64 functions from the same `io.h` file, even if there are more assembly instructions on LA64 than on x86/x86_64;
> * likewise for the `mmio_*` functions in patch 27, though the inspiration could also be drawn from the memrw.h functions;
> * I'd move the changes from patch 10 "system/cpuinfo: Add Loongson family CPU IMC support" to patch 38 "system/loongarch: Add LoongArch ARCH specific files", since they're unused until that point;
> * likewise for the CPU IMC handler: squash the tiny patch 35 into patch 36;
> * likewise, I'd squash patches 16 "pci: Add Loongson PCI vendor ID", 17 "system/hwquirks: Add Loongson 7A chipset EHCI workaround" and 25 "system/usbhcd: Add hwquirk support" together;
> * `system/loongarch/smp.c` has too many spurious differences with `system/smp.c`, which make the diff harder to read. When those are removed, we may be able to see about using a single file, i.e. moving all LA64 changes to `system/smp.c`. The cost would be several preprocessor conditionals... but IMO, those are better in the long run than having to maintain largely duplicated files for currently 2 architecture families, maybe one day 3 or 4.

Actually, I am not satisfied with the LoongArch version of `smp.c`, because it contains most of the same functions as the X86 version. I suggest separating the x86 version like ACPI table functions into a separate file, and only keep the ARCH-specific code in the `smp.c` of other ARCH.

File: system/i2c.c
Comment (x86fr) [inline system/i2c.c:1]: There is something confusing about the split between spd.c, smbus.c, and i2c.c.

Splitting the SPD-decoding code from the SMBus/I2C code is fine. The issue lies in the distinction between I2C and SMBus. As I see the code here, it appears that I2C is only for LA and SMBus for x86, which is not accurate.

SMBus is basically a subset of I2C. I initially retained Intel's "SMBus" terminology, but I don't think we ever used any of the (very few) specific features that originally differentiated SMBus from I2C. It would make sense to rename all references to SMBus as I2C.

I think smbus.c should be removed, and all I2C/"SMBus" code for both x86 and LA should be merged into a single i2c.c file.

Comment (kilaterlee) [reply system/i2c.c:1]: Hi @x86fr , I will try to merge the smbus.c into i2c.c in the next commit, and @debrouxl request for the change will also appear in the next submit.

**Comment by debrouxl** [comment]
I like the logical smbus / spd / i2c split requested by Sam, but the "physical" split, i.e. building files separately after removing a bunch of `static`, is a significant optimization regression (~900 bytes here on amd64 !): `print_smbus_info` currently greatly enjoys inlining of the `parse_spd_*` functions...
It can be trivially overcomed, though, *while keeping the logical split*: basically, restore the `static`s, and `#include "spd.c"` in `i2c.c`. On top of "system: Split the spd functions in smbus.c" and "system/smbus: Rename smbus.c to i2c.c" (which should be squashed), the following patch:

```
index 3ae434e..fcd05b2 100644
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -35,17 +35,16 @@ SYS_OBJS = system/acpi.o \
            system/heap.o \
            system/hwctrl.o \
            system/hwquirks.o \
+           system/i2c.o \
            system/keyboard.o \
-           system/ohci.o \
            system/memctrl.o \
+           system/ohci.o \
            system/pci.o \
            system/pmem.o \
            system/reloc.o \
            system/screen.o \
            system/serial.o \
            system/smbios.o \
-           system/spd.o \
-           system/i2c.o \
            system/smp.o \
            system/temperature.o \
            system/timers.o \
index 0b65a48..30f1ebb 100644
--- a/build64/Makefile
+++ b/build64/Makefile
@@ -35,18 +35,17 @@ SYS_OBJS = system/acpi.o \
            system/font.o \
            system/hwctrl.o \
            system/heap.o \
+           system/i2c.o \
            system/hwquirks.o \
            system/keyboard.o \
-           system/ohci.o \
            system/memctrl.o \
+           system/ohci.o \
            system/pci.o \
            system/pmem.o \
            system/reloc.o \
            system/screen.o \
            system/serial.o \
            system/smbios.o \
-           system/spd.o \
-           system/i2c.o \
            system/smp.o \
            system/temperature.o \
            system/timers.o \
diff --git a/system/i2c.c b/system/i2c.c
index 8a134d3..2211204 100644
--- a/system/i2c.c
+++ b/system/i2c.c
@@ -15,7 +15,9 @@
 #include "smbus.h"
 #include "smbios.h"
 #include "hwquirks.h"
-#include "spd.h"
+
+#include "spd.c"
+
 
 int smbdev, smbfun;
 unsigned short smbusbase = 0;
diff --git a/system/spd.c b/system/spd.c
index c87df02..ffdec23 100644
--- a/system/spd.c
+++ b/system/spd.c
@@ -14,13 +14,14 @@
 
 ram_info ram = { 0, 0, 0, 0, 0, 0, "N/A"};
 
+static uint8_t get_spd(uint8_t slot_idx, uint16_t spd_adr);
+
 static inline uint8_t bcd_to_ui8(uint8_t bcd)
 {
     return bcd - 6 * (bcd >> 4);
 }
 
-
-void print_spdi(spd_info spdi, uint8_t lidx)
+static void print_spdi(spd_info spdi, uint8_t lidx)
 {
     uint8_t curcol;
     uint16_t i;
@@ -89,7 +90,7 @@ void print_spdi(spd_info spdi, uint8_t lidx)
     }
 }
 
-void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len)
+static void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len)
 {
     uint8_t sku_len;
 
@@ -113,7 +114,7 @@ void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len)
     sku[sku_len] = '\0';
 }
 
-void parse_spd_ddr5(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_ddr5(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "DDR5";
 
@@ -293,7 +294,7 @@ void parse_spd_ddr5(spd_info *spdi, uint8_t slot_idx)
     spdi->isValid = true;
 }
 
-void parse_spd_ddr4(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_ddr4(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "DDR4";
 
@@ -412,7 +413,7 @@ void parse_spd_ddr4(spd_info *spdi, uint8_t slot_idx)
     spdi->isValid = true;
 }
 
-void parse_spd_ddr3(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_ddr3(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "DDR3";
 
@@ -569,7 +570,7 @@ void parse_spd_ddr3(spd_info *spdi, uint8_t slot_idx)
     spdi->isValid = true;
 }
 
-void parse_spd_ddr2(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_ddr2(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "DDR2";
 
@@ -706,7 +707,7 @@ void parse_spd_ddr2(spd_info *spdi, uint8_t slot_idx)
     spdi->isValid = true;
 }
 
-void parse_spd_ddr(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_ddr(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "DDR";
 
@@ -795,7 +796,7 @@ void parse_spd_ddr(spd_info *spdi, uint8_t slot_idx)
     spdi->isValid = true;
 }
 
-void parse_spd_rdram(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_rdram(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "RDRAM";
 
@@ -868,7 +869,7 @@ void parse_spd_rdram(spd_info *spdi, uint8_t slot_idx)
     spdi->isValid = true;
 }
 
-void parse_spd_sdram(spd_info *spdi, uint8_t slot_idx)
+static void parse_spd_sdram(spd_info *spdi, uint8_t slot_idx)
 {
     spdi->type = "SDRAM";
 
diff --git a/system/spd.h b/system/spd.h
index 2ac94b4..bd15f2c 100644
--- a/system/spd.h
+++ b/system/spd.h
@@ -49,16 +49,4 @@ typedef struct ram_infos {
 
 extern ram_info ram;
 
-void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len);
-
-void parse_spd_rdram (spd_info *spdi, uint8_t slot_idx);
-void parse_spd_sdram (spd_info *spdi, uint8_t slot_idx);
-void parse_spd_ddr   (spd_info *spdi, uint8_t slot_idx);
-void parse_spd_ddr2  (spd_info *spdi, uint8_t slot_idx);
-void parse_spd_ddr3  (spd_info *spdi, uint8_t slot_idx);
-void parse_spd_ddr4  (spd_info *spdi, uint8_t slot_idx);
-void parse_spd_ddr5  (spd_info *spdi, uint8_t slot_idx);
-void print_spdi(spd_info spdi, uint8_t lidx);
-uint8_t get_spd(uint8_t slot_idx, uint16_t spd_adr);
-
 #endif // SPD_H
```

restores the binary:
```
Before the split:
  12264    1446      18   13728    35a0 system/smbus.o

After the split:
   2795       2      18    2815     aff system/i2c.o
  10351    1444       0   11795    2e13 system/spd.o

With fixup:
  12264    1446      18   13728    35a0 system/i2c.o
```

**Comment by debrouxl** [comment]
Even more importantly, the LA64 I2C readout code needs to be made conditional on `__loongarch_lp64`, and conversely, the SMBus readout code needs to be made conditional on x86/x86_64, since the two current reading methods are ISA-exclusive in practice. ARM or RISC-V platforms would have different methods as well. Building the LA64 code on x86/x86_64 costs ~1.5 KB of effectively dead code + data, and compiling the more complex x86/x86_64 code on LA64 certainly costs far more...

I have pushed something to that effect on the `410_smbus` branch, since it's getting late here, but as I'm writing this message, I don't think that what I did it's the best approach down the road. As it stands, `print_smbus_startup_info()` is the only bit from `i2c.c` common to both x86/x86_64 and LA64.
I see two main ways to (almost) completely avoid sprinkling preprocessor conditionals throughout the current `i2c.c`:
* move x86/x86_64 code to a restored `smbus.c` which includes `spd.c`, keep LA64 code in `i2c.c` which includes `spd.c`, and build `smbus.c` xor `i2c.c` (probably to be renamed e.g. `smbus_x86.c` and `i2c_la64.c`) in the Makefiles;
* reorder functions in the current `i2c.c` in order to minimize preprocessor conditionals.
In both cases, a slight refactor - and certainly a rename, while at it - of `print_smbus_startup_info()` is needed.

I'd think that the former approach of compiling separate files per ISA scales better with the addition of support for select ARM and RISC-V platforms, if someone contributes it... but it's not a big deal either way.

**Comment by kilaterlee** [comment]
> Even more importantly, the LA64 I2C readout code needs to be made conditional on `__loongarch_lp64`, and conversely, the SMBus readout code needs to be made conditional on x86/x86_64, since the two current reading methods are ISA-exclusive in practice. ARM or RISC-V platforms would have different methods as well. Building the LA64 code on x86/x86_64 costs ~1.5 KB of effectively dead code + data, and compiling the more complex x86/x86_64 code on LA64 certainly costs far more...
> 
> I have pushed something to that effect on the `410_smbus` branch, since it's getting late here, but as I'm writing this message, I don't think that what I did it's the best approach down the road. As it stands, `print_smbus_startup_info()` is the only bit from `i2c.c` common to both x86/x86_64 and LA64. I see two main ways to (almost) completely avoid sprinkling preprocessor conditionals throughout the current `i2c.c`:
> 
> * move x86/x86_64 code to a restored `smbus.c` which includes `spd.c`, keep LA64 code in `i2c.c` which includes `spd.c`, and build `smbus.c` xor `i2c.c` (probably to be renamed e.g. `smbus_x86.c` and `i2c_la64.c`) in the Makefiles;

This way is looks like the way of old version was commit, except the old version used `spd.o`, and your version include `spd.c` right?
> * reorder functions in the current `i2c.c` in order to minimize preprocessor conditionals.
>   In both cases, a slight refactor - and certainly a rename, while at it - of `print_smbus_startup_info()` is needed.
> 
> I'd think that the former approach of compiling separate files per ISA scales better with the addition of support for select ARM and RISC-V platforms, if someone contributes it... but it's not a big deal either way.

**Comment by kilaterlee** [comment]
> > Even more importantly, the LA64 I2C readout code needs to be made conditional on `__loongarch_lp64`, and conversely, the SMBus readout code needs to be made conditional on x86/x86_64, since the two current reading methods are ISA-exclusive in practice. ARM or RISC-V platforms would have different methods as well. Building the LA64 code on x86/x86_64 costs ~1.5 KB of effectively dead code + data, and compiling the more complex x86/x86_64 code on LA64 certainly costs far more...
> 
> Emm... I think if the platform uses i2c to obtain SPD, the i2c method should be universal. They only need to adjust the i2c address and device address. When I designed it, I thought that this function could be reused by other ARCHs.
> 
> > I have pushed something to that effect on the `410_smbus` branch, since it's getting late here, but as I'm writing this message, I don't think that what I did it's the best approach down the road. As it stands, `print_smbus_startup_info()` is the only bit from `i2c.c` common to both x86/x86_64 and LA64. I see two main ways to (almost) completely avoid sprinkling preprocessor conditionals throughout the current `i2c.c`:
> > 
> > * move x86/x86_64 code to a restored `smbus.c` which includes `spd.c`, keep LA64 code in `i2c.c` which includes `spd.c`, and build `smbus.c` xor `i2c.c` (probably to be renamed e.g. `smbus_x86.c` and `i2c_la64.c`) in the Makefiles;
> 
> This way is looks like the way of old version was commit, except the old version used `spd.o`, and your version include `spd.c` right?
> 
> > * reorder functions in the current `i2c.c` in order to minimize preprocessor conditionals.
> >   In both cases, a slight refactor - and certainly a rename, while at it - of `print_smbus_startup_info()` is needed.
> > 
> > I'd think that the former approach of compiling separate files per ISA scales better with the addition of support for select ARM and RISC-V platforms, if someone contributes it... but it's not a big deal either way.

@x86fr @debrouxl 
Hi Sam and Lionel,
About the i2c.c, which way should I go?

**Comment by x86fr** [comment]
As mentioned by @debrouxl, for architecture-dependent code like I2C, I think that compiling files according to the target ISA is the way to go.

x86 has a 25+ year legacy for I2C, with many proprietary tricks. Also, this is one of the very few pieces of code that can physically damage hardware (e.g., while writing the early DDR5 code, I accidentally wrote to SPD, preventing boot), so it must be tested extensively. Not having code for other ISAs in the binary makes it easier to avoid potential side effects.

Let's also get rid off the "smbus" terminology (rename `print_smbus_startup_info()` to `print_spd_startup_info()` ). 
For files, `i2c_x86.c` and `i2c_la64.c` sounds good.

**Comment by kilaterlee** [comment]
> As mentioned by @debrouxl, for architecture-dependent code like I2C, I think that compiling files according to the target ISA is the way to go.
> 
> x86 has a 25+ year legacy for I2C, with many proprietary tricks. Also, this is one of the very few pieces of code that can physically damage hardware (e.g., while writing the early DDR5 code, I accidentally wrote to SPD, preventing boot), so it must be tested extensively. Not having code for other ISAs in the binary makes it easier to avoid potential side effects.
> 
> Let's also get rid off the "smbus" terminology (rename `print_smbus_startup_info()` to `print_spd_startup_info()` ). For files, `i2c_x86.c` and `i2c_la64.c` sounds good.

OK, I will adjust in next commit:
1. Split SPD series function from smbus.c.
2. Rename smbus.c to i2c_x86.c
3. Add i2c.c for LoongArch.

BTW, what about the approach mentioned by @debrouxl of having `i2c_x86.c` include `spd.c`? It looks like a 900 bytes reduction on x64 platforms, but I don't think it's appropriate for some `.c` to include other `.c` files, but it does reduce about 1K bytes, which should be a trade-off.

**Comment by debrouxl** [comment]
Well... in the absence of LTO, it's entirely appropriate to include .c files from other .c files.
The alternative is to yield code which is larger, nullifying at once multiple rounds of optimizations I did, and besides, has worse encapsulation. Refactoring for better portability and maintainability is a good thing to do, but it doesn't have to worsen generated code when it's so easy to write the code in such a way that it doesn't :)

In late 2022, I spent time trying to add LTO support to this code base, because inlining functions which have a single caller is among the optimizations which LTO usually does well, and then LTO performs other optimizations which are hard to do manually. However, my attempts to link working binaries were ultimately unsuccessful. I published the `lto` branch so as not to lose the work.
A year and a half, and two compiler generations later, it might be good to revisit this work, even if we may want it to use only for release builds, given the impact on debuggability and build time.

**Comment by kilaterlee** [comment]
> Well... in the absence of LTO, it's entirely appropriate to include .c files from other .c files. The alternative is to yield code which is larger, nullifying at once multiple rounds of optimizations I did, and besides, has worse encapsulation. Refactoring for better portability and maintainability is a good thing to do, but it doesn't have to worsen generated code when it's so easy to write the code in such a way that it doesn't :)
> 
> In late 2022, I spent time trying to add LTO support to this code base, because inlining functions which have a single caller is among the optimizations which LTO usually does well, and then LTO performs other optimizations which are hard to do manually. However, my attempts to link working binaries were ultimately unsuccessful. I published the `lto` branch so as not to lose the work. A year and a half, and two compiler generations later, it might be good to revisit this work, even if we may want it to use only for release builds, given the impact on debuggability and build time.

OK, I will include the `spd.c` in next commit, BTW, if we enable the LTO in the future, we need to optimize it again and remove `#include "spd.c"`.

**Comment by kilaterlee** [comment]
> > As mentioned by @debrouxl, for architecture-dependent code like I2C, I think that compiling files according to the target ISA is the way to go.
> > x86 has a 25+ year legacy for I2C, with many proprietary tricks. Also, this is one of the very few pieces of code that can physically damage hardware (e.g., while writing the early DDR5 code, I accidentally wrote to SPD, preventing boot), so it must be tested extensively. Not having code for other ISAs in the binary makes it easier to avoid potential side effects.
> > Let's also get rid off the "smbus" terminology (rename `print_smbus_startup_info()` to `print_spd_startup_info()` ). For files, `i2c_x86.c` and `i2c_la64.c` sounds good.
> 
> OK, I will adjust in next commit:
> 
> 1. Split SPD series function from smbus.c.
> 2. Rename smbus.c to i2c_x86.c
> 3. Add i2c.c for LoongArch.
> 
> BTW, what about the approach mentioned by @debrouxl of having `i2c_x86.c` include `spd.c`? It looks like a 900 bytes reduction on x64 platforms, but I don't think it's appropriate for some `.c` to include other `.c` files, but it does reduce about 1K bytes, which should be a trade-off.

@x86fr Another quesiton: should we drop all of the "smbus" terminology? E.g. rename `nv_mcp_get_smb` to `nv_mcp_get_i2c` and so on?

**Comment by x86fr** [comment]
> OK, I will include the `spd.c` in next commit, BTW, if we enable the LTO in the future, we need to optimize it again and remove `#include "spd.c"`.

While including .c files within other .c files may be valid and easier to optimize from a compiler's perspective, it leads to a loss of modularity, complicates dependency management, can interfere with code tree parsing in many IDEs, and adds confusion. As a dev encountering this code for the first time, I would question the reason for this unusual inclusion.

Memtest86+ already contains code that may appear strange at first glance, such as seemingly 'useless' initial declarations in arrays, but these are included for good reasons.

Given that this practice can be applied in several other places and result in significant structural changes, I believe we should avoid adding such .c includes without establishing a clear code convention (if we decide to proceed with this approach).

Let's allow @martinwhitaker to have the final say on this matter. He organized the code very cleanly in v6+ and wrote the code convention.

> @x86fr Another quesiton: should we drop all of the "smbus" terminology? E.g. rename nv_mcp_get_smb to nv_mcp_get_i2c and so on?

Just rename `print_smbus_startup_info()` to `print_spd_startup_info()` (and potential other smbus references used by the LA64 implementation) and I will rename the other x86-specific references later. I will handle renaming the x86-specific references later. I need to write some code to support the latest southbridge from Intel and AMD.

**Comment by debrouxl** [comment]
> it leads to a loss of modularity

Here, we're replacing a single file containing SMBus + SPD by two files containing SMBus and SPD separately but compiled together. I view this as both a source modularity improvement over the single file (due to the SMBus / SPD split, these functions now appear in separate files), and an encapsulation improvement over two files compiled separately (since the many symbols with a single caller, used below `print_spd_startup_info()`, don't leak across TUs) ?
For the same reason, having the single-function IMC support files included into their caller keeps the source modularity and improves encapsulation.

> can interfere with code tree parsing in many IDEs

Er... seriously, there are IDEs which can't behave when .c files include .c files ? I wasn't aware of that, but dealing with IDEs which suck is a poor reason for programmers to pessimize the generated code :)

I've created #424 for the second iteration of the broken LTO support code. If 1) we can make the LTO build work and 2) it actually yields the expected effect for SMBus + SPD and IMC, plans for including .c files into other .c files can be scrapped.

**Comment by 01e3** [comment]
Sorry for coming here a little bit late and potentially creating more noise, but I wonder if there is another path we can take for the smbus.c -> smbus.c + spd.c split. Assuming we want just the SPD parsing code in spd.c, nothing more, nothing less and we eventually want to make get_spd() a platform specific callback, which means removing static from it, here is potential implementation to consider that does not require to include *.c files: https://github.com/01e3/memtest86plus/commit/5894c732306c397c65a990fa0b6d53aa1d678629

I hope this may be a relatively non-controversial change, allowing to keep working on this PR while I can work on converting floating-point into static-point.

If all three of you are fine with this idea, I can create a PR.

before:
```
$ make && size -t */smbus.o ; stat -c "%n %s" memtest{.bin,.efi,_shared,_shared.bin}
  12352    1446      18   13816    35f8 system/smbus.o
  12352    1446      18   13816    35f8 (TOTALS)
memtest.bin 151896
memtest.efi 153088
memtest_shared 176120
memtest_shared.bin 150360

```

after:
```
$ make ; size -t */{smbus,spd}.o ; stat -c "%n %s" memtest{.bin,.efi,_shared,_shared.bin}
   4407    1446      18    5871    16ef system/smbus.o
   7885       0       0    7885    1ecd system/spd.o
  12292    1446      18   13756    35bc (TOTALS)
memtest.bin 151896
memtest.efi 153088
memtest_shared 176184
memtest_shared.bin 150360
```

**Comment by debrouxl** [comment]
This looks like it can make everyone happy: it keeps the split of the SPD parsing code into its own file, without either slaughtering code generation or having to resort to including .c files in other .c files... I like it, thanks for the idea and implementation :)

**Comment by x86fr** [comment]
Looks fine to me

**Comment by 01e3** [comment]
Created https://github.com/memtest86plus/memtest86plus/pull/426, @kilaterlee please take a look and let me know if you have any concerns. I hope this is a good foundation for tour changes.

**Comment by kilaterlee** [comment]
> Created #426, @kilaterlee please take a look and let me know if you have any concerns. I hope this is a good foundation for tour changes.

This is a clever design, I think it's great, and I hope your PR gets merged soon. :)

**Comment by kilaterlee** [comment]
@x86fr @debrouxl 
I suggest #426 should be merged before #410,  witch would remove d88e0c2055f48be250e67af45daaa827bf50696e.

**Comment by kilaterlee** [comment]
> Some conflicts must be resolved due to #426 merge.
> 

OK, I will fix them today.
> We won't merge any more PR until this one is merged.

Thanks. :)

File: build64/la64/ldscripts/memtest_shared.lds [hunk 28]
```diff
@@ -0,0 +1,56 @@
...
+ENTRY(startup64);
+SECTIONS {
+	. = 0;
+	.text : {
+		_start = .;
+		*(.text)
+		*(.text.*)
+		*(.plt)
+		_etext = . ;
+	} = 0x9090
```
debrouxl, line 13: IIUC, you want to replace that by the encoding of NOP on your architecture.

debrouxl:
Besides the nitpick I've just posted, it looks reasonable to me.
After this PR is merged, for symmetry reasons with `*/loongarch/*` files added by this PR, we can move several x86-specific files to `*/x86/*`, but this clearly doesn't have to block the merging of this PR.
The refactors done before merging this PR have already benefited the memtest86+ code base :)

build64/la64/ldscripts/memtest_shared.lds
kilaterlee: Yes! I forgot to adjust this instruction when I porting the code, it should be `= 0x00004003`. :)
I will fix in the next commit. Thanks!

kilaterlee:
Hi every one,
Thanks for you guys working so hard to review this series, I think this series may or may not have taken the longest time to review in this repo. :)
Because memtest86+ uses `-b binary` option when linking, our compiler team added this option and meged it into binutils master branch a few days ago, it will be supported by default in binutils 2.44, here is a cross compiler based the binutils master branch with GCC 14.2: https://github.com/YongbaoOS/Yongbao-Toolchains/releases/download/2024.8.6/x86_64-cross-tools-loongarch64-binutils_git60d4fed4e364-gcc_14.2.0.tar.xz . This compiler can build the LA64 code and run correctly.
If passable, I want add the LA64 CI workflows after this series is meged, may I?

kilaterlee:
> Hi every one, Thanks for you guys working so hard to review this series, I think this series may or may not have taken the longest time to review in this repo. :) Because memtest86+ uses `-b binary` option when linking, our compiler team added this option and meged it into binutils master branch a few days ago, it will be supported by default in binutils 2.44, here is a cross compiler based the binutils master branch with GCC 14.2: https://github.com/YongbaoOS/Yongbao-Toolchains/releases/download/2024.8.6/x86_64-cross-tools-loongarch64-binutils_git60d4fed4e364-gcc_14.2.0.tar.xz . This compiler can build the LA64 code and run correctly. If passable, I want add the LA64 CI workflows after this series is meged, may I?

I have added LA64 build CI check(668144254929ffab3ec97a86f8c646d43de3cb3c). If you don't like it, PLS let me know, and I will remove it. :)

x86fr:
That's more than welcome! thank you.

app/display.c
```diff
@@ -144,11 +144,14 @@ void display_init(void)
     clear_screen_region(ROW_FOOTER, 0, ROW_FOOTER, SCREEN_WIDTH - 1);
     prints(ROW_FOOTER, 0, " <ESC> Exit  <F1> Configuration  <Space> Scroll Lock");
     prints(ROW_FOOTER, 64, MT_VERSION "." GIT_HASH);
```
01e3, line 146: MT_VERSION "." GIT_HASH is 4+1+7 = 12 characters log so arch-id has to start at column 76, otherwise it will overlap with GIT_HASH (or word "unknown)".  The code below shifts everything to 74, where we really need only more more character (at least for now) to fit "la64".

I think we just need to shift it one character left:
`prints(ROW_FOOTER, 63, MT_VERSION "." GIT_HASH);`

Then, arch-id should start at row 75.

Perhaps something like this would work even better:
```
    prints(ROW_FOOTER, 63, MT_VERSION "." GIT_HASH "."
#if defined (__x86_64__)
        "x64"
#elif defined (__i386__)
        "x32"
#elif defined (__loongarch_lp64)
        "la64"
#else
        "??"
#endif
    );
```

Note that I have pending changes for this file, if we don't want to change this patchset again, I can fix it later with a different PR.

system/i2c_universal.c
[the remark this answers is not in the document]
martinwhitaker: Or put in the loongarch subdirectory? And i2c.h above also looks to be loongarch-specific.

system/loongarch/cpucfg.c
martinwhitaker, line 1: Shouldn't this be named cpuid.c?

boot/loongarch/header.S
```diff
@@ -0,0 +1,154 @@
...
+	.long	_virt_text_size			# VirtualSize
+	.long	_virt_text_start		# VirtualAddress
+	.long	_file_text_size			# SizeOfRawData
+	.long	_file_text_start		# PointerToRawData
+	.long	0				# PointerToRelocations
+	.long	0				# PointerToLineNumbers
+	.short	0				# NumberOfRelocations
+	.short	0				# NumberOfLineNumbers
+	.long	IMAGE_SCN_MEM_READ		\
+	      | IMAGE_SCN_MEM_EXECUTE		\
```
martinwhitaker, line 114: I suggest adding `IMAGE_SCN_MEM_WRITE` as I have just done in the x86 version..

martinwhitaker (commented):
A few comments from me, but nothing that couldn't be left for future tidying up.

system/io.h
```diff
@@ -145,4 +147,61 @@ outb_p (unsigned char __value, unsigned short int __port)
...
+#define LOONGSON_IO_PROT_BASE 0x0efdfc000000
+
+#define __IN(s, s1) \
+static inline RETURN_TYPE __in##s(unsigned short port) { \
+    RETURN_TYPE _v; \
+    __asm__ __volatile__ ( \
+        "li.d $t0, 0x28\n\t" \
+        "csrwr $t0, 0x0\n\t" \
+        "ld." s1 " %0, %1, 0 \n\t" \
+        "csrwr $t0, 0x0\n\t" \
```
xen0n, line 161: Why write `0x28` to `CSR.CRMD` for port MMIO? Just issuing accesses to the mapped "port" address should be enough.

The original CRMD content is not preserved so I highly suspect this is not the correct thing to do. Do you mean some kind of barrier here? If so `dbar 0` could be appropriate I guess.

This also applies to the `out` definitions below.

system/loongarch/cpucfg.c
xen0n: The LoongArch equivalent of x86 `cpuid` is indeed `cpucfg`. To add to the confusion, according to the LoongArch ISA Manual, the name `CPUID` refers to a CSR containing the "logical CPU index" -- but in this patchset the CSR gets named `CPUNUM` instead.

Because `cpuid.h` is shared across architectures though, I'd agree with renaming the file.

system/loongarch/cpuinfo.c
```diff
@@ -0,0 +1,281 @@
...
+
+static void determine_cache_size()
+{
+  uint32_t cache_info = 0;
+  uint32_t cache_present = 0;
+  uint32_t ways = 0;
+  uint32_t index = 0;
+  uint32_t line_size = 0;
+
+  // Get the persentence of caches
```
xen0n, line 60:
```suggestion
  // Get the presence of caches
```

system/loongarch/cpuinfo.c
```diff
@@ -0,0 +1,281 @@
...
+  cache_present = __cpucfg(0x10);
+
+  // L1 D cache persent
```
xen0n, line 63:
```suggestion
  // L1 D cache present
```

Same below.

system/loongarch/cpuinfo.c
```diff
@@ -0,0 +1,281 @@
...
+    index = 1 << ((cache_info >> 16) & 0xFF);
+    line_size = 1 << ((cache_info >> 24) & 0x7F);
+    l3_cache = (ways * index * line_size) / 1024;
+  }
+}
+
+static void determine_imc(void)
+{
+  // Check Loongson IMC
+  if (cpuid_info.vendor_id.str[0] == 'L') {
```
xen0n, line 94: Do we want to check more bytes here?

system/loongarch/cpuinfo.c
```diff
@@ -0,0 +1,281 @@
...
+}
+
+//------------------------------------------------------------------------------
+// Public Functions
+//------------------------------------------------------------------------------
+
+void cpuinfo_init(void)
+{
+    // Get cache sizes for most AMD and Intel CPUs. Exceptions for old
+    // CPUs are handled in determine_cpu_model().
```
xen0n, line 270: Remove the two lines? They seem to be leftover from the x86 port.

system/smp.c
```diff
@@ -682,11 +725,26 @@ static bool parse_slit(uintptr_t slit_addr)
...
+    if (mode == APIC_DELMODE_STARTUP) {
+        //
+        // Set the AP mailbox0
+        //
+        __iocsrwr_d((1ULL << 31 | ((0x0 << 1) + 1) << 2 | apic_id << 16 | (((uintptr_t)ap_startup_addr) & 0xFFFFFFFF00000000ULL)), 0x1048);
+        __iocsrwr_d((1ULL << 31 | (0x0 << 1) << 2 | apic_id << 16 | (((uintptr_t)ap_startup_addr << 32))), 0x1048);
+    }
+
+    //
+    // Trigger IPI interrupt
```
xen0n, line 744: "IPI" is "inter-processor interrupt". ;-)
```suggestion
    // Trigger IPI
```

system/imc/loongson/loongson_la.c
```diff
@@ -0,0 +1,132 @@
...
+
+void read_imc_sequence(void)
+{
+    imc.tCL     = (uint16_t)read8((uint8_t *)(MC_CONF_ADDRESS + 0x1060));
+    imc.tCL_dec = 0;
+    imc.tRP     = (uint16_t)read8((uint8_t *)(MC_CONF_ADDRESS + 0x1006));
+    imc.tRCD    = (uint16_t)read8((uint8_t *)(MC_CONF_ADDRESS + 0x1047));
+    imc.tRAS    = (uint16_t)read8((uint8_t *)(MC_CONF_ADDRESS + 0x1040));
+
+    switch ((read8((uint8_t *)(MC_CONF_ADDRESS + 0x1024)) & 0x7)) {
```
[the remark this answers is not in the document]
kilaterlee: Yes, it was my oversight, it should be 0x1203.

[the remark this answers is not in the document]
kilaterlee: > Yes, it was my oversight, it should be 0x1203.

The offset was wrong, in fact, I want to read the `Narrow Channel` field, which is better to read `Channel Width` or `Narrow Channel` ?

system/i2c_universal.c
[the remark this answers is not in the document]
kilaterlee: Emm... I think the i2c_universal method can be adpeted more ARCHs who uses the i2c bus to get the spd info, such as `get_spd` and `i2c_read_byte`, which only change the base address and device id, and then you can read the data. In addition, in `i2c.h`, those prefixed with `LOONGSON` are Loongson-specific, and the others can be reused.

boot/loongarch/header.S
kilaterlee: OK

system/imc/loongson/loongson_la.c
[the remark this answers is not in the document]
x86fr: The value reported in `imc.width` should be the whole memory controller width, so channel Width * number of active channels. Usually 128 bit for dual-channel DIMMs, 256 bits for Quad Channel, and so on.

system/i2c_universal.c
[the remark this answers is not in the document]
x86fr: Agree with Martin. Moving LA specific code for I2C to loongarch subdir is probably the way to go. This is currently too confusing as the code is only used by LA right now and seems to be "universal" (without function prefix or specific subdir).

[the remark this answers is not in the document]
kilaterlee: OK, I will work on the issues in this PR today, I will move them into `*/loongarch/*`. :)

system/imc/loongson/loongson_la.c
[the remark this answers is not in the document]
kilaterlee: OK, let me tweak the logic there.

system/hwquirks.c
kilaterlee: This workaround is used to turn off EHCI controller of Loongson7A series. Currently, memtest86+ doesn't support usb hot-plugging, and only keyboards. However, the EHCI controller and OHCI controller of Loongson7A seems to  be mutually exclusive, which means if the EHCI controller is probed and started , OHCI will no longer be able to scan any devices, so I turned it off.
@x86fr Aother quesiton, do we need to use the EHCI U-disk? If so, it seems reasonable to detect and start EHCI, if no, why we have to support the EHCI?

system/i2c_universal.c
```diff
@@ -0,0 +1,197 @@
...
+    } else {
+        return false;
+    }
+    return false;
+}
+
+static uint8_t i2c_read_byte(uint8_t *base, uint8_t dev_addr, uint16_t offset)
+{
+    uint8_t buf;
+    volatile uint8_t spd_adr = (uint8_t)(offset & 0xff);
```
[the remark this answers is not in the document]
kilaterlee: Because this file was meged into the i2c.c in previous commit, both x86 and LA use the same file, and the CI of the i386 platform cannot pass, because i386 passes the `offset` via a register, the `offset` size mismatch will be reported during build, so I added `volatile` to prevent compiler optimization.

system/io.h
kilaterlee: The old value is stored in $t0, please refer to Section 4.2.1 of the LoongAch Reference Manual Volume 1:
image

Updated `CRMD` to change the CPU to instruction fetch cacheable, load/store uncacheable, DA mode when access to an IO region is required.

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,476 @@
...
+    # Disable global intrrupts
+    li.w    $t0, (0x1 << 2)
+    csrxchg $zero, $t0, LOONGARCH_CSR_CRMD
+
+    # Init Core
+    li.w    $t0, 0xA8        # DA mode, load/store cacheable, instructions cacheable, interrupts disabled, DMWn has no effect.
+    csrwr   $t0, LOONGARCH_CSR_CRMD
+
+    //
+    // Make sure the legacy mode works properly.
```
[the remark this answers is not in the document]
kilaterlee: The link address is decided in the EFI header, so just clear the hight 32 bits is fine.
As your saied, the "Legacy mode" is mean using DMW method BIOS, and this mode is explained in the BIOS setup page, so I don't think it's necessary to comment further here.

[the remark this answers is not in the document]
xen0n: > The link address is decided in the EFI header, so just clear the hight 32 bits is fine. As your saied, the "Legacy mode" is mean using DMW method BIOS, and this mode is explained in the BIOS setup page, so I don't think it's necessary to comment further here.

Okay, then maybe "the legacy boot protocol" is better -- I wondered whether there's a "legacy mode" in memtest86+ for a bit while reviewing. ;-)

system/io.h
xen0n: > The old value is stored in $t0, please refer to Section 4.2.1 of the LoongAch Reference Manual Volume 1: image
> 
> Updated `CRMD` to change the CPU to instruction fetch cacheable, load/store uncacheable, DA mode when access to an IO region is required.

Ha, thanks! Apparently I've forgotten this somehow. It's totally fine then.

system/hwquirks.c
martinwhitaker: We support EHCI because some machines don't provide a companion controller and rely on downstream hubs to perform the high speed to low speed translation.

Did you try setting the `USB_IGNORE_EHCI` flag in `usb_init_options` to prevent the EHCI contoller being started? That would be a simpler workaround.

boot/loongarch/startup64.S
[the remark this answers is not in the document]
kilaterlee: Maybe "Legacy boot mode" ?

app/loongarch/interrupt.c
```diff
@@ -0,0 +1,232 @@
...
+    display_pinned_message(8, 0, "ESTAT: %0" CSR_REG_DIGITS "x", (uintptr_t)system_context->estat);
+
+    display_pinned_message(3, 25, "A0: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r4);
+    display_pinned_message(4, 25, "A1: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r5);
+    display_pinned_message(5, 25, "A2: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r6);
+    display_pinned_message(6, 25, "A3: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r7);
+    display_pinned_message(7, 25, "A4: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r8);
+    display_pinned_message(8, 25, "A5: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r9);
+    display_pinned_message(9, 25, "RA: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r1);
+    display_pinned_message(10, 25, "SP: %0" GP_REG_DIGITS "x", (uintptr_t)system_context->r3);
```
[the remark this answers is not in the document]
kilaterlee: OK

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,476 @@
...
+#define LOONGARCH_CSR_ECFG       0x4
+#define LOONGARCH_CSR_ESTAT      0x5
+#define LOONGARCH_CSR_ERA        0x6
+#define LOONGARCH_CSR_BADV       0x7
+#define LOONGARCH_CSR_BADI       0x8
+#define LOONGARCH_CSR_EBASE      0xC     // Exception entry base address
+#define LOONGARCH_CSR_CPUNUM     0x20    // CPU core number
+#define LOONGARCH_CSR_KS0        0x30
+#define LOONGARCH_CSR_TLBREBASE  0x88    // TLB refill exception entry
+#define LOONGARCH_CSR_DMWIN0     0x180   // 64 direct map win0: MEM & IF
```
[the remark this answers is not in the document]
kilaterlee: These comments are from the Linux kernel...

system/i2c.h
```diff
@@ -0,0 +1,69 @@
...
+#define RPA  0x6d
+
+#define LOONGSON_I2C0_ADDR 0x1fe00120
+#define LOONGSON_I2C1_ADDR 0x1fe00130
+#define LOONGSON_I2C2_ADDR 0x1fe00138
+
+typedef union {
+    struct {
+         uint32_t slot0_addr : 8;
+         uint32_t slot1_addr : 8;
```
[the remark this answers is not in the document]
kilaterlee: Because the number of slots in the same MC in the future is uncertain, currently it is usually 1 or 2, but I am not sure whether it will increase to 4, so I use bit fields to keep the structure size to 4 bytes.

build64/la64/Makefile
```diff
@@ -0,0 +1,205 @@
+CC ?= gcc
+LD ?= ld
+OBJCOPY ?= objcopy
+
+GIT = git
```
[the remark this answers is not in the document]
kilaterlee: OK

system/loongarch/cpuinfo.c
kilaterlee: Currently, LoongArch vendor is only Loongson, I guess it's not too late for us to add more byte checks here if another vendor name also starts with the letter "L" in the future. :)

boot/loongarch/startup64.S
```diff
@@ -0,0 +1,476 @@
...
+#define LOONGARCH_CSR_EUEN       0x2
+#define LOONGARCH_CSR_MISC       0x3
+#define LOONGARCH_CSR_ECFG       0x4
+#define LOONGARCH_CSR_ESTAT      0x5
+#define LOONGARCH_CSR_ERA        0x6
+#define LOONGARCH_CSR_BADV       0x7
+#define LOONGARCH_CSR_BADI       0x8
+#define LOONGARCH_CSR_EBASE      0xC     // Exception entry base address
+#define LOONGARCH_CSR_CPUNUM     0x20    // CPU core number
+#define LOONGARCH_CSR_KS0        0x30
```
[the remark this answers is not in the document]
kilaterlee: `KS0` also comes from the Linux kernel, and I think it’s better to keep the name consistent with the Linux kernel.

kilaterlee:
Hi all,
Sorry for the late update, been a bit busy these past few days.
This commit addresses most of the issues raised in the comments.

x86fr:
I think we're ready to merge this. I will Squash & Merge it in a couple of days if no objection.

debrouxl:
No objections. We'll perform various changes mentioned in the comment thread or review comments after merging the PR :)

debrouxl:
Sam, would it help if I click the Squash and merge button before you come back from your current business trip ?

x86fr:
Back today. Will do it in a couple of hours. Have a last thing to check

PR #417, opened 2024-07-16
Pass bit count from build system

debrouxl:
That's patch 1 of #410 , rebased onto the memrw changes from #415 . I'll rebase onto main when the latter is integrated.


martinwhitaker (approved):
Looks good to me.

x86fr (approved):
All good.

PR #419, opened 2024-07-21
Correct check for console "ttyS..." parameter

tormodvolden:
I think this was just a typo in commit 2e048a7c, and that the intention is to return early unless the string starts with "ttyS". The only way the old code would return is if the string is exactly "ttyS", and with any other string it would carry on to try read out a port number and speed.

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -152,7 +152,7 @@ static void parse_serial_params(const char *params)
     }
 
     // No TTY port passed, use default ttyS0
-    if (strncmp(params, "ttyS", 5) == 0) {
+    if (strncmp(params, "ttyS", 4) != 0) {
         return;
     }
 


01e3:
Hello Tormod and thank you for the patch!

I think the check is actually *mostly* correct given strings in C are null terminated so the 5 characters that are compared are "ttyS\0". If "params" points to memory that has exactly these values, then strncmp return 0, which is exactly what we want here, supporting all variants like "console", "console=", "console=ttyS", "console=ttyS0" to enable serial output on the "COM1" port using 115200 8n1.

Changing the code as you suggested could lead into reading garbage for the "console=ttyS" case by the code below, except currently we are protected by the next condition, that would also return for '\0' (0x00):
```
    // Configure TTY port or use default
    if (params[4] >= '0' && params[4] <= '3') {
        tty_address = serial_io_ports[params[4] - '0'];
    } else {
        return;
    }

```

What we may be missing is a more sophisticated check to catch things like console=abcd0,38400 given currently "abcd" part would be ignored and handled as "ttyS".

@x86fr Sam, do you think it is worth adding it?

tormodvolden:
Oh, I didn't know that just "ttyS" was a valid value. That of course changes everything. I can't have read the documentation carefully enough.

tormodvolden:
Actually, the README.md doesn't say that `=`, `ttyS` or. `x` are optional:
```
console=ttySx,y

    activate serial/tty console output, where x is one of the following IO port
    ... 
```

tormodvolden:
And I don't buy into your defense of the original code. You are not supposed to include the terminating zero character in a string match against a constant string. If you want to compare a (complete) string with "ttyS" you simply do `strcmp(s, "ttyS) == 0`, whereas using strncmp for this makes no sense. However you'd want to use strncmp when you want to match a substring, or if you are not sure the string has a terminating null and its buffer is smaller than the other string - the latter is not the case here, and I think the former is the original intention.

I'd like to believe that this was originally a typo or thinko, which had the unintended effect of allowing console, console=, console=ttyS (and console=abcd) to work all this time and some people got used to it and though it was intended to be this way. Would be interesting to know what the original author thinks! Sam?

x86fr:
The only supported console syntax is as stated on README.md (`console=ttySx,y`). 

The code will just fallback to standard COM1/115200bps parameter if the syntax is wrong. That's the expected behavior.

01e3:
Fair, in this case I agree with Tormod (we may want to update the comment however), just a note that in this situation enable_tty is set to true already, so "return" does not change anything - indeed, COM1,115220,8n1 will always be used.

Next, in addition to updating the "ttyS" check we may also want to change the following check:

```
    // Configure TTY port or use default
    if (params[4] >= '0' && params[4] <= '3') {
        tty_address = serial_io_ports[params[4] - '0'];
    } else {
        return;
    }
```

into something like:
```
    //  Use the default TTY port if wrong TTY port or end of string (\0)
    if (params[4] < '0' || params[4] > '3') {
        return;
   }

    // Configure TTY port
    tty_address = serial_io_ports[params[4] - '0'];
```

Technically the same, but I hope it would be easier to understand the code flow and the checks.

Also a note that according to the documentation, console=ttySx should also be considered as supported given "console=ttySx,y (...) and y is an optional baud rate (...) 115200 (default if not specified or invalid)".

BTW: Tormod, Memtest86+ is a "freestanding" system with very limited "stdc" functions defined. In particular, we do not have strcmp defined.

tormodvolden:
(blush) That would be a good reason for using strncmp instead! Sorry I missed that and for my "mansplaining".

I fully agree with your suggested change.

In any case, thanks for your review and the discussion!

x86fr:
Well, if you REALLY want that... I was about to directly push a commit to main, but it's so minor that I can't even find a decent commit message to justify. I added it to my TODO list for the next time I touch config.c.

tormodvolden:
I added 01e3's suggestion as another commit (since it is a totally orthogonal change), and added more description of the consequences to my first commit. But since the PR is closed, it doesn't show here and you'll have to look up my branch.

We can come back to this when there is less coffee involved :)

PR #421, opened 2024-07-23
efisetup: Ignore Unicode byte-order mark in command line

tormodvolden:
References #418

The Unicode command line  can have a Unicode BOM:
https://wiki.archlinux.org/title/EFISTUB#bcfg

### DIFF
--- a/boot/efisetup.c
+++ b/boot/efisetup.c
@@ -179,6 +179,11 @@ static int get_cmd_line_length(efi_loaded_image_t *image)
     efi_char16_t *cmd_line = (efi_char16_t *)image->load_options;
     int max_length = image->load_options_size / sizeof(efi_char16_t);
     int length = 0;
+    // Skip Unicode byte order mark if present
+    if (cmd_line[0] == u'\uFEFF') {
+        cmd_line = &cmd_line[1];
+        max_length--;
+    }
     while (length < max_length && cmd_line[length] > 0x00 && cmd_line[length] < 0x80) {
         length++;
     }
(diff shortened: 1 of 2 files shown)


debrouxl:
Interesting.
All of x86/x86_64, ARM/AArch64, RISC-V and LoongArch being either exclusively or typically little-endian, it's probably safe to assume that this code is supposed to be fed only the LE BOM, right ?

tormodvolden:
I don't know if the UEFI spec says anything about it. For what I know it may dictate that the data should be LE regardless of the machine byte order. Then what actual UEFI implementations on BE hardware do, could be another thing.

martinwhitaker:
The UEFI spec. states that `LoadOptions` points to a binary data buffer of `LoadOptionsSize` bytes. It says nothing about what might be found in that buffer. The expectation that it contains a UTF-16 string came from observing what gets passed by the boot managers I tested.

martinwhitaker:
Looks good to me.

debrouxl:
Likewise, as long as it's reasonable to assume that we don't need to bother supporting UTF-16BE.

tormodvolden:
> Likewise, as long as it's reasonable to assume that we don't need to bother supporting UTF-16BE.

Well, even if you would like to support the the 0.01% case of BE, wouldn't this commit be a step in the right direction and work fine for the 99.99% LE in the meantime? The commit doesn't worsen anything for LE.

tormodvolden:
I finally figured out how to add command line parameters directly, without writing them to a file first. So I updated the documentation accordingly.

Issue #425, opened 2024-08-04
Regression: wrong DDR speed / frequency after switching to -std=gnu11


01e3:
I noticed that git HEAD version of Memtest86+ shows the wrong DDR speed / frequency. For example, on my Q9650, it started reporting memory as DDR2-799 (with 399MHz) clock:

![obraz](https://github.com/user-attachments/assets/c7013558-f6be-41f6-b625-6626b11e1ada)

To compare, you can see in one of my older bugs how this looked until recently:
https://github.com/memtest86plus/memtest86plus/issues/125#issue-1315037162, note this is before https://github.com/memtest86plus/memtest86plus/commit/0f8981412c49c630c312a086687878c89275d0df, so the timings were bad back then.

I was able to identify the change that introduced the regression using git bisect: https://github.com/memtest86plus/memtest86plus/commit/53ca89f8aeeea649a1b8c2d2d14ec121f7025f54: "Add initial NUMA awareness support".

To be more specific, the regression is caused by:
```
-CFLAGS = -std=c11 -Wall -Wextra -Wshadow -m64 -march=x86-64 -mno-mmx -mno-sse -mno-sse2 \
+CFLAGS = -std=gnu11 -Wall -Wextra -Wshadow -m64 -march=x86-64 -mno-mmx -mno-sse -mno-sse2 \
```

I can also verify that it is enough to make the following change to fix the issue:
```
diff --git a/boot/macros.h b/boot/macros.h
index 4dd6787..99a0dbb 100644
--- a/boot/macros.h
+++ b/boot/macros.h
@@ -12,7 +12,8 @@

 #ifndef __ASSEMBLY__

-#ifdef __GNUC__
+//#ifdef __GNUC__
+#if 0
 // Enhanced definitions under GCC and compatible, e.g. Clang.

 // These are from GPLv2 Linux 6.7, for erroring out when the argument isn't an array type.
diff --git a/build32/Makefile b/build32/Makefile
index db86418..17e2ff8 100644
--- a/build32/Makefile
+++ b/build32/Makefile
@@ -10,7 +10,7 @@ else
   GIT_AVAILABLE = true
 endif

-CFLAGS = -std=gnu11 -Wall -Wextra -Wshadow -m32 -march=i586 -fpic -fno-builtin \
+CFLAGS = -std=c11 -Wall -Wextra -Wshadow -m32 -march=i586 -fpic -fno-builtin \
[...]

PR #426, opened 2024-08-04
Split SPD parsing and printing code from smbus.c to spd.c

01e3:
Move all SPD parsing and printing code from smbus.{c,h} to spd.{c,h}.

Introduce parse_spd() function, moving the parse_spd_* selection logic from print_smbus_startup_info(), allowing to keep parse_spd_* static.

Remove static from get_spd() and update print_smbus_startup_info() to use parse_spd() which also simplifies the code flow.

Move LINE_SPD into display.h and rename it to ROW_SPD. Update print_spdi() to use explicit row number where the SPD info needs to be printed.

Move ram_info and ram into display.{c,h} - this data is used for post_display_init(), so really belongs there.

Do not initialize ram.freq to 0, this is the initial value already.

Do not set curspd.isValid to False, the first thing that parse_spd() does is setting the entire struct to 0, that also sets isValid to False.

print_smbus_startup_info() from smbus.c is technically a skeleton now so each arch can have its own version, adjusted as needed.

I hope this will allow to keep working on https://github.com/memtest86plus/memtest86plus/pull/410 (such as renaming smbus.{c,h} into smbus_x86.{c,h} and creating new i2c_la64.{c,h} for LA64) and https://github.com/memtest86plus/memtest86plus/issues/425 without future conflicts. 

before:
```
$ make ; size -t */{smbus,display}.o ; stat -c "%n %s" memtest{.bin,.efi,_shared,_shared.bin}
  12352    1446      18   13816    35f8 system/smbus.o
   6959       4    1046    8009    1f49 app/display.o
  19311    1450    1064   21825    5541 (TOTALS)
memtest.bin 151896
memtest.efi 153088
memtest_shared 176120
memtest_shared.bin 150360

```

after:
```
$ make ; size -t */{smbus,spd,display}.o ; stat -c "%n %s" memtest{.bin,.efi,_shared,_shared.bin}
   2602       2      18    2622     a3e system/smbus.o
   9672    1420       0   11092    2b54 system/spd.o
   6969      28    1046    8043    1f6b app/display.o
  19243    1450    1064   21757    54fd (TOTALS)
memtest.bin 151896
memtest.efi 153088
memtest_shared 176216
memtest_shared.bin 150360
```


system/smbus.c
```diff
@@ -59,18 +43,11 @@ static uint8_t nf_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
 static uint8_t ali_m1563_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
 static uint8_t ali_m1543_read_spd_byte(uint8_t smbus_adr, uint8_t spd_adr);
 
-static inline uint8_t bcd_to_ui8(uint8_t bcd)
-{
-    return bcd - 6 * (bcd >> 4);
-}
-
 void print_smbus_startup_info(void)
```
debrouxl, line 46: `print_spd_startup_info()`, as suggested by Sam in #410 , so that the caller doesn't need to use preprocessor conditionals to select the function according to the ISA :)

system/spd.c
```diff
@@ -0,0 +1,978 @@
...
+#include "stdbool.h"
+#include "stdint.h"
+#include "string.h"
+
+#include "smbus.h"
+#include "spd.h"
+#include "jedec_id.h"
+
+// Functions Prototypes
+static void read_sku(char *sku, uint8_t slot_idx, uint16_t offset, uint8_t max_len);
```
debrouxl, line 15: I think that this refactor is a good opportunity to reorder functions in this file and get rid of these early prototype declarations.
Sam, Martin ?

debrouxl:
A couple nitpicks, but overall, we all seem to like the general direction of this refactor :)

system/smbus.c
01e3: Ah, sure, will do.

Overall, I tried to limit the scope of my changes, expecting it would be done as a part of https://github.com/memtest86plus/memtest86plus/pull/410, but happy to handle it here.

system/spd.c
01e3: Sure, done.

01e3:
Thank you for your quick review Lionel. I have implemented the changes as requested, waiting for Sam and Martin and then will push the updated version.

system/spd.c
```diff
@@ -0,0 +1,978 @@
+// SPDX-License-Identifier: GPL-2.0
+// Copyright (C) 2004-2023 Sam Demeulemeester
+
+#include "display.h"
+
```
martinwhitaker, line 5: My aim was to have a clean separation between the application code and the low-level libraries, with the former depending on the latter, but not vice-versa. So I don't like the inclusion of `display.h` here, and would like to move the display code into the `app` directory. But that can be another step if you don't want to do it now.

system/spd.h
```diff
@@ -0,0 +1,47 @@
...
+
+/** Rounding factors for timing computation
+ *
+ *  These factors are used as a configurable CEIL() function
+ *  to get the upper int from a float past a specific decimal point.
+ */
+
+#define DDR5_ROUNDING_FACTOR    30
+#define ROUNDING_FACTOR         0.9f
+
```
martinwhitaker, line 21: These two definitions are only used inside `spd.c` so are not part of the unit API and should be moved into `spd.c`.

app/display.c
```diff
@@ -86,6 +86,8 @@ int max_cpu_temp = 0;
 
 display_mode_t display_mode = DISPLAY_MODE_NA;
 
+ram_info ram = { 0, 0, 0, 0, 0, 0, "N/A"};
+
```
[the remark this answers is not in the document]
01e3: Sure, moved "back", i.e. into sdp.c

app/display.h
```diff
@@ -39,6 +41,18 @@ typedef enum {
...
+    uint16_t    tCL;
+    uint8_t     tCL_dec;
+    uint16_t    tRCD;
+    uint16_t    tRP;
+    uint16_t    tRAS;
+    char        *type;
+} ram_info;
+
+extern ram_info ram;
+
```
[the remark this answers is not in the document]
01e3: Done & done.

system/spd.c
01e3: Understood. We can limit this to` #include "print.h"` for now. Longer term, perhaps both print_spd_startup_info() and print_spdi() should go to a new place. Once LA64 changes land, I will think about a good way to do it.

system/spd.h
01e3: Good idea, done.

01e3:
Thank you for your quick review Martin!

I just pushed a new version of the commit that includes all requested changes.

system/spd.c
```diff
@@ -0,0 +1,977 @@
...
+#define ROUNDING_FACTOR         0.9f
+
+ram_info_t ram = { 0, 0, 0, 0, 0, 0, "N/A"};
+
+static inline uint8_t bcd_to_ui8(uint8_t bcd)
+{
+    return bcd - 6 * (bcd >> 4);
+}
+
+void print_spdi(spd_info spdi, uint8_t row)
```
x86fr, line 29: Agree with Martin that `print_spdi` should be moved to app ultimatly, but that could be done later.

build64/Makefile
x86fr, line 1: Would use this PR to add `-fexcess-precision=standard` in CFLAGS (same in build32) to avoid potential issues here and there (no matter if these computations are switched to int later).

x86fr (commented):
Split looks OK. Will merge it asap after CFLAGS addition.

system/spd.c
01e3: OK, will handle this post LA64 merge.

build64/Makefile
01e3: Thank you. I can also add another commit for this if it would make your life easier, please let me know?

Also, I have the float-to-int commit almost ready, so as soon as you merge I should be able to send another PR. :)

x86fr: No more PR merged after this one until #410 is merged. It adds unnecessary work to Loongson team. We'll deal with minor PRs afterward.

01e3: Understood, thanks!

debrouxl:
LGTM, ready to merge after the CFLAGS addition requested by Sam in both x86 Makefiles.

01e3:
> LGTM, ready to merge after the CFLAGS addition requested by Sam in both x86 Makefiles.

Thank you. To clarify, should I add a commit for *this single change*?

debrouxl:
I'd say yes; then, it will be up to Sam or whoever merges this PR to decide whether to use Rebase and merge, or Squash and merge :)

Issue #265, opened 2023-02-10
Add a .osrel section to EFI image


medhefgo:
If the EFI image had a `.osrel` section, boot loaders that support the [Boot Loader Spec](https://uapi-group.org/specifications/specs/boot_loader_specification/) like systemd-boot can automatically find and add an entry to memtest if it is located under `$BOOT\EFI\Linux\`.

The contents of the section would be as minimal as this (simple ASCII/UTF-8 encoding):
```
ID=memtest86plus
NAME=memtest86+
VERSION=6.10
```

debrouxl:
Technically, the implementation seems to be straightforward. With this ^ patch, objdump says that .osrel appears in the binary, with a size of 0x0200 bytes like the other additional sections.

I'd say that we need further input before merging something like that, though :)
This grows the size of the header's contents again. What will happen when there are enough sections not to fit in the first 497 bytes of the header ? Put another way: is this going to hamper future evolution or compatibility with something ?
In any case, binaries containing a new section of short arbitrary name warrant compatibility testing with UEFI BIOS and other bootloaders or parsers.

debrouxl:
@medhefgo Do you have any answers to the questions in my previous message ? :)

medhefgo:
> This grows the size of the header's contents again. What will happen when there are enough sections not to fit in the first 497 bytes of the header ? Put another way: is this going to hamper future evolution or compatibility with something ?

You can always resort to hiding the x86 boot protocol header data in the DOS area because the PE header offset can be chosen freely: https://github.com/memtest86plus/memtest86plus/pull/136#issuecomment-1501057998

medhefgo:
Also, my original osrel definition is slightly against spec/recommendations. The `NAME` and `VERSION` values should be quoted:
```
ID=memtest86plus
NAME="memtest86+"
VERSION="6.10"
```

debrouxl:
Now that #427 is merged, I could resume work on this and send a PR, on another day.

PR #427, opened 2024-08-07
Merge bootsect.S and setup.S into header.S

01e3:
_Note: This is a slightly different approach for implementing the idea from RFC PR https://github.com/memtest86plus/memtest86plus/pull/136, but this time without the addition of serious code-size restrictions and the requirement of counting every byte at each modification, which was a show stopper for the original attempt._ 

Currently, Memtest86+ provides two (three counting mbr.S) different headers: bootsect.S and header.S that together with setup.S are used to build the final binaries:

Both support loading by an intermediate bootloader. Also, both are derived from the Linux kernel, but from a very different chapter in its evolution. FDD loading was removed a very long time ago, many, many years before adding PE/COFF headers for UEFI. Also, the PE/COFF headers are large (even larger for x86-64) and take a lot of very limited space, leaving very almost no headroom for the actual code.

This commit merges bootsect.S and setup.S into header.S, adding BIOS (FDD) loading support, which should eventually enable to maintain a single, all-in-one binary that can be booted either by an intermediate bootloader (such as GRUB or LILO), UEFI and BIOS. Easier and less confusing for users, as I have continued to see cases when people assume memtest.bin is needed for non-UEFI systems, even if never intended for FDD loading.

To ensure we have enough space for both the boot code (limited to 512B including Linux boot header and and address of the New EXE header) and UEFI headers, the PE/COFF headers got moved into the ".setup" section of the code where *currently* we have 1024B total and about 90B free. Should more space be needed in the future, we can grow that section or perhaps even better - move PE/COFF into its own section that will not even need to be loaded, as also suggested in https://github.com/memtest86plus/memtest86plus/pull/136#issuecomment-1501057998

Overall, the code got copied from bootsect.S and setup.S with very, very minimal changes, most importantly:

 - Dedicated routine for printing strings (print_string) got added and lives in the ".bootsect" section.

 - Print "Loading " message almost immediately after BIOS loads and calls code from the .bootsect" section, even before setting up the custom FDD parameter table. The sooner the better.

 - Set DX to 0x0, as the code depends on its value. While the BIOS is expected to set DL register to the boot drive number which is 0 for FDD, to my knowledge there is no guarnatee for DH to have any particular value. Also, when executing "MZ" signature as x86 instructions, the original value of DX becomes instantly and irrevocably lost. This is fine as we only support booting from FDD anyway.

 - Print "Memtest86+" add its version after loading the ".setup" section. We get the string from mt86plus_version (kernel_version), no need to duplicate in the boot sector where we are very space limited and as a bonus we can provide the version information.

obraz

As diffstat shows a lot of code changed and added, it may be easier use the following commands:
 diff -Nur bootsect.S header.S
 diff -Nur setup.S header.S

While there are several more changes and fixes I would like to make in the boot code (some of them were already included in the earlier RFC version [1] before we got UEFI signing support added) for now I tried this to be as least disruptive as possible and avoid unnecessary code changes to make the review easier.

Also, I kept memtest.bin alive for now, in case we would still like to provide it in the next release. Let me know if killing it earlier would make more sense.

[1] https://github.com/memtest86plus/memtest86plus/pull/136


boot/header.S
```diff
@@ -1,11 +1,31 @@
...
+// 2) To support loading from the UEFI, it provides the MS-DOS & PE/COFF headers.
+// The PE/COFF headers point to offset 0x10000 + 0x1E0 where efi_boot from
+// startup32.S / startup64.S that handles EFI handover and the rest of booting.
+//
+// 3) When using an intermediate bootloader, it provides the first few bytes of
+// the Linux boot header (at the end of the ".bootsect" section), with the
+// remainder of the header and the setup code being provided in the ".setup".
+// section.
 //
 // Copyright (C) 2020-2023 Martin Whitaker.
```
martinwhitaker, line 30: Update copyright to 20234

boot/header.S
```diff
@@ -21,63 +41,750 @@
...
-	# int 0x19 should never return. In case it does, invoke the BIOS.
-        # reset code.
-	ljmp	$0xf000,$0xfff0
+	boot_msg:
+	.ascii	"Loading "
+	boot_msg_end:
 
-        # The PE header pointer.
 	.org	0x3c
+	e_lfanew:
```
martinwhitaker, line 101: Is this label used anywhere?

boot/header.S
```diff
@@ -21,63 +41,750 @@
...
+	int	$0x10
+	ret
+
+# Local variables.
+
+sectors:
+	.word	0
+
+
+# Emulate the Linux boot header, to allow loading by intermediate boot loaders.
```
martinwhitaker, line 406: Include the couple of extra comment lines from setup.S, suitably modified.

martinwhitaker (commented):
OK, I like it. A few minor quibbles and I haven't tested it yet, but looks good to me. I think the phrasing and grammar in the initial comments could be improved, but I can do that after this is merged.

I don't see any reason to keep memtest.bin. I suggest we name the all-in-one binary just `memtest`. The README will also need to be updated,

boot/header.S
01e3: Done

boot/header.S
01e3: Not currently - it is a leftover from my objdump debugging. Deleted, unless you prefer for example to add an assert for it?

boot/header.S
01e3: Done.

boot/header.S
martinwhitaker: The .org should fail if there's a problem, so I don't think we need an assert for this.

01e3:
Thanks for your quick review Martin!

> OK, I like it. A few minor quibbles and I haven't tested it yet, but looks good to me. I think the phrasing and grammar in the initial comments could be improved, but I can do that after this is merged.

Happy to make some changes now if you flag the problematic parts, but fixing it later is also fine.

> I don't see any reason to keep memtest.bin. I suggest we name the all-in-one binary just `memtest`. The README will also need to be updated,

Would you prefer that I handle all of this in a single commit, or rather add a separate commit for updating the Makefiles, README etc?

boot/header.S
01e3: Actually https://github.com/memtest86plus/memtest86plus/commit/e38ad8e6a406ef7a5fd6cdfcbf2c4ab5859cb2af already took care of this?

martinwhitaker:
> Would you prefer that I handle all of this in a single commit, or rather add a separate commit for updating the Makefiles, README etc?

I favour multiple commits.

boot/header.S
martinwhitaker: Ah, good point!

debrouxl:
I like it as well. It's good to see the limitation of the previous version fixed :)
It shall unblock the addition of a `.osrel` section (if we merge #265 ), a `.data` section if we eventually need one (say, because "secure" boot signing requirements end up mandating that somehow even for special-purpose kernels), etc.

I agree with Martin about the naming of the unified binary, in that it would seem easier to me to explain that `memtest` works for both BIOS and UEFI loading, than to explain that either `memtest.bin` or `memtest.efi` works for both (probably yielding "where is the UEFI version?" and "where is the BIOS version?" questions, respectively).
Or maybe the build system should simply produce all three of `memtest`, `memtest.bin` and `memtest.efi`, the latter two being `cp`'ed from the former, and we would call it a day ?

01e3:
Yes, we can add more sections, at least for now. If one day we need even more space, we can just move PE/COFF into its own section that will not even need to be loaded.

Sorry, I missed the request to change the name to just "memtest". Will update the last patch, unless "memetest.efi" is acceptable, please let me know? Overall, I think ".efi" extension as the default makes more sense given that it is needed for loading from UEFI and for other use-cases (dd to /dev/fdd or loading from non-UEFI loaders like LILO) the name does not matter. That said, I don't have a strong opinion here. 

Regarding cp to produce three different names - I'm concerned it is going to create even more confusion that these are three different files. I think one is just enough and people (distributions) that later compile it, will name it accordingly.

x86fr:
Name it `mt86plus` to avoid confusion with other "memtest". 

Also, this really deserve some serious testing with all intermediate boot loaders (syslinux, iPXE, shim, etc.) to avoid potential compatibility regression.

01e3:
> Name it `mt86plus` to avoid confusion with other "memtest".

Sure. mt86plus or mt86plus.efi?

> Also, this really deserve some serious testing with all intermediate boot loaders (syslinux, iPXE, shim, etc.) to avoid potential compatibility regression.

Agreed. Just a note that I did some testing on my own such as floppy, BIOS CD (w/ floppy emulation), UEFI, iPXE, GRUB, this is of course far from 100% coverage.

01e3:
Alright, I think I updated all references to memtest.efi with mt86plus.efi. Will need to do the same for build64/la64/Makefile but I would prefer to handle this as a different PR.

martinwhitaker:
I think using the .efi extension will continue to cause confusion amongst users, who will just assume we've dropped support for legacy boot.

debrouxl:
Agreed.

01e3:
Sure, no problem - I can use "mt86plus".

But.... my point is that for BIOS loading (FDD or CD with emulation) the extension clearly does not matter, where for UEFI we need *.efi. Because of this, the first thing all distros will do is to rename it to [hopefully] mt86plus.efi. For "we've dropped support for legacy boot" my hope was that it could be addressed by the documentation, which we could augment even more if what we have right now is not sufficient.

Anyway, will update.

debrouxl:
Ah right, on the one hand, we need *.efi... but on the other hand, I'm afraid that no amount of documentation will be able to prevent user confusion about memtest86+ dropping BIOS support because the binary name has become `mt86plus.efi`.

Maybe my previous suggestion of `cp`'ing files wasn't bad, though 3 files is probably a bit much if 2 files do the job :)
`ln mt86plus.efi mt86plus` in the Makefiles would make it even clearer than `cp mt86plus.efi mt86plus` that the files are the same.
Sam, Martin ?

martinwhitaker:
In general the `.efi` extension is not required - all bootmanagers/bootloaders I know about (rEFInd, GRUB, systemd-bootd, Tiano Core EFI shell) can happily load a file without that extension. The GRUB ISO built by our Makefile renames the binary to simply `memtest`. The .efi extension is only needed for the falback bootloader, which must be named `boot{arch}.efi` and stored in `\EFI\BOOT`.

In any case, except when building ISOs, our Makefile just builds the binary. It doesn't install it in the ESP. So whatever performs the installation can add the .efi extension at the same time if it wants to.

x86fr:
Agree. `mt86plus` without extension is neutral and probably the way to go.

01e3:
Right, no issues with the bootmanagers/bootloaders as far as I'm concerned. However, in most of my system I have memtest added as a boot option and UEFI only allows to select files with *.efi extension.

That said, I have made the change as requested - I only kept mt86plus.efi in debug_memtest.sh, unless you also want to change that one?

01e3:
Hi there. What would be the next step here to get this merged?

debrouxl:
I'd say copious amounts of beta-testing on many computer models, as was performed for e.g. #34 , only wider as both UEFI and legacy BIOS modes need to be tested on a couple decades of motherboards. Legacy BIOS mode should be undisturbed by this, but as we know, BIOS/UEFI implementations on many MBs have *quirks*.

Sam ?

martinwhitaker:
Sam normally does the copious amounts of beta testing before making a release, so I don't see any reason not to merge this now. @x86fr , any objection?

x86fr:
Please wait a bit (max 2 weeks). A release is almost ready and tested, then we can merge this one and start a new round of tests with all these PRs included.

x86fr:
Seems there is a conflict on `doc/HOW_TO_DEBUG_WITH_GDB.md` that online github crash while trying to resolve... 

@debrouxl > as a github guru, can you solve it using command line? 😉

debrouxl:
I can apply the patches into a new branch for this repo, in an author-respecting way... but I'm not sure how to shoehorn the end result into this PR in such a way as not to screw up the state of this PR :)

@01e3 could you perform that rebase on your side ? TIA.

01e3:
I'm sorry, I created some mess and github does not like what I did here and does not allow to reopen this PR. So, I ended up opening a new one: https://github.com/memtest86plus/memtest86plus/pull/467

:/

debrouxl:
No problem, don't worry :)

PR #429, opened 2024-08-09
Change SPD math from floating point to integer (fixed point)

01e3:
There is really no need to use floating point for calculating freq and timings since there are no floats in the spd_infos struct. Doing so reduces the code size and protects us from bugs like the one discovered in https://github.com/memtest86plus/memtest86plus/issues/425.

Summary of the changes:
 - introduce ceildiv() function to perform int division with ceiling.

 - ddr5: trivial fix, spdi->freq can be easily calculated without floats just by changin the order of operations.

 - ddr4: calculate tCK/tns as 10^-3 (mili) int. Use ceildiv() instead of adding ROUNDING_FACTOR. Simplify spdi->freq rouding and explain the logic used there.

 - ddr3: calculate mtb/tCK/tns as 10^-3 (mili) int. Use ceildiv() instead of adding ROUNDING_FACTOR.

 - ddr2: calculate tckns/tns as 10^-3 (mili) int. Use ceildiv() instead of adding ROUNDING_FACTOR. Use "/ 2" instead of "* 0.5" for calculating spdi->tCL.

 - sdram: calculate tckns/tns as 10^-3 (mili) int. Use ceildiv() instead of adding ROUNDING_FACTOR.

 - remove ROUNDING_FACTOR - no longer used.

On top of the above changes, provide min() and max() macros in macros.h and move boot/macros.h into lib/macros.h.

```
Before:
  13350     868       0   14218    378a system/spd.o (x86-32)
  10279    1444       0   11723    2dcb system/spd.o (x86-64)

After:
  10277     868       0   11145    2b89 system/spd.o (x86-32)
   7770    1444       0    9214    23fe system/spd.o (x86-64)
```

### DIFF
--- a/lib/macros.h
+++ b/lib/macros.h
@@ -20,9 +20,23 @@
 #define __same_type(a, b)    __builtin_types_compatible_p(typeof(a), typeof(b))
 #define __must_be_array(a)   BUILD_BUG_ON_ZERO(__same_type((a), &(a)[0]))
 #define ARRAY_SIZE(arr)      (sizeof(arr) / sizeof((arr)[0]) + __must_be_array(arr))
+
+#define min(x, y) ({                        \
+    typeof(x) _min1 = (x);                  \
+    typeof(y) _min2 = (y);                  \
+    (void) (&_min1 == &_min2);              \
+    _min1 < _min2 ? _min1 : _min2; })
+
+#define max(x, y) ({                        \
+    typeof(x) _max1 = (x);                  \
+    typeof(y) _max2 = (y);                  \
+    (void) (&_max1 == &_max2);              \
+    _max1 > _max2 ? _max1 : _max2; })
 #else
 // Fallback definitions.
 #define ARRAY_SIZE(var_) (sizeof(var_) / sizeof((var_)[0]))
+#define min(x, y) (x < y ? x : y)
+#define max(x, y) (x > y ? x : y)
 #endif
 
 #endif
--- a/system/spd.c
+++ b/system/spd.c
@@ -5,6 +5,7 @@
 #include "stdint.h"
 #include "string.h"
 
+#include "macros.h"
 #include "smbus.h"
 #include "spd.h"
 #include "jedec_id.h"
@@ -17,9 +18,12 @@
  */
 
 #define DDR5_ROUNDING_FACTOR    30
-#define ROUNDING_FACTOR         0.9f
 
 ram_info_t ram = { 0, 0, 0, 0, 0, 0, "N/A"};
+
+static int ceildiv(int a, int b) {
+    return (a + b - 1) / b;
+}
 
 static inline uint8_t bcd_to_ui8(uint8_t bcd)
 {
@@ -221,7 +225,7 @@ static void parse_spd_ddr5(spd_info *spdi, uint8_t slot_idx)
         return;
     }
 
-    spdi->freq = (float)(1.0f / tCK * 2.0f * 1000.0f * 1000.0f);
+    spdi->freq = 2 * 1000 * 1000 / tCK;
     spdi->freq = (spdi->freq + 50) / 100 * 100;
 
     // Module Timings
@@ -315,34 +319,28 @@ static void parse_spd_ddr4(spd_info *spdi, uint8_t slot_idx)
     spdi->hasECC = (((get_spd(slot_idx, 13) >> 3) & 1) == 1);
 
     // Module max clock
-    float tns, tCK, ramfreq, fround;
+    int tns, tCK, ramfreq;
 
     if (get_spd(slot_idx, 384) == 0x0C && get_spd(slot_idx, 385) == 0x4A) {
         // Max XMP
-        tCK = (uint8_t)get_spd(slot_idx, 396) * 0.125f +
-              (int8_t)get_spd(slot_idx, 431)  * 0.001f;
+        tCK = (uint8_t)get_spd(slot_idx, 396) * 125 +
+              (int8_t)get_spd(slot_idx, 431);
 
         spdi->XMP = 2;
 
     } else {
         // Max JEDEC
-        tCK = (uint8_t)get_spd(slot_idx, 18) * 0.125f +
-              (int8_t)get_spd(slot_idx, 125) * 0.001f;
+        tCK = (uint8_t)get_spd(slot_idx, 18) * 125 +
+              (int8_t)get_spd(slot_idx, 125);
     }
 
-    ramfreq = 1.0f / tCK * 2.0f * 1000.0f;
+    ramfreq = 2 * 1000 * 1000 / tCK;
 
-    // Round DRAM Freq to nearest x00/x33/x66
-    fround = ((int)(ramfreq * 0.01 + .5) / 0.01) - ramfreq;
-    ramfreq += fround;
-
-    if (fround < -16.5) {
-        ramfreq += 33;
-    } else if (fround > 16.5) {
-        ramfreq -= 34;
-    }
-
-    spdi->freq = ramfreq;
+    // Round DRAM Freq to the nearest x00/x33/x66:
+    // xx00..xx16 -> xx00, xx17..xx49 -> xx33, xx50..xx82 -> xx66, xx84..xx99 -> x[x+1]00
+    // special case: xx83 -> xx66  - via min(x, 2) to avoid xx99
+    ramfreq += 16;
+    spdi->freq = 100 * (ramfreq / 100) + min((ramfreq % 100) / 33, 2) * 33;
 
     // Module Timings
     if (spdi->XMP == 2) {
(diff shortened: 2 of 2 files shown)


x86fr:
Unfortunately, it seems there is many cases where the ceildiv function has a different result than the previous cast + float rounding_factor, leading to timings errors (usually adding +1 to the correct result). I'll double-checl and try to post some example asap.

01e3:
Nothing is easy. :/

Sure, we should be able to restore the ROUNDING_FACTOR behavior, but if I can get some sanitized .spd files, it would be great to make sure that's the problem. There is no checksum versification in Memtest86+ for SPD, so simply replacing SN with 0 should do, as long as this is the only requirement. If you want, I can quickly write something, so you don't have to spend too much time on this?

01e3:
[spd-sanitize.txt](https://github.com/user-attachments/files/16591644/spd-sanitize.txt)

github does not allow to upload .c files, so I had to rename it to .txt

Please double check the offsets. Also, I did not know what to set for RAMBUS (also 95?) but it does not matter as I did not touch that part of the code, so we don't need to test it.

x86fr:
You can easily simulate all possible scenarios with the following code. Currently, there are quite a few cases where it fails (approximately 5 to 10%). While real-world diversity isn't as extensive, there are still many obscure SPD configurations where failures will occur. The current code has been fine-tuned to achieve an almost perfect match after significant effort, so any "cosmetic" changes must maintain the same level of accuracy.

```
#include <stdio.h>
#include <stdint.h>
#include <time.h>
#include <stdlib.h>

#define MIN(x, y) (((x) < (y)) ? (x) : (y))

static int ceildiv(int a, int b) {
    return (a + b - 1) / b;
}

int main()
{
    float tCKf, tnsf, tCLf;
    int tCKi, tCLi, tnsi;
    uint16_t tCL;

    int val1, val2, val3, val4;

    int i = 0;

    srand(time(NULL));

    while(1) {
        val1 = rand() % 256;
        val2 = rand() % 256;
        val3 = rand() % 256;
        val4 = rand() % 256;

        tCKf = val1 * 0.125f + val2  * 0.001f;
        tnsf = val3 * 0.125f + val4  * 0.001f;
        tCLf = (uint16_t)(tnsf/tCKf + 0.9f);

        if(tCLf < 2) continue; 

        tCKi = val1 * 125 + val2;
        tnsi = val3 * 125 + val4;
        tCLi = ceildiv(tnsi, tCKi);

        if((int)tCLf != tCLi) {
           printf("tCLf: %i - tCLi: %i --> FAILED! (%i, %i, %i, %i)\r\n", (int)tCLf, tCLi, val1, val2, val3, val4);
        } else {
           printf("tCLf: %i - tCLi: %i\r\n", (int)tCLf, tCLi);
        }

        i++;
        if(i==250) break;
    }

    return 0;
}
```

01e3:
Alright, let's get back to this. First, thank you so much Sam for the details and for the PoC.

Indeed, ceildiv() as currently implemented does produce different results because, perhaps somehow naively, I tried to use different logic than the current "fine-tuned"  one.

To avoid wasting more of your time, I can implement the same "+0.9" logic with integers which hopefully should be "close enough", but I want to make it clear that getting 100% match is not going to be possible due to a tiny precision difference.

Also, it is not even obviously clear to me what is the current "golden" reference. Using your slightly modified PoC:

```
#include <stdio.h>
#include <stdint.h>
#include <time.h>
#include <stdlib.h>

static int ceildiv(int a, int b) {
    return (a + b - 1) / b;
}

static int rounddiv(int a, int b) {
    return (a + b * 9 / 10) / b;
}

int main(int argc, char **argv)
{
    float tCKf, tnsf, tCLf;
    int tCKi, tCLic, tCLir, tnsi;
    int val1, val2, val3, val4;

    val1 = atoi(argv[1]);
    val2 = atoi(argv[2]);
    val3 = atoi(argv[3]);
    val4 = atoi(argv[4]);

    tCKf = val1 * 0.125f + val2  * 0.001f;
    tnsf = val3 * 0.125f + val4  * 0.001f;
    tCLf = (tnsf/tCKf + 0.9f);

    tCKi = val1 * 125 + val2;
    tnsi = val3 * 125 + val4;

    tCLic = ceildiv(tnsi, tCKi);
    tCLir = rounddiv(tnsi, tCKi);

    printf("(%i %i %i %i -> %i / %i): tCLf(%.10f)=%i, tCLi_ceil=%i, tCLi_round=%i: %s\r\n",
           val1, val2, val3, val4,
           tnsi, tCKi,
           tCLf, (int)tCLf, tCLic, tCLir,
           ((int)tCLf != tCLir) ? "Problem!" : "OK");

    return 0;
}

```
$ cc -Wall poc.c -o poc -mfpmath=387 && ./poc 53 107 216 123 && ./poc 69 145 146 167 && ./poc 4 120 39 147
(53 107 216 123 -> 27123 / 6732): tCLf(4.9289660454)=4, tCLi_ceil=5, tCLi_round=4: OK
(69 145 146 167 -> 18417 / 8770): tCLf(2.9999997616)=2, tCLi_ceil=3, tCLi_round=3: Problem!
(4 120 39 147 -> 5022 / 620): tCLf(9.0000000000)=9, tCLi_ceil=9, tCLi_round=9: OK

$ cc -Wall poc.c -o poc -mfpmath=sse && ./poc 53 107 216 123 && ./poc 69 145 146 167 && ./poc 4 120 39 147
(53 107 216 123 -> 27123 / 6732): tCLf(4.9289660454)=4, tCLi_ceil=5, tCLi_round=4: OK
(69 145 146 167 -> 18417 / 8770): tCLf(3.0000000000)=3, tCLi_ceil=3, tCLi_round=3: OK
(4 120 39 147 -> 5022 / 620): tCLf(8.9999990463)=8, tCLi_ceil=9, tCLi_round=9: Problem!

As you can see, we are not only get different results for -mfpmath=sse vs -mfpmath=387 but also run into issues like
2.1 + 0.9 = 2 for i387 or 8.1 + 0.9 = 8 for SSE.

Overall, rounddiv() is much closer to the current floating-point code, but is it close enough to satisfy the "fine-tuning"?

BTW: while I assume Memtest86+ currently uses i387 floating-point on both i386 and x86-64 (due to -mno-sse) so for both arch we should get the same values, I suspect things may behave different on non-x86 architectures, like for example on the recently added LA64.

x86fr:
To summarize my thoughts: the current timing computation code has been fine-tuned over years and is now stable, as we haven’t received new bug reports recently—aside from cases where we changed flags. Admittedly, the code has edge cases involving float rounding that depend on both architecture and compiler flags, but it works reliably as is.

If there’s a way to achieve the same behavior using only integer calculations, that could be worth considering. However, starting a lengthy testing phase from scratch to stabilize a new parsing implementation—just to save a few hundred bytes—doesn't seem worthwhile from what I understand.

01e3:
To recap from the discussion i another PR:

- There are still 586-class system with no FPU, like Vortex86SX. so trying to parse SPD would lead into a crash in case we have  working I2C...

On top of this, we have also started supporting different architectures (hello LA64) potentially with different float precision and behavior. Even i387 vs SSE behaves differently, so could end up with different results, as I flagged here, with some examples.

I believe changing ceildiv() back to rounddiv() will make the int code close to 100%-compatible with with the current floating point one, for the remaining ones I don't believe we even understand what the right results is, as it depends on instruction-set.

Should I do this and we can give it another try, or should I close the PR?

01e3:
> If you can get the exact same results without using floats with the code at [#429 (comment)](https://github.com/memtest86plus/memtest86plus/pull/429#issuecomment-2324226338) (with current Memtest86+ compiler flags), that's great and it can be merged.

No, I don't believe I can do that, unfortunately.

For just completely random values, for which I would even guess many of them cannot exist "in real life", I cannot get 100% the same result - as discussed earlier. 

At the same time - we cannot even get the same result between FPU and SSE, so the current code is only guaranteed to provide the "blessed" values when compiled for i387 (and if specific gcc flags are being used - as seen earlier). I would not be surprised if LA64 produces even another set of results.

So, while trying to go with ceildiv() was a little bit... welll. naive on my part, roundiv() should be much closer compared to the current code, and depending on the set of variables - behave either like SSE or FPU, where ceildiv() had its own... deviation...

Anyway, given that I'm repeating the same arguments, I think it is time to close this PR and move on....

x86fr:
There is no black magic on the current code, there is just a compiler flag to tell the compiler how to deal with excess precision. 

With the proper flag set, the result is the same with i387, SSE or x64. Relying on the default value of `-fexcess-precision` (which changed from `standard` to `fast`) was the cause of the issue we had with rounding. Now this flag is set explicitly to avoid this.

01e3:
I don't believe this is correct. Even with -fexcess-precision=standard we get different results for -mfpmath=387 and -mfpmath=sse and standard / fast does not change this:

# cc -fexcess-precision=standard -Wall poc.c -o poc -mfpmath=387 && ./poc 4 120 39 147 ; cc -fexcess-precision=standard -Wall poc.c -o poc -mfpmath=sse  && ./poc 4 120 39 147
(4 120 39 147 -> 5022 / 620): tCLf(9.0000000000)=9, tCLi_ceil=9, tCLi_round=9: OK
(4 120 39 147 -> 5022 / 620): tCLf(8.9999990463)=8, tCLi_ceil=9, tCLi_round=9: Problem!

cc -fexcess-precision=fast -Wall poc.c -o poc -mfpmath=387 && ./poc 4 120 39 147 ; cc -fexcess-precision=fast -Wall poc.c -o poc -mfpmath=sse  && ./poc 4 120 39 147
(4 120 39 147 -> 5022 / 620): tCLf(9.0000000000)=9, tCLi_ceil=9, tCLi_round=9: OK
(4 120 39 147 -> 5022 / 620): tCLf(8.9999990463)=8, tCLi_ceil=9, tCLi_round=9: Problem!

Perhaps there is a bug in the poc code, but I don't see it... That said, I already close the PR.

Issue #403, opened 2024-05-10
Init process in SMP PAR mode is very slow


kilaterlee:
Hi,
I have a problem that I found it is very slow in init process when in SMP(PAR) mode with 256G memory 64 cores, I checked the code, it seems each core will runs the sample test twice, and each sample case iterates through all of the address windows. 
So if it has 64 core-256G memory, I test it all night and still didn't pass the init process.
Has anyone encountered the same problem?

debrouxl:
My computers with 64 cores and 256 GB of RAM normally spend several seconds in the init phase. All night without proceeding to test #0 shouldn't happen, though.
First of all: which version of memtest86+ are you using: 7.00 release, nightly version, something else ?

kilaterlee:
> My computers with 64 cores and 256 GB of RAM normally spend several seconds in the init phase. All night without proceeding to test #0 shouldn't happen, though. First of all: which version of memtest86+ are you using: 7.00 release, nightly version, something else ? Also, could you give some details about your platform: for instance, processor model, number of sockets, possibly commercial computer model name ?

Thanks for you reply, I'm using the code after V7 and the comiit id is ded371e9dafcc.
However, I'm using a non-x64 platform to test with SMP PAR mode. Currently, I know that the memtest86+ does not allow other patform testing, but I'm one of LoongArch maintainers in edk2 and edk2-platfroms, I'm also work at Loongson, so it is easier for me to add the LoongArch suppot into memtest86+, and I promise the code added just for LoongArch supports it will not affect other test logic.
This a other topic that whether memtest86+ going to accommodate more ARCH or whether add multi-ARCH support. I hope it can become possible after my patch, I think I took into account the rules the multi-ARCH in memtest86+, and the architecture related code should be independent from the test logic code.
That should be:
1. All of the API or UI related should be public, and everyone must follow the public API, such as cpu info, clock, IMC info, flush cache, AMO function, etc.
2. ARCH init and the ARCH related codes should be private and independent.
3. The test logic is public and does not depend on any feature or ARCH. It is looking good now.

After all, the memtest86+ can made more powerful and can accommodate more ARCH. If I submit the reivew PRs, would you be interested in reviewing them?

x86fr:
Yes! Also, we'll be very interested to add Longson Arch on all the detection/ecc polling code. Are datasheets public?

martinwhitaker:
Does the boot option `powersave=off` make any difference?

[...]

PR #430, opened 2024-08-11
x86_64: preserve r8-11 over call to interrupt handler.

martinwhitaker:
As noted in https://github.com/memtest86plus/memtest86plus/issues/403#issuecomment-2206973854, the x86_64 SysV calling convention does not require r8-r11 to be preserved by the callee. So we need to save and restore them in the low-level interrupt handling code in case the C interrupt() function uses them.

### DIFF
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-2.0
-// Copyright (C) 2020-2022 Martin Whitaker.
+// Copyright (C) 2020-2024 Martin Whitaker.
 //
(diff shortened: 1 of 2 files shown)


debrouxl:
That's the straightforward, rock-solid solution to the matter of saving and restoring these registers: no need to depend on compiler flags to force it either not to use them, or save+restore the subset of r8-r11 it feels like using, in that function.

Being x86-specific, this doesn't have to be blocked on merging #410 . We can rename `app/interrupt.c` to `app/x86/interrupt.c` later.

PR #435, opened 2024-09-05
Add faulty memory range display mode

dvdkon:
I've added a new error reporting mode to roughly show which ranges of memory are affected. It's pretty similar to the BadRAM mode, except more "dumb": Instead of showing an address and mask, it just shows you a number of (address, range size) pairs covering all reported errors. This has lower information density, but this format is readily usable with a [stock Linux kernel](https://unix.stackexchange.com/a/76188) or maybe even [Windows](https://www.memtest86.com/blacklist-ram-badram-badmemorylist.html). The BadRAM patches are neat, but they seem to only be available [for very old Linux versions](http://rick.vanrein.org/linux/badram/download.html).

I tested it in QEMU by manually calling `data_error()`. Sadly I couldn't test it on real hardware, because the machine I developed this for seems to have magically healed its memory errors. Go figure.

Comments and/or suggestions are welcome.

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -223,6 +223,8 @@ static void parse_option(const char *option, const char *params)
             error_mode = ERROR_MODE_ADDRESS;
         } else if (strncmp(params, "badram", 7) == 0) {
             error_mode = ERROR_MODE_BADRAM;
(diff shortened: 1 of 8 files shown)


x86fr:
Added in #454

Issue #438, opened 2024-09-18
Feature request: AMD k10 based APU temperature


hap2001gh:
I cannot see the CPU temperature on my Stoney Ridge laptop, I hope this feature can be added in future versions.

jonathan-teh:
Stoney Ridge is Excavator, which does have a different method of reading the temperature compared to everything else in family 10h-16h. The mentioned PR should fix it.

x86fr:
Should be fixed by #440

PR #440, opened 2024-09-26
temperature: Add support for AMD Excavator

jonathan-teh:
Read current temperature from SMU via root complex.

### DIFF
--- a/system/temperature.c
+++ b/system/temperature.c
@@ -103,6 +103,14 @@ int get_cpu_temperature(void)
 
             return cpu_temp_offset + 0.125f * (float)((regl >> 21) & 0x7FF);
 
+        } else if (cpuid_info.version.extendedFamily == 6 && (cpuid_info.version.extendedModel == 6 || cpuid_info.version.extendedModel == 7)) { // Target family 15h (Excavator)
+
+            pci_config_write32(0, 0, 0, AMD_SMU_INDEX_ADDR_REG, AMD_F15_M60H_TEMP_CTRL_OFFSET);
+            regl = pci_config_read32(0, 0, 0, AMD_SMU_INDEX_DATA_REG);
+            int raw_temp = ((regl >> 21) & 0x7FF) / 8;
+
+            return (raw_temp > 0) ? raw_temp : 0;
+
         } else if (cpuid_info.version.extendedFamily > 0) { // Target K10 to K15 (Bulldozer)
 
             regl = pci_config_read32(0, 24, 3, AMD_TEMP_REG_K10);
--- a/system/temperature.h
+++ b/system/temperature.h
@@ -14,6 +14,10 @@
 #define AMD_TEMP_REG_K8     0xE4
 #define AMD_TEMP_REG_K10    0xA4
 
+#define AMD_SMU_INDEX_ADDR_REG 0xB8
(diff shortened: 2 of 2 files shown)


jonathan-teh:
Fixes #438.
I don't have the hardware to test. @hap2001gh please test and post screenshot.

hap2001gh:
I'm really sorry, I got rid of the AMD laptops because of RAM/controller problems.

jonathan-teh:
OK, I guess we'll have to wait for someone else who has this CPU to test this.

01e3:
I have a motherboard with A10-7850K (Kaveri KV-A1 [Steamroller]) where I just checked and while the temperature is reported by Memtest86+, the readings are incorrect - 23C or lower, vs 58C in BIOS.

It has extended family = 0x6, extended model = 0x3 [1] but the patch checks for (cpuid_info.version.extendedModel == 6 || cpuid_info.version.extendedModel == 7) so I suspect it cannot be used for the validation?

[1]
   vendor_id = "AuthenticAMD"
   version information (1/eax):
      processor type  = primary processor (0)
      family          = 0xf (15)
      model           = 0x0 (0)
      stepping id     = 0x1 (1)
      extended family = 0x6 (6)
      extended model  = 0x3 (3)
      (family synth)  = 0x15 (21)
      (model synth)   = 0x30 (48)

jonathan-teh:
Steamroller is one generation before Excavator, please raise a separate issue. What temperature does Linux's k10temp report?

01e3:
BTW - just for fun I purchased a AMD X4 845 CPU so in several days, when it finally arrives, I should be able to confirm the code works. To be precise, this my 3rd attempt for buying it, the first two were... well... less than successful so I'm dealing with returns. Not sure what is so special about that CPU...

And for A10-7850K, yes - I will boot Linux and see if temperature reporting works there, and then file a bug.

01e3:
So... the patch works but the temperature readings are wrong (too low): 16-23C:

obraz

Immediately after the reboot BIOS was showing 42C, so we may be off by at least 25C, potentially more.

Linux (kernel 6.6) seems to have the same problem - when idle, temperature goes to 0, I suspect it may even be negative just capped to 0. Under load it shows 20C - 30C.

Given the code has been merged, I suspect at this point the best action would be to file a bug?

jonathan-teh:
Thanks for testing. According to the BKDG for Family 15h Models 60-6fh (AMD Document 50742), the reported temperature is on its own scale. Section 2.10.1 The T_ctl Temperature Scale (page 154) says:

>  T_ctl is a temperature on its own scale aligned to the processors cooling requirements. Therefore T_ctl does not represent a temperature which could be measured on the die or the case of the processor. Instead, it specifies the processor temperature relative to the maximum operating temperature, T_ctl_max. T_ctl_max is specified in the power and thermal data sheet.

As Linux has the same issue, it might be better to report it there to see if there's a fix. FWIW, family 16h (Jaguar) processors have the same issue and the same definition for T_ctl. When I had one in use, it used to read about 15C below room temperature at idle.

x86fr:
I will assemble a platform with this CPU and solve this in the next couple weeks.

Anyway, please don't submit "Ready to merge" PRs that hasn't been tested on actual hardware. 
Keep them in "draft" state and clearly say in description that the PR requires validation on real-world hardware.

PR #447, opened 2024-10-17
system/loongarch: Fix a bug about the memory boundary.

kilaterlee:
In LoongArch, the addresses from 0x30000000 to 0x80000000 are 32-bit PCI memory space, If the BAR happens start at 0x30000000, a mapping error will occur according to the old logic, and 0x30000000 should also be mapped, so fix it.

### DIFF
--- a/system/loongarch/vmem.c
+++ b/system/loongarch/vmem.c
@@ -29,7 +29,7 @@ extern uint8_t highest_map_bit;
 
 uintptr_t map_region(uintptr_t base_addr, size_t size __attribute__((unused)), bool only_for_startup __attribute__((unused)))
 {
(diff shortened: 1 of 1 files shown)


kilaterlee:
@debrouxl @x86fr Please review this patch and merge it if you looks it is fine.

kilaterlee:
@x86fr Hi Sam, I forgot to adjust the 64-bit case, 64-bit should be fixed in this patch... I will adjust the 64-bit PCIe memory attributes today, and after confirmation, I will submit a new patch, sorry.

kilaterlee:
@x86fr The new PR is here: #450 . :)

PR #449, opened 2024-10-21
Refactor the tree for better multiarch support

debrouxl:
An unexciting no-op patch for making x86 & x86_64 support more similar to la64 support, and hopefully one day aarch64 and rv64 support if someone contributes these.


kilaterlee:
The changes is looks good to me. :)

martinwhitaker:
Looks good to me.

x86fr:
Will be merged at the end of the PR merge streak to resolve all conflict at once.

01e3:
Hi.

While we are doing all this cleanup, would it be possible / desired to unify loongarch vs loongson vs la64 given we have been using all the names?

Under build64 we have la64 vs x86:
```
build64/la64
build64/x86
```

Under system we have loongarch vs x86:
```
system/loongarch
system/x86
```

Under system/imc we have loongson vs x86:
```
system/imc/loongson
system/imc/x86
```
  
Under boot loongarch vs x86:
```
boot/loongarch
boot/x86
```

@kilaterlee for vis / comment.

Also, we also have system/{arch} and system/imc/{arch}. Would it better for "imc" to live under each {arch} assuming we can align on loongson vs loongarch? Something like:

```
system/la64
system/la64/imc
system/x86
system/x86/imc
```

If so, we may want to look at splitting `system/imc/imc.h`.

debrouxl:
Your suggestions are interesting and make sense to me. Let's see what kilaterlee and others have to post about them :)

x86fr:
PR looks good as-is to me, but as @01e3 said, I would unify the Loongson's names to `la64` (replacing `loongson` and `loongarch`).

01e3:
BTW -  for the memtest_bin.lds files, they should be deleted instead of being moved. This should have been done in https://github.com/memtest86plus/memtest86plus/pull/467 but I missed it.

kilaterlee:
Sorry, I just come back from the CNY holiday, let's me take a look. :)

kilaterlee:
> Hi.
> 
> While we are doing all this cleanup, would it be possible / desired to unify loongarch vs loongson vs la64 given we have been using all the names?
> 
> As a data-point, Linux kernel seems to be using "arch/loongarch".
> 
> Under build64 we have la64 vs x86:
> 
> ```
> build64/la64
> build64/x86
> ```
> 
> Under system we have loongarch vs x86:
> 
> ```
> system/loongarch
> system/x86
> ```
> 
> Under system/imc we have loongson vs x86:
> 
> ```
> system/imc/loongson
> system/imc/x86
> ```
> 
> Under boot loongarch vs x86:
> 
> ```
> boot/loongarch
> boot/x86
> ```
> 
> @kilaterlee for vis / comment.
> 
> Also, we also have system/{arch} and system/imc/{arch}. Would it better for "imc" to live under each {arch} assuming we can align on loongson vs loongarch vs la64? For example something like using la64:
> 
> ```
> system/la64
> system/la64/imc
> system/x86
> system/x86/imc
> ```
> 
> If so, we may want to look at splitting `system/imc/imc.h`.

@01e3 I have thinking about this and here are some my opinion:
Loongson is just one user of LoongArch, and there may be others who will use it in the future. So I suggest the `imc` and `build64` should be adjust:
Under system/imc:
#### Plan A:
```
system/imc/loongarch/{vendor}
system/imc/x86/{vendor}
```
#### Plan B:
```
system/loongarch/imc/{vendor}
system/x86/imc/{vendor}
```
P.S. the vendor folder may not be necessary.
I prefer plan A because the IMC not depends on architectures, it just depends on each manufacturer's implementation, so some code under the imc folder may be can reused by different manufacturer although they may be using different architectures. @x86fr @debrouxl What are you think?


Under build64:
```
build64/loongarch
build64/x86
```
Just use the LoongArch full name, so PLS change it from la64 to loongarch.

Other changes look good to me. :)

x86fr:
Thinking about this, I wonder if the proper solution isn't to use the following :

```
build/x86 (or build/i386)
build/amd64
build/la64
```

We don't really need a build32\ and build64\ directory. A unique build folder and subfolder by arch is what many linux distros use and that seems familiar and convenient.

kilaterlee:
> So, let's try to unify the suggestions posted here :)
> 
> ```
> system/loongarch
> system/x86
> 
> system/imc/loongarch
> system/imc/x86
> 
> build/x86
> build/x86_64 (while AMD did obviously invent AMD64, some Linux distros use the more neutral x86_64)
> build/loongarch (while I think that la64 is clearer, and prevents having to rename anything later _if_ someone ports memtest86+ to la32, that's what the LoongArch port prefers, and the fact is that Linux uses arch/loongarch)
> ```
> 
> I'll do this and force push later.

If we merge build32 and build64 into the build folder, then I recommend naming the loongarch folder loongarch64:
```
build/x86
build/x86_64
build/loongarch64
```
After all we only support LoongArch64 now. Thanks! :)

x86fr:
Linux kernel only have x86/ arch, no x86_64/ (https://github.com/torvalds/linux/tree/master/arch)
Many major distros uses `amd64` (debian, ubuntu, ...). And for bash autocompletion, it's easier to do "a"+tab then having 2 directory starting with the same letter 😅

I agree with @kilaterlee about `build/loongarch64` (even if `build/la64` seems shorter)

kilaterlee:
Looks good to me, @debrouxl thanks! :)

kilaterlee:
It is recommended to update Readme.md after this PR is merged.

01e3:
I like the direction we are going here a lot, some minor comments;

1 Since we are moving things around in `/boot` for x86, and `/boot/loongarch` now has its dedicated version of `header.S` and `startup64.S`, would it make sense to also move `mbr.S` and `header.S` from /boot to `/boot/x86`? There is also `boot.h` that has a lot of very x86 specific content, but perhaps we can leave it there for now given it is currently used in all architectures.

2. We now have the following directories for `/build`:
- loongarch64 has 
- x86
- x86_64

While for LA we now have a clear distinction between the overall architecture (loongarch) vs ISA/system (loongarch64), this seems a little bit inconsistent for x86, where x86 is now used for both. Technically speaking, since 2003, x86 covers three different ISAs (16-bit, 32-bit and 64-bit).  Another reasons why I'm asking is that I have been working for some time on a PR that allows to build a multi-arch boot image, which includes both mt86plus files and supports booting on on both architectures. Being just able to use `x86` directory for that (with a `Makefile` calling `make` in the other two - 32-bit and 64-bir directories), would IMO make a lot of sense.

For UEFI, the names are `ia32` and `x64`, so perhaps this is what we could use here? Another option would be `x32` and `x64` since this is what we display currently on the screen. Both options are vendor (intel vs AMD) neutral.

3. If possible, instead of moving, please remove the memtest_bin.lds files.

kilaterlee:
> It is recommended to update Readme.md after this PR is merged.

I notice the x86 family changed them name to `mt86plus` when I update the README.md, LoongArch should follow this and add the grub support, so I will submit a PR after this PR is merged to change the memtest.efi to mt86plus, add grub support and update READEME.md.

x86fr:
I can't find a major source using `x86` and `x86_64` at the same time, and I will really hate it every times I'll use auto-completion on these names. 

If you really want to use `x86_64` instead of `amd64`, please use `ia32` or  `i586` for 32-bit to match the most common naming in the Linux community.

01e3:
If we pick `ia32` and `x64` we cam be consistent UEFI (and binary names in the Makefiles), especially that we ended up doing this already with BOOTLOONGARCH64.EFI.

Plus the names are short so we can use the same name on the screen, be be also consistent there.

Finally, there is no auto-completion collision once we add the `x86` directory.

debrouxl:
Yeah, mbr.S and header.S should probably be moved as well.
I was hoping to avoid the meaningless "x64", but yeah, that's what UEFI uses, sadly. Could you confirm the UEFI name for LA64: "loongarch64" ?
I forgot to remove the memtest_bin.lds files, I'll do that later.

kilaterlee:
> Yeah, mbr.S and header.S should probably be moved as well. I was hoping to avoid the meaningless "x64", but yeah, that's what UEFI uses, sadly. Could you confirm the UEFI name for LA64: "loongarch64" ?

Yes, I either use LoongArch64 or LoongArch when submitting LoongArch to UEFI. Some code that is only related to LoongArch SPEC will name the directory "LoongArch", if the codes are only used by LoongArch64, it's called "LoongArch64", but some details may not matched, and I don't know if the LoongArch32 uses UEFI, if so, some code should be adjusted in the future, and Memtest86+ uses the same idea.  Please refer: https://github.com/tianocore/edk2/blob/master/MdePkg/Include/LoongArch64/ProcessorBind.h, https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Uefi/UefiSpec.h
>  I forgot to remove the memtest_bin.lds files, I'll do that later.

01e3:
Right, this is what we have currently in the Makefile for LA64:
```
cp memtest.efi iso/EFI/BOOT/BOOTLOONGARCH64.EFI
```

01e3:
> I notice the x86 family changed them name to `mt86plus` when I update the README.md, LoongArch should follow this and add the grub support, so I will submit a PR after this PR is merged to change the memtest.efi to mt86plus, add grub support and update READEME.md.

Ah, yes, thank you for taking care of this.

The renames was done in https://github.com/memtest86plus/memtest86plus/pull/467, which was landing in parallel with the LA64 work, so as stated there, I was planing to do this next - was waiting for this very PR to also land.

kilaterlee:
These changes look good to me, and if `build/x64` and `build/ia32` have no objections, I will update README.md with this version.

martinwhitaker:
Personally I would prefer i586 and x86_64 in the build directory, but will go with the majority. Other than that, looks good to me.

x86fr:
`i586` & `x86_64` (with `.x32` and `.x64 `on footer) sounds good to me too.

PR #450, opened 2024-10-24
loongarch: Add 64-bit PCIe memory space mapping and change the stable timer as the main timer

kilaterlee:
Added the 64-bit PCIe memory space mapping. From 0x8000000000ULL to 0xFD00000000ULL are LoongArch 64-bit PCIe memory spaces and need to be mapped.

Since some LoongArch64 CPUs stop the performance counters when ilde, the running time displayed on the screen is incorrect. Using stable counter can solve this problem, so remove the performance counters time, and add stable counter time.


kilaterlee:
@x86fr @debrouxl PLS review this patch set, THX! :)

PR #457, opened 2024-11-11
Fix LD error while linking LoongArch64 EFI binary

x86fr:
I reinstalled from scratch the latest arch linux on the LA64 dev plaform (with all latest LA64 packages).
Unfortunately, I noticed the linker now fails while trying to link the efi file with the following message:

`ld: error: cannot change output format whilst linking LoongArch binaries`

I'm not sure if a special configuration was added previously on the dev platform or if something changed in ld

Anyway, slightly changing the build process to output an elf then converting it to binary using objcopy solves the issue.

 @kilaterlee > does this looks file to you?

### DIFF
--- a/build64/la64/Makefile
+++ b/build64/la64/Makefile
@@ -183,7 +183,9 @@ memtest.debug: memtest_shared
 
 memtest.efi: memtest_shared.bin boot/loongarch/header.o ldscripts/memtest_efi.lds
 	$(eval SIZES=$(shell size -B -d memtest_shared | grep memtest_shared))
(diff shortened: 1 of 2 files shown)


kilaterlee:
> I reinstalled from scratch the latest arch linux on the LA64 dev plaform (with all latest LA64 packages). Unfortunately, I noticed the linker now fails while trying to link the efi file with the following message:
> 
> `ld: error: cannot change output format whilst linking LoongArch binaries`
> 
> I'm not sure if a special configuration was added previously on the dev platform or if something changed in ld
> 

This error is because LoongArch binutils doesn't support direct output the binary file via `-b` option. This feature will be supported in binutils v2.44, so I using a temporary version of the compiler on CI.
> Anyway, slightly changing the build process to output an elf then converting it to binary using objcopy solves the issue.
> 
> @kilaterlee > does this looks file to you?

These changes look good to me, and I think if it hits the mark it's better than waiting for the new binutils. :)

PR #467, opened 2024-11-25
Merge bootsect.S and setup.S into header.S

01e3:
_Note: I made some mess with my rebase attempt for https://github.com/memtest86plus/memtest86plus/pull/427 that github did not like, so I'm now forced to create a new PR. Apologies for the noise._

Currently, Memtest86+ provides two (three counting mbr.S) different headers for x86: bootsect.S and header.S that together with setup.S are used to build the final binaries:

Both support loading by an intermediate bootloader. Also, both are derived from the Linux kernel, but from a very different chapter in its evolution. FDD loading was removed a very long time ago, many, many years before adding PE/COFF headers for UEFI. Also, the PE/COFF headers are large (even larger for x86-64) and take a lot of very limited space, leaving very almost no headroom for the actual code.

This commit merges bootsect.S and setup.S into header.S, adding BIOS (FDD) loading support, which should eventually enable to maintain a single, all-in-one binary that can be booted either by an intermediate bootloader (such as GRUB or LILO), UEFI and BIOS. Easier and less confusing for users, as I have continued to see cases when people assume memtest.bin is needed for non-UEFI systems, even if never intended for FDD loading.

To ensure we have enough space for both the boot code (limited to 512B including Linux boot header and and address of the New EXE header) and UEFI headers, the PE/COFF headers got moved into the ".setup" section of the code where *currently* we have 1024B total and about 90B free. Should more space be needed in the future, we can grow that section or perhaps even better - move PE/COFF into its own section that will not even need to be loaded, as also suggested in https://github.com/memtest86plus/memtest86plus/pull/136#issuecomment-1501057998

Overall, the code got copied from bootsect.S and setup.S with very, very minimal changes, most importantly:

 - Dedicated routine for printing strings (print_string) got added and lives in the ".bootsect" section.

 - Print "Loading " message almost immediately after BIOS loads and calls code from the .bootsect" section, even before setting up the custom FDD parameter table. The sooner the better.

 - Set DX to 0x0, as the code depends on its value. While the BIOS is expected to set DL register to the boot drive number which is 0 for FDD, to my knowledge there is no guarnatee for DH to have any particular value. Also, when executing "MZ" signature as x86 instructions, the original value of DX becomes instantly and irrevocably lost. This is fine as we only support booting from FDD anyway.

 - Print "Memtest86+" add its version after loading the ".setup" section. We get the string from mt86plus_version (kernel_version), no need to duplicate in the boot sector where we are very space limited and as a bonus we can provide the version information.

obraz

As diffstat shows a lot of code changed and added, it may be easier use the following commands:
 diff -Nur bootsect.S header.S
 diff -Nur setup.S header.S

Also, as there is no need to keep a separate memtest.bin after all these changes, so remove it completely. In addition, rename the main binary to "mt86plus" (no ".efi" extension!) as requested in https://github.com/memtest86plus/memtest86plus/pull/427#issuecomment-2284449652 to avoid confusion with other Memtests. Note: similar change will need to be done for the recently added LA64 arch. I will handle it in a separate PR where we can discuss all the details.


debrouxl:
Still looks reasonable to me. We'll see what the beta testing yields.

martinwhitaker (approved):
Sorry for the slow response. This still looks good to me.

martinwhitaker:
@x86fr , any objection to merging this now?

x86fr:
I’m still waiting for feedback from some beta testers, who are currently swamped with work. There’s a lot to verify, including floppy drives, PXE, 32/64-bit builds, and various bootloaders. This is a significant commit that will be challenging to roll back if any issues arise. Please allow approximately 10 more days, as I’ll be visiting the factories on the 22nd and 23rd.

x86fr:
Despite a couple of odd (but probably unrelated) issues reported, most of the compatibility tests worked as expected and the PR seems good enough to merge. I'll do that tomorrow as merging it requires some immediate changes to the Windows Installer script (and these changes must also be replicated to the nightly build process to avoid breaking it).

Good work!

debrouxl:
Great news :)

PR #473, opened 2024-12-04
system/hwquirks: Optimization Loongson 7A2000 EHCI quirk

kilaterlee:
Currentilly, it only compares the root bus DID is 0x7A00 which sometimes fails, a new logic was added to compare the 7A2000 EHCI DID to fix it.


kilaterlee:
@x86fr @debrouxl @martinwhitaker Please help review this change, THX! :)

PR #475, opened 2024-12-16
Update jedec_id.h - adding Thermaltake RAM

Firminator:
Adding Thermaltake RAM ( https://thermaltakeusa.com/collections/memory )


x86fr:
Added manually

PR #485, opened 2025-01-27
Switch Address Tests from SEQ to ONE for Faster Error Detection on HCC CPUs

x86fr:
This PR will significantly reduces the time required to identify the first memory error during the startup process on high-core-count CPUs by modifying the address testing method. Instead of running the tests across all CPU cores simultaneously (SEQ), the tests will now be performed on a single core (ONE).

The rationale behind this change is that in almost all scenarios, a computer will fail the basic Power-On Self-Test (POST) if even one core is unable to access all regions of the RAM. Consequently, performing these tests on all cores is largely redundant and inefficient, particularly in systems with a high number of cores. By testing on a single core, we maintain system reliability while dramatically improving the time needed to find the first error.


tests/tests.c
```diff
@@ -50,9 +50,9 @@
 
 test_pattern_t test_list[NUM_TEST_PATTERNS] = {
     // ena,  cpu, stgs, itrs, errs, description
-    { true,  SEQ,    1,    6,    0, "[Address test, walking ones, no cache] "},
+    { true,  ONE,    1,    6,    0, "[Address test, walking ones, no cache] "},
     {false,  SEQ,    1,    6,    0, "[Address test, own address in window]  "},
```
debrouxl, line 54: ONE for test 1 as well ? It's disabled by default, but it should probably have the same running mode as test 1.

debrouxl:
Right, running at least tests 1 and 2 (and then probably also 0, which doesn't test the entire memory range anyway) on a single core is the way to go, as we've already concluded in another item on this tracker.

Sure, philosophically, I'd be more of a fan of the "one core per socket" tradeoff. However, in practice... I would think that determining siblings requires enumerating CPUID leaves, possibly dealing with quirks on some more or less obscure platforms dating 30 years back for x86, then parsing ACPI data (which we already partially do, but still) to cope with cores disabled by the firmware according to user configuration, and possibly performing something else I'm not foreseeing here... all of this for real-world platforms where the number of sockets is 1 for virtually all laptops, 1 for a crushing majority of desktops, 1 or 2 for towers, and rarely > 2 for servers, especially in recent generations.
It's hard to argue convincingly, with a straight face, that the added complexity is really worth it :)

build64/mt86plus
debrouxl, line 1: I missed it earlier, but this was mistakenly added to this commit :)

martinwhitaker (changes requested):
OK from me once the binary file is removed. Agree with @debrouxl that test 1 should be changed to ONE too.

build64/mt86plus
01e3: Right, I have a fix for this in https://github.com/memtest86plus/memtest86plus/pull/486 - https://github.com/memtest86plus/memtest86plus/pull/486/commits/2941880efcf1e4336f27ec09e18adebe6b7db4df, perhaps I should move it to a different PR?

x86fr:
Fixed

debrouxl:
Good. Now, what should we do about test 0: let it `SEQ`, or change to `ONE` ?

x86fr:
Changing all address tests to `ONE` for consistency.

PS: I'm not even sure that the way we disable caches in `cache.h` still has any impact on "modern" CPUs. You're supposed to set the CD bit in CR0, wb invalidate **AND** set all MTRRs to uncached, which we don't do. Having MTRR set to cacheable may ignore CD bit. But well, that was added 30 years ago, when the L2 cache was on the motherboard, and way before any multi-threaded CPU.

PR #486, opened 2025-01-29
Initial set of improvements for the BIOS loader post bootsect.S+setup.S merge and  .gitignore fix.

01e3:
- Update .gitignore to reflect the new name used for the binary.

 - Do not override the first 12 bytes of the code with the FDD Parameter
   Table. While mostly harmless *for now*, it is better to use the last
   12 bytes of the stack instead. While at this, optimize the code a
   bit.

 - Print Memtest86+ version immediately after loading the Setup code.
   The sooner the better. Also - once done, $MAIN_SEG can be immediately
   loaded into %es.

 - Use $MAIN_SEG:0000 as the output buffer for probing the number of
   sectors per track. No need to use the stack for this, so
   $BOOT_STACK_TOP can be removed and %es needs to be set only once.

 - Fix print_loop to compare %cl to value 5 not to %cs:5 so no "@x"
   register name is printed for the error code - as originally intended.

 - Fix alignment for "boot_msg": and "boot_msg_end" labels and
   for the ".ascii" under the "mt86plus_version" label.

 - Several code optimizations and tweaks:
     - Do not initialize %bp which saves 2 bytes.

     - Do not initialize %fs and %gs.

     - "movw $im, %bp" instead of "leaw im, %bp" takes one less byte and
       provides no benefit when a direct value is loaded without any
       calculations:
         `bd 32 00                mov    $0x32,%bp`
         `8d 2e 32 00             lea    0x32,%bp`

     - Use "mov $0x0007, %bx" to set both %bh and %bl to the desired
       value instead of doing this in two separate instructions,

     - "inc %bp, inc %bp" (2x1 = 2 bytes) instead of "addw %bp, 2" (3 bytes),
         `45                      inc    %bp`
         `83 c5 02                add    $0x2,%bp`

     - "Reset FDC" logic can be inlined and called only once yet still
       cover the initial reset after FDD Parameter Table update and retry
       in "load_setup",

     - "cwtd" (1 byte) can be used to zero %dl if %ah < 127 (0x7f).

     - kill_motor and turn_off_cursor can be inlined as they are only
       called once and turn_off_cursor can be moved much earlier and
       fit into the pre-PE header section if "cld" gets moved back
       just before movsb.

All combined yields about 45 bytes. Not enough to think about merging mbr.S yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ gdbscript
 memtest_shared_debug.lds
 
 # Binaries
+mt86plus
 memtest_shared
 memtest_shared.bin
 *.bin
--- a/boot/boot.h
+++ b/boot/boot.h
@@ -43,9 +43,8 @@
 #define KERNEL_CS	0x10		/* 32-bit segment address for code */
 #define KERNEL_DS	0x18		/* 32-bit segment address for data */
 
-/* The following addresses are offsets from BOOT_SEG. */
+/* The following address is an offset from BOOT_SEG. */
 
-#define BOOT_STACK	((1 + SETUP_SECS) * 512)
 #define BOOT_STACK_TOP	((MAIN_SEG - BOOT_SEG) << 4)
 
 /* The following definitions must match the Linux boot_params struct. */
--- a/boot/header.S
+++ b/boot/header.S
@@ -69,33 +69,36 @@ init:
 	movw	%ax, %ds
 	movw	%ax, %es
 	movw	%ax, %ss
-	movw	$BOOT_STACK_TOP, %bp
-	movw	%bp, %sp
+	movw	$(BOOT_STACK_TOP - 12), %sp	# Reserve 12 bytes for the FDD Parameter Table
 	sti
-	cld
 
-	leaw	boot_msg, %bp
-	movw	$(boot_msg_end - boot_msg), %cx
+	# Turn off the text display cursor.
+	movb	$0x01, %ah
+	xorb	%bh, %bh			# %bh = 0: page 0
+	movw	$0x2000, %cx
+	int	$0x10
+
+	movw	$boot_msg, %si
 	call	print_string
 
 	jmp	init_fdd
 
+# Print NULL terminated string from %ds:%si via BIOS routine int 10h: 0eh (WRITE_TTY)
 print_string:
-	# Print string from %es:%bp of length %cx via BIOS routine int 10h: 1301h
-	push	%cx
-	movb	$0x03, %ah		# read cursor pos
-	xorb	%bh, %bh		# page 0
-	int	$0x10
+	lods	%ds:(%si), %al
+	cmp	$0, %al
+	jz	0f
 
-	pop	%cx
-	movb	$0x07, %bl		# (page 0 set above) attribute 7 (normal)
-	movw	$0x1301, %ax		# write string, move cursor
+	movb	$0x0e, %ah
+	xorw	%bx, %bx			# %bh = 0: page 0, %bl = 0: foreground color
 	int	$0x10
+	jmp	print_string
+0:
 	ret
 
-	boot_msg:
+boot_msg:
 	.ascii	"Loading "
-	boot_msg_end:
+	.byte	0
 
 	.org	0x3c
 # The PE header pointer.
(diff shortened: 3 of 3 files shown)


debrouxl:
All of this looks reasonable, as long as it doesn't depend on non-guaranteed BIOS properties, e.g. value of registers and flags after `int 0x10` / `int 0x13` return.
I had to look up for `cwtd`, learning that this is one of the instructions which have different mnemonics depending on the assembler dialect. `as` supports both `cwtd` and `cwd`.
`read_it` also has a single caller, but of course, we don't get to move code around freely...

01e3:
> All of this looks reasonable, as long as it doesn't depend on non-guaranteed BIOS properties, e.g. value of registers and flags after `int 0x10` / `int 0x13` return.

Well, to be completely transparent, there is a single change worth 1 byte, done in `print_string` that kinda does this. Register `%bl` is set before the first `int 10h` call, and then used during the second one:

```
print_string:
        # Print string from %es:%bp of length %cx via BIOS routine int 10h: 1301h
        push    %cx
        movb    $0x03, %ah              # read cursor pos
        movw    $0x0007, %bx            # %bh: page 0, %bl: attribute 7 (normal) - used below
        int     $0x10
        pop     %cx

        movw    $0x1301, %ax            # write string, move cursor
        int     $0x10
        ret
```

While I'm pretty confident it is fine to do, there is a bigger fish to fry here, so I'm happy to change it back. Please let me know.

> I had to look up for `cwtd`, learning that this is one of the instructions which have different mnemonics depending on the assembler dialect. `as` supports both `cwtd` and `cwd`.

Ah yes. If you prefer, I'm happy to use `cwd`, but `cwtd` is what I understand to be the official name in AT&T syntax, so this is what I have always used in such cases.

> `read_it` also has a single caller, but of course, we don't get to move code around freely...

Indeed, indeed. Note that I have some larger plans for that one and several other parts of the boot / setup code, but I don't want to make too many changes at once. I'm worried it would make things much harder to review, test, and overall - would take much longer to commit.

01e3:
Alright, how about  a new, shorter `print_string` code without register hacks, that is also shorter to call, which uses `0eh (WRITE_TTY)` instead of `1301h (write string, move cursor)`?

It also uses `%ds:%si`, so no need to worry about `%es`. For callers, using a null-terminated string costs just one byte vs 3 bytes to load the length into `%cx`.

Before (16 bytes):
```
<print_string>:
  25:   51                      push   %cx
  26:   b4 03                   mov    $0x3,%ah
  28:   30 ff                   xor    %bh,%bh
  2a:   cd 10                   int    $0x10
  2c:   59                      pop    %cx
  2d:   b3 07                   mov    $0x7,%bl
  2f:   b8 01 13                mov    $0x1301,%ax
  32:   cd 10                   int    $0x10
  34:   c3                      ret

```

After (14 bytes):
```
<print_string>:
  25:   ac                      lods   %ds:(%si),%al
  26:   3c 00                   cmp    $0x0,%al
  28:   74 08                   je     32 <print_string+0xd>
  2a:   b4 0e                   mov    $0xe,%ah
  2c:   31 db                   xor    %bx,%bx
  2e:   cd 10                   int    $0x10
  30:   eb f3                   jmp    25 <print_string>
  32:   c3                      ret
```

As a bonus, we can now disable cursor even earlier.

martinwhitaker:
What's the eventual goal here? Small savings in the code size don't really help us much - it's all thrown away as soon as we branch to the startup code.

x86fr:
I honestly don’t see the point of this change either. This section of code is critical to our startup process, and the risk of introducing unexpected bugs far outweighs the negligible benefit of saving 45 bytes. We just completed an extensive round of beta testing for the previous PR, which took quite a bit of time and effort. Restarting that entire process yet again, just for a marginal optimization, doesn’t seem justified. Stability should take priority over micro-optimizations in this case.

01e3:
On top of the bug fixes, the goal would be to have a single boot loader supporting both FDD and HDD, being able to boot from USB flash on all type of systems, see: https://wiki.osdev.org/Problems_Booting_From_USB_Flash#Booting_Reliability

I would would like to improve the boot experience - provide better progress report, avoid reading past the end of binary for the FDD loader, etc.

If we can add (fit) EBP and EBPB, we could also easily remove dependency on dosfstools from the build system, especially that mkdosfs (legacy, no present in some system) vs mkfs.fat (modern, not present on some system, at least by default) issue.

Anyway, I totally take the point of the cost of incremental changes, especially when comes to testing. How about I drop it for now, and come back when I have everything _mostly_ ready?

PR #490, opened 2025-02-06
build64: Fix iso creation on LoongArch

kilaterlee:
Ubuntu 24.04 uses xorriisofs V1.5.6, which checks for hybird formats when `-part_like_isohybird` is used, and will report the core dump if the MBR is not included.

Removed the `-part_like_isohybird` option on LoongArch to fix it, and only use EFI boot mode on LoongArch.


debrouxl:
LGTM.

PR #491, opened 2025-02-09
Display related optimizations and performance improvements 

01e3:
- Reduce screen writes by preventing unnecessary clear_screen_region activity.
- Do not perform Frame Buffer updates when nothing is changing.
- Don't pad test descriptions.

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -141,7 +141,7 @@ void display_init(void)
 
     set_foreground_colour(BLUE);
     set_background_colour(WHITE);
-    clear_screen_region(ROW_FOOTER, 0, ROW_FOOTER, SCREEN_WIDTH - 1);
+    clear_screen_region(ROW_FOOTER, 0, ROW_FOOTER, SCREEN_MAX_COL);
     prints(ROW_FOOTER, 0, " <ESC> Exit  <F1> Configuration  <Space> Scroll Lock");
     prints(ROW_FOOTER, 64, MT_VERSION "." GIT_HASH);
 #if defined (__x86_64__)
@@ -306,10 +306,10 @@ void display_start_run(void)
     if (ecc_status.ecc_enabled) {
         clear_screen_region(8, 49, 8, 53);                  // pass number
         clear_screen_region(8, 61, 8, 68);                  // error count
-        clear_screen_region(8, 74, 8, SCREEN_WIDTH - 1);    // ecc error count
+        clear_screen_region(8, 74, 8, SCREEN_MAX_COL);      // ecc error count
     } else {
         clear_screen_region(8, 49, 8, 59);                  // pass number
-        clear_screen_region(8, 68, 8, SCREEN_WIDTH - 1);    // error count
+        clear_screen_region(8, 68, 8, SCREEN_MAX_COL);      // error count
     }
 
     display_pass_count(0);
@@ -330,7 +330,7 @@ void display_start_run(void)
 
 void display_start_pass(void)
 {
-    clear_screen_region(1, 39, 1, SCREEN_WIDTH - 1);    // progress bar
+    clear_screen_region(1, 39, 1, SCREEN_MAX_COL);      // progress bar
     display_pass_percentage(0);
     pass_bar_length = 0;
     pass_ticks = 0;
@@ -338,10 +338,9 @@ void display_start_pass(void)
 
 void display_start_test(void)
 {
-    clear_screen_region(2, 39, 3, SCREEN_WIDTH - 1);    // progress bar, test details
-    clear_screen_region(4, 39, 4, SCREEN_WIDTH - 6);    // Avoid erasing paging mode
-    clear_screen_region(5, 39, 5, SCREEN_WIDTH - 1);
-    clear_screen_region(3, 36, 3, 37);                  // test number
+    clear_screen_region(2, 39, 2, SCREEN_MAX_COL);      // test progress bar
+    clear_screen_region(4, 39, 4, SCREEN_MAX_COL - 5);  // test addr / desc, keep addr mode
+    clear_screen_region(5, 39, 5, SCREEN_MAX_COL);      // test pattern
     display_test_percentage(0);
     display_test_number(test_num);
     display_test_description(test_list[test_num].description);
--- a/app/display.h
+++ b/app/display.h
@@ -157,7 +157,7 @@ typedef enum {
     printi(3, 36, number, 2, false, true)
 
 #define display_test_description(str) \
-    prints(3, 39, str)
+    printf_region(3, 39, SCREEN_MAX_COL, "[%s]", str)
 
 #define display_test_addresses(pb, pe, total) \
     printf_region(4, 39, SCREEN_MAX_COL - 5, "%kB - %kB [%kB of %kB]", pb, pe, (pe) - (pb), total)
--- a/app/display.h
+++ b/app/display.h
@@ -35,11 +35,19 @@
 
 #define ERROR_LIMIT     UINT64_C(999999999999)
 
+#define SCREEN_MAX_COL  (SCREEN_WIDTH - 1)
+
 typedef enum {
     DISPLAY_MODE_NA,
     DISPLAY_MODE_SPD,
     DISPLAY_MODE_IMC
 } display_mode_t;
+
+#define printf_region(row, col_start, col_end, ...) \
+    clear_screen_region(row, printf(row, col_start, __VA_ARGS__), row, col_end)
+
+#define prints_region(row, col_start, col_end, str) \
+    clear_screen_region(row, prints(row, col_start, str), row, col_end)
 
 #define display_cpu_model(str) \
     prints(0, 30, str)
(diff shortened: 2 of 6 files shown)


debrouxl:
I was wondering whether drawing functions slowed down memtest86+, especially for short-lived tests such as test 0, but I never tried to investigate.
Speed-wise, it certainly makes sense to add shortcuts to `lfb*_put_char()` functions, as these shortcuts are probably taken frequently enough to offset their cost. 

The third patch undoes an earlier change of mine. On x86_64, switching to fixed-size strings replaced (8 bytes of pointer to string data + 24 bytes of relocation + > 8 bytes of string data) by a constant 40 bytes of string data, so it was, and still is, a clear optimization in terms of raw bytes: on my computer, in `memtest_shared` (which contains sections not present into the final binary, but still), the end of `.rela.got` is at offset 1f9b0 instead of 1fc48 with the third patch:
```
(baseline)
  0 .text         00016c49  0000000000000000  0000000000000000  00001000  2**5
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  1 .rodata       000039f0  0000000000016c60  0000000000016c60  00017c60  2**5
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  2 .dynsym       000019b0  000000000001a650  000000000001a650  0001b650  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .dynstr       00000e16  000000000001c000  000000000001c000  0001d000  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  4 .gnu.hash     00000970  000000000001ce18  000000000001ce18  0001de18  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  5 .dynamic      00000100  000000000001d788  000000000001d788  0001e788  2**3
                  CONTENTS, ALLOC, LOAD, DATA
  6 .rela.data    00001128  000000000001d888  000000000001d888  0001e888  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  7 .rela.got     000000a8  000000000001e9b0  000000000001e9b0  0001f9b0  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  8 .data         00006b28  000000000001f000  000000000001f000  00020000  2**12
                  CONTENTS, ALLOC, LOAD, DATA
  9 .got          00000050  0000000000025b28  0000000000025b28  00026b28  2**3
                  CONTENTS, ALLOC, LOAD, DATA
 10 .bss          0004a380  0000000000025b80  0000000000025b80  00026b78  2**5
                  ALLOC

(with patch 3 only)
  0 .text         00016c70  0000000000000000  0000000000000000  00001000  2**5
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  1 .rodata       00003b5a  0000000000016c80  0000000000016c80  00017c80  2**5
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  2 .dynsym       000019b0  000000000001a7e0  000000000001a7e0  0001b7e0  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .dynstr       00000e16  000000000001c190  000000000001c190  0001d190  2**0
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  4 .gnu.hash     00000970  000000000001cfa8  000000000001cfa8  0001dfa8  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  5 .dynamic      00000100  000000000001d918  000000000001d918  0001e918  2**3
                  CONTENTS, ALLOC, LOAD, DATA
  6 .rela.data    00001230  000000000001da18  000000000001da18  0001ea18  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  7 .rela.got     000000a8  000000000001ec48  000000000001ec48  0001fc48  2**3
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  8 .data         000069c8  000000000001f000  000000000001f000  00020000  2**12
                  CONTENTS, ALLOC, LOAD, DATA
  9 .got          00000050  00000000000259c8  00000000000259c8  000269c8  2**3
                  CONTENTS, ALLOC, LOAD, DATA
 10 .bss          0004a2e0  0000000000025a20  0000000000025a20  00026a18  2**5
                  ALLOC
```
The 4K alignment constraint for .data currently (but certainly not back then) nullifies the advantage of the fixed-size strings into the final binary. Let the binary grow a bit, in such a way that the end of `.rela.got` spills beyond 0x20000 when using string pointers but not when using fixed-size strings, and the change in patch 3 temporarily raises the size by 4K... until the program grows further and string pointer + relocation + string data saves a bit of space again.
The optimal representation effectively depends on the rest of the program :)

x86fr:
PS: I've posted a recap of previous discussions about code size and cpu cycle counting in #492

01e3:
Oh, I was not aware that a "constant 40 bytes of string data" was a recent change and done intentionally, and I assumed it had always been like this - since the very, very early days. I should have checked, my bad.

Now, while I find "regular" text strings easier to use and maintain, we should not expect this section of the code to change frequently. I assume I should drop it then, especially under the spirit of Sam's guideline https://github.com/memtest86plus/memtest86plus/discussions/492?

What about the two other commits?

x86fr:
Adding a new definition for `SCREEN_MAX_COL`, well, why not.
Adding another level of nested macros with `printf_region()` is only a matter of code readability. I'm not sure if that many level of macro/function nested is really better, but that's trivial.
About returning if char in FB is identical, it saves a few CPU cycles, but can add potential issues (or not) with the serial TTY code that rely on the FB. This must be checked to be sure.

martinwhitaker:
Personally I find `SCREEN_WIDTH - 1` just as easy to read and understand as `SCREEN_MAX_COL`.

IMO, if you can see visible flickering that is reduced by using `printf_region()`, that is adequate justification for making the change. I've never noticed it myself, but I don't usually pay much attention to the screen when tests are running.

01e3:
I can drop the SCREEN_MAX_COL change if this is preferred.

We don't have to introduce print{f,s}_region, so this change can be also dropped.

Instead of:
```
#define display_test_pattern_value(pattern) \
    printf_region(5, 39, SCREEN_MAX_COL, "0x%0*x", TESTWORD_DIGITS, pattern)
``` 
... we would need to keep using something like for every place and live with the need to write the row (`5` in this case) written twice:
```
#define display_test_pattern_value(pattern) \
    clear_screen_region(5, printf(5, 39, "0x%0*x", TESTWORD_DIGITS, pattern), 5, SCREEN_WIDTH - 1);
```

It still prevents the clear / write pattern plus is IMO still better than what we have right now. with row repeated three times and column two times :
```
#define display_test_pattern_value(pattern) \
    { \
        clear_screen_region(5, 39, 5, SCREEN_WIDTH - 1); \
        printf(5, 39, "0x%0*x", TESTWORD_DIGITS, pattern); \
    }
```
Please let me know.

01e3:
Alright, I tested my HP Z440 system with 256GB RAM and E5-2667v4 (8C, 16T) by looping test #1. Memtest86+ booted in UEFI mode to get FB.

Without the changes in this PR, I get 8 passes in 60 seconds. The flickering effect is easily visible:
obraz

With the PR, I can get 24 runs in 60 seconds, so 3(!) times more. For completeness, I can re-do the test in BIOS (text) mode but I expect the result to be similar. 

The flickering effect is now also gone, but due to the frequency of updates we still get some character mixing:
obraz

As predicted, most of the CPU cycles were going to the FB handling. I would even say we could consider setting test #1 back to SEQ - the cost of running this test is now 1/3 of what it used to be. Here it seems to be about 2.5s per 256GB RAM per thread, and this is a pretty old system. I will re-enable SEQ and measure the time needed, but if that rule is correct, it would expect 40 seconds vs 2 minutes.

Later this week, I can re-assembly my E5-2697Av4 (16 cores, 32 threads) with 512GB RAM system (the best what I have on hand in the place I'm right now) and do some more benchmarks.

01e3:
> About returning if char in FB is identical, it saves a few CPU cycles, but can add potential issues (or not) with the serial TTY code that rely on the FB. This must be checked to be sure.

I don't believe TTY (Serial) relies on FB. Even before https://github.com/memtest86plus/memtest86plus/commit/1c88824a7d760353368946859ce5221ea637a45a, TTY has always relied on shadow_buffer, so unless there is a strange bug somewhere, it should even work in a headless mode, of course assuming the system itself can boot like that...

01e3:
I have done some more testing. Running `test #1` in SEQ over 16 cores and 256GB RAM took 108s for the original code and 38s with the PR. So, this is seems like an useful benchmark for the changes here.

Also, this is still a far cry from the text mode (when booted in BIOS legacy mode), where it takes about 20s to do full SEQ on either the old or the new code, but at least we are much closer now to the golden target. I guess the (rather expected) story here is that the font rendering is not super fast, and while the text mode was rather forgiving, graphics is not.

01e3:
Final update for now - time to run `test #1 [SEQ]` over 8C/16T E5-2667v4 with 256GB RAM:

|Code|time|
|---|---|
|Baseline|108 s|
|[clear_screen/print[fs] optimization](https://github.com/memtest86plus/memtest86plus/pull/491/commits/e59d075c658614b47f616bb79df36f6fe12989d3)|86 s|
|[Frame Buffer optimization](https://github.com/memtest86plus/memtest86plus/pull/491/commits/305c55e01e22bb6008aa9168b04346efc91d7474)|63 s|
|Both optimization|38 s|
|VGA|20 s|

debrouxl:
Yeah, I had also noticed the flickering on the early tests.
Besides my own R820 and DL380 Gen9, I might have a one-off opportunity to benchmark patches 1 and 2 on a CLX-SP platform soon. However, I expect the benchmark results to be similar to yours.
The fact that on your UEFI platform, test 1 currently spends over two thirds of the time drawing pixels and a third of the time actually testing memory, is really bad... Being shorter, test 0 is probably worse. Making `lfb*_put_char()` functions faster seems to be something we need to do ASAP. Looking at the code, I can't find an easy way to go further than you already did, though.

01e3:
Ah, based on your comment @debrouxl I just realized I may have created some confusion by counting tests from `#1` instead of 0. I have always been talking about `#0` - i.e. the first test, the enabled. So, I'm going through s/1/0/ in all the updates above. Sorry about that.

x86fr:
Well, if you see any improvement, that's fine. This PR has minimal benefits (with ONE), but it also comes at a low cost since it doesn't affect any critical parts of the code.

That said, I don't understand how this could be possible, unless the method we use to disable the cache in test #0 has a significant impact on framebuffer write speed.

martinwhitaker:
OK, test 0 is an outlier. On all my current machines it completes so fast I never see the display update. Testing on my old laptop (i7-3630QM 4C/8T, 16GB RAM) and booting in legacy mode, when using text mode it completes 100 passes looping test 0 in just under 10s. When using graphics mode that increases to 54s. But when looping test 1, it completes 10 passes in 34s regardless of the screen mode.

I don't object to merging this PR, but it does seem to be optimising a corner case.

01e3:
I disagree with that this is a corner case, unless we target Memtest86+ only for desktop-class systems. For high-performance systems with a lot of memory, how much CPU tax is being paid for rendering the output seem to be cost for each of the test (assuming similar number of pattern updates), it is just the other tests are generally more CPU intensive, so the tax becomes a much lover fraction.

There is also a user-visible impact due to frequent write / clear operations.

Anyway, under the new PRAC, more specifically:

> CPU cycles. The same is valid for CPU cycles. Minor speed improvements are not considered as a benefit unless they are noticeable by a human. A 1-second reduction in a 5-minute test is insignificant, let alone millisecond-level optimizations. Speed gains must not have any impact on test reliability.

... I don't believe the PR passes the bar, so I guess it is better to just withdraw it.

debrouxl:
Salvaged onto the `optimizations` branch, where I'll also take a clue from your #489 :)

martinwhitaker:
As I said, I'm not opposed to merging this. If you see noticeable flickering, that seems enough to pass the bar.

x86fr:
At least adding the return if char is unchanged (it's almost free)

debrouxl:
I'll try to test on my 32C / 256 GB RAM servers tomorrow, but yeah, I think patches 1 and 2 should be merged.

debrouxl:
On the R820 with its usual 4 x E5-4627v2 + 16 x 16 GB DDR3-PC14900R, in NUMA-aware mode to avoid absurd slowness, the timing traces enabled in `app/display.c` print the following output:
```
without these changes:
T 0: 0:00:00
T 2: 0:00:08
T 3: 0:02:06
T 4: 0:06:22
T 5: 0:25:18

with these changes:
T 0: 0:00:00
T 2: 0:00:03 (-0:00:05)
T 3: 0:01:59 (-0:00:07)
T 4: 0:06:13 (-0:00:09)
T 5: 0:25:02 (-0:00:16)
```

Test 0 is the most impacted one in relative terms, though it's pretty quick now that it only runs on ONE core. Flicker during test 0 is basically eliminated. All tests benefit from the minor speed improvement.

x86fr:
Can you check the impact with https://github.com/memtest86plus/memtest86plus/pull/491/commits/305c55e01e22bb6008aa9168b04346efc91d7474 only?

debrouxl:
Predictably, my benchmark results on the R820 match the ones posted by 01e3 at https://github.com/memtest86plus/memtest86plus/pull/491#issuecomment-2647183102 . A single run of test 3 started at 02:02, which is between 01:59 and 02:06. We need both patches for the full effect.

PR #494, opened 2025-02-10
Add support for color schemes and implement a dark mode

### LINKED ISSUE #439: Feature request: Black background
The current blue background strain user's eye a lot when used at night, especially with screens that cannot be adjusted (such as laptops).
Debian CLI installer have such option to make background black, I hope we can also have that.

x86fr:
Color vs Colour. That is the question.

Closes #439


x86fr:
Well I messed the PR by importing the code in the wrong branch. Fortunately, it will be squashed..

app/display.c
```diff
@@ -87,6 +87,37 @@ int max_cpu_temp = 0;
 
 display_mode_t display_mode = DISPLAY_MODE_NA;
 
+screen_palette_t mt_palette = {BLUE, WHITE, WHITE, BLACK, WHITE, BLUE, BLACK};
```
martinwhitaker, line 90: Do we need the `mt_` prefix?

martinwhitaker (commented):
From an Englishman. the answer is definitely colour!

app/display.c
x86fr: Not really

app/display.c
```diff
@@ -87,6 +87,37 @@ int max_cpu_temp = 0;
...
+
+//------------------------------------------------------------------------------
+// Private Functions
+//------------------------------------------------------------------------------
+
+void set_screen_palette(screen_palette_t *palette)
```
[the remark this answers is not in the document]
x86fr: Done

system/screen.h
```diff
@@ -46,9 +46,21 @@ typedef enum {
...
  */
+typedef struct {
+    screen_colour_t background;
+    screen_colour_t foreground;
+    screen_colour_t title_background;
+    screen_colour_t title_foreground;
+    screen_colour_t footer_background;
+    screen_colour_t footer_foreground;
+    screen_colour_t popup_background;
+} screen_palette_t;
```
01e3, line 59: I'm not sure if "screen_palette" - in participial palette, may be the best name, perhaps more like "color_template"? After all, we are not changing the color palette -  it stays the same, with the "classic" 15+1 VGA colors.

app/display.c
```diff
@@ -87,6 +87,37 @@ int max_cpu_temp = 0;
...
+            .background        = BLUE,
+            .foreground        = WHITE,
+            .title_background  = WHITE,
+            .title_foreground  = BLACK,
+            .footer_background = WHITE,
+            .footer_foreground = BLUE,
+            .popup_background  = BLACK
+        };
+    }
+}
```
01e3, line 119: If we convert this to a list of screen_palette_t[] and add a .name field with values like "classic" and "dark" then it would be super trivial to add another option in the future if we ever decide to do so.

By doing so, the initial palette (mt_palette) can simply point to entry [0] in this table so we can avoid re-peating these values in two independent places, one could even think abut adding a define for it into ./app/config.c instead of just using 0.

app/config.c
```diff
@@ -214,6 +215,8 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "rr", 3) == 0 || strncmp(params, "one", 4) == 0) {
             cpu_mode = ONE;
         }
+    } else if (strncmp(option, "dark", 5) == 0) {
+        dark_mode = true;
```
01e3, line 219: If we want, we could make it more future proof with a very little effort - look for option "color" or "color_template "here and pass the `param` to set_screen_palette(), so it can switch to any matching entry if it can find one...

system/screen.h
```diff
@@ -46,9 +46,21 @@ typedef enum {
 } screen_colour_t;
 
 /**
- * BIOS/UEFI(GOP) agnostic framebuffer copy
+ * Colour Palette definition
  */
+typedef struct {
+    screen_colour_t background;
+    screen_colour_t foreground;
```
01e3, line 53: Suggestion: we can just use "f_" and "b_" prefixes if we prefer shorter names...

system/screen.h
```diff
@@ -46,9 +46,21 @@ typedef enum {
...
+    screen_colour_t title_background;
+    screen_colour_t title_foreground;
+    screen_colour_t footer_background;
+    screen_colour_t footer_foreground;
+    screen_colour_t popup_background;
```
01e3, line 58: Do we also want another set for a settings/menu? In the "classic" one the text is white, disabled is gray, error text is yellow, bg is black. If we already have black background, we probably want this to be inverted? Background - white, text  -black, not sure about f_menu_error, f_menu_disabled?

app/display.c
```diff
@@ -95,21 +126,24 @@ void display_init(void)
 {
     cursor_off();
 
+    set_screen_palette(&mt_palette);
+    set_background_colour(mt_palette.background);
+
```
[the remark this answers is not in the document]
x86fr: it really doesn't matter if we have the blue background for a few seconds on UEFI at startup.

app/config.c
x86fr: That's unnecessary efforts right now. Users just requested a black background.

system/screen.h
x86fr: Tested menus with inverted colors, it doesn't look good. Black background for popups looks fine, even on dark mode.

tests/tests.c
```diff
@@ -50,9 +50,9 @@
 
 test_pattern_t test_list[NUM_TEST_PATTERNS] = {
     // ena,  cpu, stgs, itrs, errs, description
-    { true,  ONE,    1,    6,    0, "[Address test, walking ones, no cache] "},
-    {false,  ONE,    1,    6,    0, "[Address test, own address in window]  "},
-    { true,  ONE,    2,    6,    0, "[Address test, own address + window]   "},
+    { true,  SEQ,    1,    6,    0, "[Address test, walking ones, no cache] "},
```
debrouxl: You need to drop this commit, which reverts tests 0-2 to their state before your HCC speedup PR was merged.

PR #496, opened 2025-02-10
Add support for color schemes and implement a dark mode

### LINKED ISSUE #439: Feature request: Black background

x86fr:
Color vs Colour. That is the question.
Closes #439

(Clean version of #494)

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -103,6 +103,7 @@ bool            enable_numa        = false;
 bool            enable_ecc_polling = false;
 
 bool            pause_at_start     = true;
+bool            dark_mode          = false;
 
 power_save_t    power_save         = POWER_SAVE_HIGH;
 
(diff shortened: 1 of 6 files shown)


janielSilva:
***@***.***

Em seg., 10 de fev. de 2025, 19:11, Sam Demeulemeester <
***@***.***> escreveu:

app/config.c
```diff
@@ -214,6 +215,8 @@ static void parse_option(const char *option, const char *params)
         } else if (strncmp(params, "rr", 3) == 0 || strncmp(params, "one", 4) == 0) {
             cpu_mode = ONE;
         }
+    } else if (strncmp(option, "dark", 5) == 0) {
```
debrouxl, line 218: In #494, 01e3 had a point, even if for now, the user only requested a dark mode as you mentioned: I too would prefer future-proofing the command-line option, by checking for e.g. "colors=dark" or "theme=dark".

x86fr: There is no request for this and I really don't want to implement many color schemes. Let's keep it simple.

PR #497, opened 2025-02-16
DDR5 DIMM Temperature Reporting

### LINKED ISSUE #302: RFE: Display DDR5 memory modules temperature (per SPD5118 / SPD5 Hub standard)

x86fr:
This PR adds individual temperature reporting for each DDR5 UDIMM detected in the system, using the SPD Hub.

It's not ready yet as it only has been tested on a limited range of motherboards.

A check must also be added to handle cases where the Manufacturer name string is too long (WIP)

@kilaterlee > Populating the newly added `ram_slot_info[]` array in spd.c on LA64 would be great. Unfortunately, I don't have access to the datasheets to check how SPD Hub registers reads work with the Loongson I2C controllers. It's probably quite trivial as the changes in both `spd.c` and `i2c.c` are limited. Can you please check? Even an insight would be nice.

Closes #302


kilaterlee:
> This PR adds individual temperature reporting for each DDR5 UDIMM detected in the system, using the SPD Hub.
> 
> It's not ready yet as it only has been tested on a limited range of motherboards.
> 
> A check must also be added to handle cases where the Manufacturer name string is too long (WIP)
> 
> @kilaterlee > Populating the newly added `ram_slot_info[]` array in spd.c on LA64 would be great. Unfortunately, I don't have access to the datasheets to check how SPD Hub registers reads work with the Loongson I2C controllers. It's probably quite trivial as the changes in both `spd.c` and `i2c.c` are limited. Can you please check? Even an insight would be nice.
> 
> Closes #302

If I understand correctly, you want to check whether the SPD Hub registers are accessible on Loongson CPU? If correct:
AFAIK, the SPD Hub registers seem to be used only on DDR5, so:
* Currently, Loongson CPU does not support DDR5, it will be supported on the next GEN CPUs (3B6600 and later).
* SPD Hub registers should be located on the DDR5 ram, inevitable refer to DDR5 SPEC. I2C is just the way or method to accessing them, it's same as `smbus`.  So I beleve they should be accessed via I2C bus as long as DRAM refer to DDR5 SPEC. If my understanding of the SPD Hub registers is wrong, please let me know. :)

I suggest you can add `ram_slot_info[]` on the LoongArch version of `i2c.c`, just like x86, and the DRAM temputers function is NULL, I will fill them up if Loongson CPU supports DDR5.

PR #498, opened 2025-02-16
Enable SIMD up to SSE2 on x86_64

x86fr:
This remove the compiler flags preventing the use of early SIMD (MMX/SSE/SSE2) instructions.

That will be required to work on new memory tests with longer, 128-bit vector. Also, newer version of gcc seems more confused with our current parameters as I get an increasing amount of the elusive "`error: SSE register return with SSE disabled`".

As every x86_64 CPU ever released supports SSE2, this should be fine. 

PS: I'm checking if it's really necessary to add some ASM code in the boot process to set bit 9/10 in CR4 and configure MXCSR.

### DIFF
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -84,10 +84,9 @@ static const char codes[][13] = {
 //------------------------------------------------------------------------------
 
 #ifdef __x86_64__
+
 typedef uint64_t    reg_t;
-#else
-typedef uint32_t    reg_t;
-#endif
+typedef float __m128 __attribute__ ((__vector_size__ (16)));
 
 struct trap_regs {
     reg_t   ds;
@@ -99,26 +98,46 @@ struct trap_regs {
     reg_t   dx;
     reg_t   di;
     reg_t   si;
-#ifdef __x86_64__
     reg_t   r8;
     reg_t   r9;
     reg_t   r10;
     reg_t   r11;
-#else
-    reg_t   reserved1;
-    reg_t   reserved2;
-    reg_t   sp;
-#endif
+    reg_t   xmm[16*2];
     reg_t   bp;
     reg_t   vect;
     reg_t   code;
     reg_t   ip;
     reg_t   cs;
     reg_t   flags;
-#ifdef __x86_64__
     reg_t   sp;
-#endif
 };
+
+#else
+
+typedef uint32_t    reg_t;
+
+struct trap_regs {
+    reg_t   ds;
+    reg_t   es;
+    reg_t   ss;
+    reg_t   ax;
+    reg_t   bx;
+    reg_t   cx;
+    reg_t   dx;
+    reg_t   di;
+    reg_t   si;
+    reg_t   reserved1;
+    reg_t   reserved2;
+    reg_t   sp;
+    reg_t   bp;
+    reg_t   vect;
+    reg_t   code;
+    reg_t   ip;
+    reg_t   cs;
+    reg_t   flags;
+};
+
+#endif
 
 //------------------------------------------------------------------------------
 // Public Functions
--- a/app/interrupt.c
+++ b/app/interrupt.c
@@ -86,7 +86,7 @@ static const char codes[][13] = {
 #ifdef __x86_64__
 
 typedef uint64_t    reg_t;
-typedef float __m128 __attribute__ ((__vector_size__ (16)));
+typedef float __m128 __attribute__((__vector_size__ (16), __aligned__ (16)));
 
 struct trap_regs {
     reg_t   ds;
@@ -102,7 +102,8 @@ struct trap_regs {
     reg_t   r9;
     reg_t   r10;
     reg_t   r11;
-    reg_t   xmm[16*2];
+    reg_t   r12;
+    __m128  xmm[16];
     reg_t   bp;
     reg_t   vect;
     reg_t   code;
--- a/boot/startup64.S
+++ b/boot/startup64.S
@@ -251,17 +251,19 @@ flush:	movw	$KERNEL_DS, %ax
 
 	finit
 
-#if 0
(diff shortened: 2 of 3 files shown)


debrouxl:
AFAICT from my work and tests on the `simd` branch, which I applied to test 5, yes, it is necessary to add some ASM code in the boot process.
I encountered `error: SSE register return with SSE disabled` as well, and it's one of the reasons why I wanted to get rid of Ubuntu 20.04 support in the CI.

x86fr:
> AFAICT from my work and tests on the `simd` branch, which I applied to test 5, yes, it is necessary to add some ASM code in the boot process.

That's right. Tested only 4 motherboards, and one of them GPF'd without the init code. I assumed that the BIOS left the system in a fully operational state after POST but... Added it.

martinwhitaker:
For safety you will need to modify the ASM interrupt handling code to also save and restore the SIMD registers, just in case the compiler chooses to use them in the C part of the interrupt handler.

debrouxl:
Changing several `orw` -> `orb` / `andw` -> `andb` also goes for other parts of the startup code, per a4bc51aac23404b47e0983b26637b039c6c213e3 .

Something I haven't done yet on the `simd` branch is to force enable SSE on Athlon XP class processors, where the BIOS of some motherboards, e.g. my K7S5A, forgets to enable it. The Linux kernel does it. This isn't part of this x86-64-only PR, though.

x86fr:
>     * is forcing the default MXCSR value really necessary ? Also, you should check whether you couldn't do better with `movw $0x1F80, %ax` followed by some form of sign-extend or zero-extend.

Looks like all BIOS don't set MXCSR  with the same value just after post. At least it ensures the same behavior

>     * `orw $0x0600, %ax` -> `orb $0x06,%ah` please: on this ISA, we have instructions for changing the value of bits 8-15, so let's use them :)

I really hate using high bytes registers when not strictly necessary. They obfuscate the code while not adding anything (same latency and throughput on anything newer than a 286) 

>     * you could add the AVX enablement code from the above commit, commented out for now, unless you want to leave it for later implementation when we actually start using it.

ok, I will add some commented code for AVX initialization.

x86fr:
> For safety you will need to modify the ASM interrupt handling code to also save and restore the SIMD registers, just in case the compiler chooses to use them in the C part of the interrupt handler.

Ah yes!

app/interrupt.c
```diff
@@ -99,27 +98,47 @@ struct trap_regs {
     reg_t   dx;
     reg_t   di;
     reg_t   si;
-#ifdef __x86_64__
     reg_t   r8;
     reg_t   r9;
     reg_t   r10;
     reg_t   r11;
+    reg_t   xmm[16*2];
```
x86fr, line 105: For some reasons, I can't define a proper 8-byte wide (128b) register using the previous typedef. 

Using `__m128   xmm[16];` make it pop a Div by 0 Int immediately.

Any idea?

boot/startup64.S
```diff
@@ -265,6 +265,24 @@ flush:	movw	$KERNEL_DS, %ax
...
 
+#if 0
+	# Enable AVX.
+
+	movl	$1, %eax
+	cpuid
+	testl	$0x18000000, %ecx	# Check bits 27 (OSXSAVE) and 28 (AVX)
+	jz	no_avx
+	movq	%cr4, %rax
+	orq	$(1 << 18), %rax	# Set bit 18 (OSXSAVE)
```
debrouxl, line 276: orl $(1 << 18), %eax

boot/startup64.S
```diff
@@ -265,6 +265,24 @@ flush:	movw	$KERNEL_DS, %ax
...
+	movq	%rax, %cr4
+	xorq	%rcx, %rcx
+	xgetbv
+	orq	$0x7, %rax		# Enable x87 (bit 0), XMM (bit 1) and YMM (bit 2)
```
debrouxl, line 280: orb $0x07, %al

app/interrupt.c
debrouxl: Do you mean that the *compiler* crashes with an Internal Compiler Error because of a divide by zero ?

x86fr: No, compile is fine without any error.
Memtest86+ crashes at init with a Div By 0 interrupt. CS:IP all 00s.

debrouxl: What if you use `movdqu` instead of `movdqa` ? Just in order to eliminate misalignment, which would normally produce #GP, as a cause of issues.

x86fr: Same. :-/

image

x86fr: Seems to be an alignment issue, but I haven't found a better solution than manually aligning the struct by adding the r12 register. Every other GCC syntax I tried has failed.

martinwhitaker:
Given `AP_STACK_SIZE` is only 1024 bytes, pushing an extra 256 bytes may cause stack overflow. I did measure the stack usage when I set the stack sizes, but can't remember how much headroom there was.

x86fr:
Maybe we can just add an attribute to explicitly tell the compiler to NOT use SSE registers for the interrupt function?

In the other hand, it now (seems to) works as expected and we can also add 256 bytes to the stack size to be sure.

martinwhitaker:
If you increase `AP_STACK_SIZE` you will either need to reduce `MAX_APS` or relocate the stacks when the program is loaded into low memory.

debrouxl:
When x2APIC support is implemented, we'll have to increase `MAX_APS` and `MAX_APIC_IDS` beyond their current values of 255 & 256.
`MAX_PROXIMITY_DOMAINS` needn't be too high, though... even CXL-equipped monster servers with 8 or more sockets should remain well below 256 proximity domains.

PR #521, opened 2025-06-15
LLVM Support for x86_64 Linux Builds

jcalvinowens:
This PR contains eight small patches which allow Memtest86+ to be built with clang and lld. Everything works, as far as I can tell.

### DIFF
--- a/app/main.c
+++ b/app/main.c
@@ -153,6 +153,7 @@ uintptr_t   test_addr[MAX_CPUS];
 static void run_at(uintptr_t addr, int my_cpu)
 {
     uintptr_t *new_start_addr = (uintptr_t *)(addr + startup - _start);
+    void (*fn)(void) = (void (*)())new_start_addr;
 
     if (my_cpu == 0) {
         // Copy the program code and all data except the stacks.
@@ -171,7 +172,7 @@ static void run_at(uintptr_t addr, int my_cpu)
     __asm__ __volatile__("movl %0, %%edi" : : "r" (new_start_addr));
 #endif
 
-    goto *new_start_addr;
+    fn();
 }
 
 static bool set_load_addr(uintptr_t *load_addr, size_t program_size, uintptr_t lower_limit, uintptr_t upper_limit)
--- a/build/x86_64/Makefile
+++ b/build/x86_64/Makefile
@@ -1,8 +1,8 @@
-AS = as -64
-CC = gcc
-OBJCOPY = objcopy
+AS ?= as -64
+CC ?= cc
+OBJCOPY ?= objcopy
 
-GIT = git
+GIT ?= git
 
 ifeq ($(GIT),none)
   GIT_AVAILABLE = false
--- a/build/x86_64/Makefile
+++ b/build/x86_64/Makefile
@@ -172,7 +172,7 @@ FORCE:
 # then link it dynamically so I have full relocation information.
 
 memtest_shared: $(OBJS) $(MS_LDS) Makefile
-	$(LD) --warn-constructors --warn-common -static -T $(MS_LDS) -o $@ $(OBJS) && \
+	$(LD) --warn-common -static -T $(MS_LDS) -o $@ $(OBJS) && \
 	$(LD) -shared -Bsymbolic -T $(MS_LDS) -o $@ $(OBJS)
 
 memtest_shared.bin: memtest_shared
--- a/build/x86_64/Makefile
+++ b/build/x86_64/Makefile
@@ -25,6 +25,11 @@ else
   MS_LDS=ldscripts/memtest_shared.lds
 endif
 
+CCVER := $(shell $(CC) --version)
+ifneq (,$(findstring clang,$(CCVER)))
+  ASMEXTRAFLAGS=-mllvm -asm-macro-max-nesting-depth=64
+endif
+
 INC_DIRS = -I../../boot -I../../system -I../../system/imc -I../../system/x86 -I../../lib -I../../tests -I../../app -Iapp
 
 SYS_OBJS = system/acpi.o \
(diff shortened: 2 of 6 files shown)


debrouxl:
Thanks for this PR. See also #405 :)

* I think that the change in patch 1 should be generalized to the other Makefiles, if not necessarily for the benefit of clang / lld;
* if you've tried the variant of patch 2 in #405 and checked that it doesn't work, then we can proceed with your version... but for the sake of compiler correctness, they'd better be equivalent ?!
* patch 3 is the obvious workaround, but LLVM/Clang generating code which calls into `memcpy()` and `memset()` instead of inline, open-coded and optimized copy / set code remains a disgrace. As I mentioned in #405 , it's not the first time a code base I'm working on relies on GCC's good implementation of `__builtin_memcpy()` / `__builtin_memset()`, and there are certainly other ones; the proper fix belongs to LLVM/Clang :)
* patch 4 is also the obvious workaround, but I think we shouldn't use custom toolchain flags when it's reasonably easy to work around the recursion limit like I did in #405 , with the added benefit of the flags parameter;
* for patch 6, although you're pointing that the flag might not be doing anything, I'd still prefer the little dance to keep it with GNU ld or compatible, if that doesn't make the Makefiles too ugly ;)
* ACK for patches 5, 7 and certainly also for patch 8, as being explicit in requesting a section we need to be exported does not hurt... though it looks like a workaround for another toolchain bug, be it `ld` not dropping that section while it should or `lld` dropping it while it shouldn't ?

jcalvinowens:
Thanks for taking a look.

> I think that the change in patch 1 should be generalized to the other Makefiles, if not necessarily for the benefit of clang / lld

Agreed, I can do that. The original reason I was poking at this at all was to be able to cross-build it in Yocto.

> if you've tried the variant of patch 2 in
https://github.com/memtest86plus/memtest86plus/pull/405 and checked that it doesn't work, then we can proceed with your version... but for the sake of compiler correctness, they'd better be equivalent ?!

Oh, I actually think yours is cleaner without the temporary.
I haven't actually checked: ostensibly it's a `call` instruction now, and a pointless turd is being dumped on the stack. I guess it doesn't matter much, but I meant to try and make `asm goto` work and haven't had a chance yet.

> patch 3 is the obvious workaround, but LLVM/Clang generating code which calls into memcpy() and memset() instead of inline, open-coded and optimized copy / set code remains a disgrace. As I mentioned in
https://github.com/memtest86plus/memtest86plus/pull/405 , it's not the first time a code base I'm working on relies on GCC's good implementation of __builtin_memcpy() / __builtin_memset(), and there are certainly other ones; the proper fix belongs to LLVM/Clang :)

Language layering is not my strong suit, so I'm certainly not the one to take that argument to them :)

I didn't try very hard, let me see if I can find a way to make it inline them. 

> patch 4 is also the obvious workaround, but I think we shouldn't use custom toolchain flags when it's reasonably easy to work around the recursion limit like I did in
https://github.com/memtest86plus/memtest86plus/pull/405 , with the added benefit of the flags parameter;

That's fine with me, mind if I just include that with your authorship when I refresh the PR?

> for patch 6, although you're pointing that the flag might not be doing anything, I'd still prefer the little dance to keep it with GNU ld or compatible, if that doesn't make the Makefiles too ugly ;)

Will do.

> ACK for patches 5, 7 and certainly also for patch 8, as being explicit in requesting a section we need to be exported does not hurt... though it looks like a workaround for another toolchain bug, be it ld not dropping that section while it should or lld dropping it while it shouldn't ?

I agree, I want to dig into exactly why it does that.

debrouxl:
> Language layering is not my strong suit, so I'm certainly not the one to take that argument to them :)
> I didn't try very hard, let me see if I can find a way to make it inline them.

In #405 , I tried fairly hard, but `__builtin_mem{cpy,set}(_inline)?()` didn't do the job last year. It's the reason why that PR remains a draft.

> That's fine with me, mind if I just include that with your authorship when I refresh the PR?

Sure, though unless it's in the commit message, the authorship will disappear when the PR is squashed as part of the integration anyway.

x86fr:
@debrouxl What are ur thoughts on this PR in its current state? Can we merge it as-is for a first step? Except for maybe the reordered sections, It looks safe to me.

debrouxl:
In its current state, this PR tries to special-case support for Clang where it's simple enough not to do so while gaining functionality, as I did in #405 . However, the `?=` changes in the Makefile are good (and they need to be expanded to other Makefiles, probably like the changes to ldscripts). I'm going to build a mix of #405 and a reduced version of this.

debrouxl:
I think that #405 can now be merged as a first step, see https://github.com/memtest86plus/memtest86plus/pull/405#issuecomment-3369295167 . This PR should remain open for the last step, though: the ability to build with lld when we have tested the section motion which avoids a `ld.lld: error: section: .got is not contiguous with other relro sections` error more widely.
memtest86+ is a special-purpose OS with limited hardware support and no way to enter arbitrary user input, and it keeps relro sections throughout the program's lifetime, so we're unlikely to care about these relro sections being contiguous. However, if some toolchains mandate that while not disturbing our program's operation, why not.

debrouxl:
All but one set of changes have been integrated as part of #405 , see the discussion there. This one can remain open for the last bit of sections reordering... unless either lld lifts that limitation, or that the got tables can be non- contiguous is shown to be a ld deficiency ?

jcalvinowens:
Thanks @debrouxl for getting this over the finish line, I should have some time to go over that and refresh this with the remaining changes in the next week or two.

I agree with you, the relro thing is a bit silly given the bare metal nature of this code, I can ask about it upstream. It was a while ago, but I think `-Wl,-z,norelro` also "fixed" the problem. I'll check.

x86fr:
Thank you for your efforts leading to #405 ! Seems we can close this one now.

PR #525, opened 2025-07-05
imc: Enable support for AMD Cezanne (Zen 3 APU)

jonathan-teh:
Tested with Ryzen 7 5700G (AM4), it matches the manual configuration in the firmware.
imc-czn-r7-5700g


debrouxl:
LGTM. Unsurprisingly, it works on my Ryzen 7 PRO 5750GE as well, the 1066 MHz / DDR4-2133 display being a consequence of the computer still running with 2 basic DDR4-2133 8 GB SO-DIMM sticks mounted on a SO-DIMM - DIMM adapter, as I could never manage to source the 16 GB DDR4-3600 CL14 DIMMs listed in the MB's Memory QVL...

Sam ?

Issue #523, opened 2025-06-29
IPMI keyboard not functioning


RSully:
Hi, I built a new server with an SIENAD8-2L2T board that has IPMI. The virtual keyboard from my web viewer does not work with memtest86+

`sudo lsusb -d -v` output for the keyboard when booted to Ubuntu 24.04.3:

```
Bus 001 Device 006: ID 046b:ff10 American Megatrends, Inc. Virtual Keyboard and Mouse
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0        64
  idVendor           0x046b American Megatrends, Inc.
  idProduct          0xff10 Virtual Keyboard and Mouse
  bcdDevice            5.04
  iManufacturer           1 American Megatrends Inc.
  iProduct                2 Virtual Keyboard Gadget
  iSerial                 3 0123456789
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0029
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          4 config 1
    bmAttributes         0xe0
      Self Powered
      Remote Wakeup
    MaxPower                2mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass         3 Human Interface Device
      bInterfaceSubClass      1 Boot Interface Subclass
      bInterfaceProtocol      1 Keyboard
      iInterface              5 HID Interface
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.01
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 Report
          wDescriptorLength      65
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0008  1x 8 bytes
        bInterval               4
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
[...]

PR #535, opened 2025-08-08
Add USB hub quirk to support AMI virtual keyboard (issue #523)

martinwhitaker:
From testing it appears that the AMI virtual devices do not work correctly if you disable one port on the virtual hub then reuse the USB address for a device connected to another port on that hub. So add a quirk to detect the virtual hub from its vendor/product ID and leave all ports on that hub active when performing the scan for USB keyboards. This will permanently reserve a bit more memory, but not significantly so.

### DIFF
--- a/system/usb.h
+++ b/system/usb.h
@@ -87,6 +87,10 @@
 #define HUB_PORT_LOW_SPEED      0x00000200
 #define HUB_PORT_HIGH_SPEED     0x00000400
 
+// Vendor IDs
+
+#define USB_VID_AMERICAN_MEGATRENDS     0x046b
+
 // Data structures.
 
 typedef struct __attribute__((packed)) {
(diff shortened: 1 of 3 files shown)


debrouxl:
It makes sense, and it's not like the implementation is complex :)
Maybe the AMI PCI vendor ID should have its own entry in `system/pci.h`, like e.g. the Supermicro PCI vendor ID, which is currently used at a single location as well ?

martinwhitaker:
> Maybe the AMI PCI vendor ID should have its own entry in `system/pci.h`, like e.g. the Supermicro PCI vendor ID, which is currently used at a single location as well ?

The USB vendor IDs don't generally match the PCI vendor IDs, so we'd need to add them in system/usb.h.

debrouxl:
Right, `system/usb.h`.

PR #537, opened 2025-08-13
Update Intel i801 PCI device IDs list

debrouxl:
Here's a mechanical synchronization of the list of SMBus controller PCI IDs from https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/i2c/busses/i2c-i801.c as of 6.16 / 6.17rc1 . The entries are disabled until a successful test on our side.
For some reason, the Linux driver does not contain the Lunar Lake SMBus controller PCI ID, namely `0xa822`, which is both on the open Internet and in https://github.com/coreboot/coreboot/blob/master/src/include/device/pci_ids.h with a `LNL` name.
The other discrepancy I noticed between our list and the ones in Linux and Coreboot is `0x5ad4` for Broxton platforms: it isn't in our list. ISTR that there's something about this one, but I don't remember what...

### DIFF
--- a/system/x86/i2c.c
+++ b/system/x86/i2c.c
@@ -172,7 +172,7 @@ static const uint16_t intel_ich5_dids[] =
     0x51A3,  // Alder Lake-P (PCH)
     0x54A3,  // Alder Lake-M (PCH)
     0x7A23,  // Raptor Lake-S (PCH)
-    //0x7E22,  // Meteor Lake-P (SOC)
+    0x7E22,  // Meteor Lake-P (SOC)
     //0xAE22,  // Meteor Lake-S (PCH)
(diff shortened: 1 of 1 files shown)


x86fr:
Good, but Meteor Lake-S does not exist. They only released the -H, -U, and -PS variants. These codenames are usually added very early to the Linux kernel and remain there even if the CPU is cancelled before tape-out.

debrouxl:
Alright, fixed.
What about `0x5ad4` for Broxton platforms ?

x86fr:
Broxton was a SoC for smartphones & tablet, so no DIMM and no SPD support (only BGAs).
It has been cancelled in 2016 before release.

x86fr:
PR Vampirized for MTL-P 😅

image

PR #543, opened 2025-09-07
Optimize screen writes

debrouxl:
I'd like to revive the two screen writes optimization patches from 01e3. This is the easiest one; the other one is currently 52d8bde0d6ec3750db2f86eca777e3dbf1306402 on the `optimizations` branch. Besides reducing flicker, each patch was shown to save seconds on the tests which are the most screen-write-intensive - basically, more RAM testing, less CPU waste.

AFAICT, the only situation this patch might have a visible impact is an external corruption of the framebuffer... which is an area of memory we don't test anyway, and I'd argue that if there really is such memory corruption, we want to make it more visible.

Of course, I can add the other patch to the same PR, but I think that this one could go separately, before the upcoming release, without invalidating ongoing testing rounds.

### DIFF
--- a/system/screen.c
+++ b/system/screen.c
@@ -140,6 +140,10 @@ static int lfb_offset(int row, int col, int x, int y, int bpp)
 
 static void lfb8_put_char(int row, int col, uint8_t ch, uint8_t attr)
 {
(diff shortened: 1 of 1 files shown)


x86fr:
Looks fine.

PR #545, opened 2025-09-14
Add support for AMD Zen5 Granite Ridge (GRG) ECC polling

x86fr:
Also fix an issue with Rembrandt ECC polling where only half of the channels where polled.


debrouxl:
Looks reasonable. Not that I could test, as I don't have access to desktop-class Zen 4 or Zen 5 hardware anyway, let alone with ECC RAM, and affected by errors :)

PR #546, opened 2025-10-03
tty mode: don't use ESC key and don't promote function keys

patrickdepinguin:
TTY escape sequence normally start with the Esc character, e.g. F1 is:
        ^[[11~
(where ^[ represents the Esc character)
This actually translates in multiple keys being pressed:
        0x1b  Esc
        0x5b  [
        0x31  1
        0x31  1
        0x7e  ~

This means that pressing a single real key can cause successive inputs to get_key(), although not all keys seem to be captured. In any case, this can trigger accidental reboot and incorrect menu operation.

The use of digits instead of function keys is already supported. Adapt the prompts to mention these digits instead of function keys in tty mode.

Additionally, replace the ESC key to exit/reboot with 'q'.

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -312,11 +312,15 @@ static void parse_command_line(char *cmd_line, int cmd_line_size)
 
 static void display_initial_notice(void)
 {
+    char *fmt;
     if (smp_enabled) {
-        display_notice("Press <F1> to configure, <F2> to disable SMP, <Enter> to start testing");
+        fmt = "Press <%s> to configure, <%s> to disable SMP, <Enter> to start testing";
     } else {
-        display_notice("Press <F1> to configure, <F2> to enable SMP, <Enter> to start testing ");
+        fmt = "Press <%s> to configure, <%s> to enable SMP, <Enter> to start testing ";
     }
+    // strlen is not fully accurate since the format specifiers can expand, but
+    // without sprintf this is the best we can do.
+    display_notice_with_args(strlen(fmt), fmt, get_function_key_label(1), get_function_key_label(2));
 }
 
 static void update_num_pages_to_test(void)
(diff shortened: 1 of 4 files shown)


debrouxl:
* the rationale for the change looks sound to me.
* the patch doesn't build as is, it needs a couple `prints` turned into `printf` at `app/config.c` lines 671 and 672.
* the implementation needs some optimization work, which I'm currently attempting to perform locally (several more ideas to try): in general, arrays of char pointers are very space-inefficient, all the more these ones are not static :)

debrouxl:
I've halved the size cost of this PR (1803 -> 906 bytes here) mainly by optimizing strings + usage thereof, and to a lesser extent by inlining two of the three keyboard functions added by this PR: [546.patch](https://github.com/user-attachments/files/22701521/546.patch) .
Checking for out of bounds values of `num` (which ought to be `unsigned int`, BTW) in `get_function_key_label` is superfluous IMO: it's a compile-time constant, we humans simply need to pay attention to what we're doing.

Actually, inlining the third keyboard function as well saves a whopping 5 more bytes, but Sam would probably yell at me for doing that, although in the newest C standards, it would also enable using a `static_assert` to enforce at compile-time that the argument of `get_function_key_label` be in bounds :)

x86fr:
I understand the issue this PR is addressing, but there’s a much simpler way to achieve the same result without adding complexity across multiple parts of the codebase.

Instead of using different keys for both mode, it would be better to handle TTY escape codes properly. That’s the standard approach for software interacting with TTY input. We could simply add a call to a small new function in check_input() when TTY is enabled and 0x1B is received. This function would parse the full TTY escape sequence (using a short loop and a small delay based on the baud rate, until either the sequence is complete or a timeout occurs) and then translate it into a standard keyboard key. That should fully solve the issue.

PS: I’m currently working on various patches for the v8 release, but if no one has time to implement this, maybe I can handle it.

martinwhitaker:
> Instead of using different keys for both mode, it would be better to handle TTY escape codes properly.

But which variety of TTY escape sequences would you support? VT52? VT100? Something else?

x86fr:
VT100/220 is the Industry standard and supported by all modern terminal emulators. We can write somewhere that VT100 is the expected protocol to use for TTY.

martinwhitaker:
I will try to implement Sam's suggestion if nobody else volunteers. I can only test it in qemu - none of my machines are old enough to have a serial port.

x86fr:
It’s crazy, but nearly all modern motherboards still include a COM header, even the latest shiny gaming models. I’m using an ASUS TUF GAMING X870-PLUS motherboard connected to my main computer through a USB-to-COM serial adapter, and it works as expected.

martinwhitaker:
I'm using a Gigabyte B550i in my main desktop and an ASUS X470i in my media PC. Neither has a COM header. One of my old motherboards up in the loft does, but I'm not keen enough to climb the ladder and blow the dust off it!

I've just coded up a VT220 function key code parser - will test it in qemu later.

patrickdepinguin:
Sorry about the printf/prints . I originally created the patch on top of 7.20, while rebasing there was a merge conflict and I failed to handle the printf/prints difference correctly. I should have built after the rebase.

I'm fine with the approach to interpret escape codes. Thanks for taking this up @martinwhitaker . I will gladly test your change on real hardware.

martinwhitaker:
Thanks @patrickdepinguin. Please test the code in PR #547.

PR #547, opened 2025-10-06
Recognise VT52/VT100/VT220 key escape sequences on serial console

martinwhitaker:
As discussed in PR #546, this is an alternative solution that recognises the escape sequences for VT52, VT100, and VT220 special keys and maps the function and cursor key sequences to the appropriate ASCII characters.

### DIFF
--- a/system/keyboard.c
+++ b/system/keyboard.c
@@ -240,17 +240,17 @@ static char get_vt220_sequence1(void)
       case 'S':
         return '4';  // VT100/VT220 PF4 (F4 in terminal emulators)
       case 'T':
-        return '5';  // VT100/VT220 PF5 (F5 in terminal emulators)
+        return '5';  // F5 in some terminal emulators
       case 'U':
-        return '6';  // VT100/VT220 PF6 (F6 in terminal emulators)
+        return '6';  // F6 in some terminal emulators
       case 'V':
-        return '7';  // VT100/VT220 PF7 (F7 in terminal emulators)
+        return '7';  // F7 in some terminal emulators
       case 'W':
-        return '8';  // VT100/VT220 PF8 (F8 in terminal emulators)
+        return '8';  // F8 in some terminal emulators
       case 'X':
-        return '9';  // VT100/VT220 PF9 (F9 in terminal emulators)
+        return '9';  // F9 in some terminal emulators
       case 'Y':
-        return '0';  // VT100/VT220 PF10 (F10 in terminal emulators)
+        return '0';  // F10 in some terminal emulators
       default:
         return '\0'; // unrecognised key
         break;
--- a/system/keyboard.c
+++ b/system/keyboard.c
@@ -239,6 +239,18 @@ static char get_vt220_sequence1(void)
         return '3';  // VT100/VT220 PF3 (F3 in terminal emulators)
       case 'S':
         return '4';  // VT100/VT220 PF4 (F4 in terminal emulators)
+      case 'T':
+        return '5';  // VT100/VT220 PF5 (F5 in terminal emulators)
+      case 'U':
+        return '6';  // VT100/VT220 PF6 (F6 in terminal emulators)
+      case 'V':
+        return '7';  // VT100/VT220 PF7 (F7 in terminal emulators)
+      case 'W':
+        return '8';  // VT100/VT220 PF8 (F8 in terminal emulators)
+      case 'X':
+        return '9';  // VT100/VT220 PF9 (F9 in terminal emulators)
+      case 'Y':
+        return '0';  // VT100/VT220 PF10 (F10 in terminal emulators)
       default:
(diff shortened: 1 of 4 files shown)


martinwhitaker:
I have tested this using `qemu`, using `screen` as a terminal emulator. We might need to adjust the maximum wait time when using real hardware.

patrickdepinguin:
With these changes I get no response on any key (from the main window F1 and ESC are supposed to do something).

Looking at the code it looks to me that the F1-F5 keys aren't handled for the escape sequence I have. I tried following code but it didn't make a difference (although there is probably a different reason for that):

```
-          case '5':
-            ch1 = '5'; break;  // F5 in terminal emulators
+          case '1':  // fallthrough
+          case '2':  // fallthrough
+          case '3':  // fallthrough
+          case '4':  // fallthrough
+          case '5':  // fallthrough
+            ch1 = ch2; break;  // F1...F5 in terminal emulators
```

The escape codes I have are the 'vt' sequences specified here:
https://en.wikipedia.org/wiki/ANSI_escape_code#Terminal_input_sequences

Do you have debug print code you were using for this patch? If so, could you share it ?
My initial attempts didn't yield anything useful yet, I can look deeper at it tomorrow.

x86fr:
I'm debugging this right now on some real hardware.

Commit e285fbb broke something. All TTY inputs are ignored because the function is always returned at the "ignore mouse events" step. Testing a fix.

martinwhitaker:
@x86fr , that suggests your hardware is continually generating mouse events. That surprises me.

@patrickdepinguin , the debugging code I used was
````
--- a/system/serial.c
+++ b/system/serial.c
@@ -251,7 +251,19 @@ char tty_get_char(int max_wait_frames)
     do {
         int uart_status = serial_read_reg(&console_serial, UART_LSR);
         if (uart_status & UART_LSR_DR) {
-            return serial_read_reg(&console_serial, UART_RX);
+            char str[2] = { 0, 0 };
+            str[0] = serial_read_reg(&console_serial, UART_RX);
+            if (max_wait_frames == 0) {
+                tty_goto(22,0);
+                serial_echo_print("        ");
+                tty_goto(22,0);
+            }
+            if (str[0] == 27) {
+                serial_echo_print("ESC ");
+            } else {
+                serial_echo_print(str);
+            }
+            return str[0];
         }
         usleep(10);
         wait_time -= 10;
`````
The F1 to F4 escape sequences mentioned in that Wikipedia page look to originate from old versions of xterm. https://invisible-island.net/xterm/xterm-function-keys.html gives some more information about that. But there's no reason we shouldn't support those as well, and at first glance your patch looks correct.

**Comment by x86fr** [comment]
The added escape codes makes it work on VT100, VT100+, VT220, VT400 and FreeBSD + Linux terminals.
Pushing TTY input checking at the top also solved the issues with mouse

**Comment by x86fr** [comment]
> @x86fr , that suggests your hardware is continually generating mouse events. That surprises me.

Only tested on two motherboards so far but both showed the same behavior. No mice were connected to either system.

**Comment by x86fr** [comment]
With the added scan codes, it also works with PuTTY/KiTTY default settings for Hardware Serial. Cool!

image

**Comment by x86fr** [comment]
PS: no issues with `#define TTY_SEQ_WAIT_TIME   2`

**Comment by martinwhitaker** [comment]
One quibble - the PF5 to PF10 sequences don't come from the VT100/VT220, so the comments are wrong.

Out of curiosity, do the motherboards which showed the mouse event problem have real PS2 mouse ports or emulated ones?

**Comment by patrickdepinguin** [comment]
Thanks, with the updated key sequences, things indeed work at my end.

I am using rxvt-unicode, which follows the rxvt escape sequences, which according to the link shared by @martinwhitaker are indeed different for F1 - F4.

**Comment by x86fr** [comment]
> Out of curiosity, do the motherboards which showed the mouse event problem have real PS2 mouse ports or emulated ones?

No real PS2 mouse ports on both.

**Review by x86fr — APPROVED** [review]

File: system/keyboard.c [hunk 1]
```diff
@@ -216,6 +223,133 @@ static const char usb_hid_keymap[] = {
...
+            ch1 = '1'; break;  // F1 in terminal emulators
+          case '2':
+            ch1 = '2'; break;  // F2 in terminal emulators
+          case '3':
+            ch1 = '3'; break;  // F3 in terminal emulators
+          case '4':
+            ch1 = '4'; break;  // F4 in terminal emulators
+          case '5':
+            ch1 = '5'; break;  // F5 in terminal emulators
+          case '7':
```
debrouxl, line 299: 1-5 map directly, and 7-9 map to 6-8... that seems cursed, but so be it, I guess :)

debrouxl (approved):
No issues about the implementation itself, beyond that oddity which I noticed, but is certainly justified... I can't easily check from where I am currently.

PR #548, opened 2025-10-08
TTY redraw fixes

patrickdepinguin:
Several fixes to make sure the TTY screen is updated correctly before and after tests are running, and in interrupt case.

### DIFF
--- a/app/config.c
+++ b/app/config.c
@@ -798,9 +798,16 @@ static void cpu_selection_menu(void)
 
     display_cpu_selection(display_offset);
 
+    bool tty_update = enable_tty;
     bool exit_menu = false;
     while (!exit_menu) {
         bool changed = false;
+
+        if (tty_update) {
+            tty_send_region(POP_REGION);
+        }
+        tty_update = enable_tty;
+
         switch (get_key()) {
           case '1':
             changed = set_all_cpus(false, display_offset);
@@ -839,6 +846,7 @@ static void cpu_selection_menu(void)
             break;
           default:
             usleep(1000);
+            tty_update = false;
             break;
         }
         if (changed) {
--- a/app/display.c
+++ b/app/display.c
@@ -229,6 +229,10 @@ void display_init(void)
     }
 
     scroll_message_row = ROW_SCROLL_T;
+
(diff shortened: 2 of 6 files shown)


patrickdepinguin:
Have you been able to look at this PR, by any chance?

x86fr:
Some simplifications are needed (no need to add if() statements in macros, etc.).
I’ll handle that, but it might not make it into v8 since the final builds are currently being tested.

debrouxl:
From a quick glance, I hadn't found anything wrong with this PR.
Avoiding the `if()` statements in `display.h` macros is indeed possible, but AFAICT, as the ternary operator isn't really cut for that purpose, the main way to achieve that is to move these `if()` statements to the callers, which would thereby increase code duplication ?

x86fr:
we can achieve the same result with much less code, just by doing a single full screen redraw at init and before restart.

debrouxl:
Doing it more often makes it easier to cope with garbled transmission, but I see your point.
The contributed changes add a redraw upon interrupt (exception) as well; arguably, we should do whatever maximizes the accuracy of reported exception data. If redrawing helps us getting cleaner exception-related reports (very infrequent, it's been a while since we received any), then we want to redraw.

patrickdepinguin:
The main goal is that experience on TTY is identical than on a local monitor.

If you reduce the amount of redraws and still reach that goal, that is fine.

But, notice that e.g. on the initial configuration screen, pressing F2 (toggle SMP mode) requires updating the screen on each toggle since the message text changes. There are other examples like that.

Like before, I'm happy to test any alternative implementation.

x86fr:
The initial TTY implementation was added for quick feedback. To enable a full remote experience, we should stop scattering TTY logic across the codebase. A good first step could be a single `fn tty_redraw(enum TTY_REDRAW_ZONE)` with an early `if (!enable_tty) return;` to reduce repetition and avoid adding branchs in macros. It might be even more efficient to handle this at the `printx()` level. The fix is simple but requires time for validation, since no beta testers use it and we must test legacy interrupt-based TTY, modern MMIO TTY, LoongArch, etc.

PR #550, opened 2025-10-23
Request cold instead of warm reboot

V10lator:
This fixes heavy display corruption while rebooting on SteamDeck.

### DIFF
--- a/system/x86/hwctrl.c
+++ b/system/x86/hwctrl.c
@@ -55,10 +55,10 @@ void hwctrl_init(void)
 void reboot(void)
 {
     // Use cf9 method as first try
-    uint8_t cf9 = inb(0xcf9) & ~6;
+    uint8_t cf9 = inb(0xcf9) & ~0x0E;
     outb(cf9|2, 0xcf9); // Request hard reset
     usleep(50);
-    outb(cf9|6, 0xcf9); // Actually do the reset
+    outb(cf9|0x0E, 0xcf9); // Actually do the reset
     usleep(50);
 
     // If we have UEFI, try EFI reset service


x86fr:
The original warm reset method (0x6) was chosen over the cold reset (0xE) for 2 reasons:

1. Cold reset can freeze some motherboards, requiring a full power cycle to recover (IIRC, that was on some early DDR chipsets (E7205) and on ATi/AMD SB800 MB).
2. Some rack-mounted SuperMicro servers (IIRC Broadwell-based) have the cold reset disabled (or bugged), making reset painful when Memtest86+ is used remotly

If we have platforms not able to handle a warm reset correctly, the proper approach is to add a "reset quirk" (based on DMI strings) for that platform to use the cold reset instead.

V10lator:
I see and will have a look into it. Maybe it might also be enough to reset the screen resolution to what UEFI had to stop the screen corruption?

V10lator:
> the proper approach is to add a "reset quirk" (based on DMI strings) for that platform to use the cold reset instead.

There's a problem with that: It seems the SMBus implementation doesn't work on the SteamDeck, so there is no DMI information available. Would it be okay to detect the SteamDeck based on the CPU instead? Vangogh APUs are used inside of the SteamDeck only.

debrouxl:
I refrained from commenting because ISTR that there were computers which don't like cold reboot, but I didn't have actual examples. Sam has them, fortunately :)

However, in the case of a recent device which is hopefully still maintained by its manufacturer, maybe the proper fix for an inability to (properly) perform a reboot operation which works on virtually any other x86 computer belongs into the firmware ? Sure, having a fixed firmware on a given SteamDeck requires that a user actually installs such a patched firmware... but they need to do that once in every while anyway in order to fix vulnerabilities in the UEFI BIOS or microcode.

While making the SMBus somehow inaccessible prevents from gathering SPD data, DMI (SMBIOS) data is gathered through other means, which had better work (again, like in pretty much any x86 computer since 25+ years ago), unless there's another bug in the firmware.

x86fr:
Yes sure. (but that platform is barely 'PC' compliant so)

V10lator:
> maybe the proper fix for an inability to (properly) perform a reboot operation which works on virtually any other x86 computer belongs into the firmware ?

Sadly Valve isn't known to listen to user requests when it comes to their firmware. Also I guess warmboot is broken by design (to show the steam deck logo flicker-free while rebooting. As said there may be other workarounds but I don't know).

> Sure, having a fixed firmware on a given SteamDeck requires that a user actually installs such a patched firmware...

That would be the simple part as Valve ships and applies firmware updates with SteamOS updates. ;)

> While making the SMBus somehow inaccessible prevents from gathering SPD data, DMI (SMBIOS) data is gathered through other means, which had better work (again, like in pretty much any x86 computer since 25+ years ago), unless there's another bug in the firmware.

To be fair I didn't look much. It's quick and simple to identify it by the CPU/APU.

> that platform is barely 'PC' compliant so

Well, it's marketed as "handheld PC". But seriously, memtest86+ is recommended everywhere when it comes to overclocking the Steam Decks RAM, so it might be used much more on Steam Decks than you think. ;)

Opened a new PR which applies a quirk.

PR #551, opened 2025-10-23
Color changes

V10lator:
This makes the big banner better themeable, changes the big banner of the dark theme ane makes the dark theme default.

### DIFF
--- a/system/screen.c
+++ b/system/screen.c
@@ -67,7 +67,7 @@ static uint32_t lfb_pallete[16];
 
 static lfb_rotate_t lfb_rotate = LFB_TOP_UP;
 
-static uint8_t current_attr = WHITE | BLUE << 4;
+static uint8_t current_attr = WHITE | BLACK << 4;
 
 //------------------------------------------------------------------------------
 // Private Functions


x86fr:
Making the dark theme default is your personal preference. What's the reason to push that to main?

V10lator:
> What's the reason to push that to main?

People telling OLED displays are way too bright at night on devices like the Steam Deck where the options aren't easily changeable as you will run memtest86+ with the "load from file" option of the Steam Decks BIOS and no way to turn the screen off.

Also dark themes are more modern (like the light theme looks like an oldscool BIOS UI while dark equals modern UEFI on most systems), so isn't it the logical step to change the default to something more modern at some point? If so why not now?

x86fr:
Memtest86+ is a text-mode utility, so calling the current look 'outdated' is a bit of an overstatement. 😅

That said, a broader discussion about the color scheme (and possibly the overall layout) could be very worthwhile. However, many technicians have been using this interface daily for decades, and a drastic visual change could disrupt established workflows. Users tend to value consistency in diagnostic tools. Finally, the blue background is widely recognized as part of Memtest86+’s identity.
Using a black background by default would make it much easier to confuse it with the commercial Memtest.

A practical compromise would be to introduce a startup option for a “dark mode”, keeping the classic layout as the default.

V10lator:
Is there a better place for discussion that GitHub because I would love to participate on such a broader discussion.

The properitary memtest86 looked like this in the past:
image
Also at some point they looked like this:
image
And today they look like this:
image
So they chaged way more dramatically than just colors already while memtest86+ still looks like a tool from the 90s (and yes, that's mostly cause of the blue color).

The compromise is in git main already, correct? Again: It's not that simple to give start arguments on some platforms (like UEFI direct file loading). Also you will still have a blue flash while loading / before the theme gets even used.

x86fr:
No that's not how the main test screen looks today. Please try it.

I've created a discussion for this : https://github.com/memtest86plus/memtest86plus/discussions/555

PR #552, opened 2025-10-23
Change the CFLAGS a bit

V10lator:
Seems to run just fine but review and more testing is highly recommended.

### DIFF
--- a/build/i586/Makefile
+++ b/build/i586/Makefile
@@ -22,7 +22,7 @@ ifeq ($(DEBUG), 1)
   OPT_FAST=-Og
   MS_LDS=memtest_shared_debug.lds
 else
-  OPT_SMALL=-Os
+  OPT_SMALL=-Os -ffast-math
   OPT_FAST=-O3 -ffast-math -ftree-vectorize
   MS_LDS=ldscripts/memtest_shared.lds
 endif
--- a/build/loongarch64/Makefile
+++ b/build/loongarch64/Makefile
@@ -20,7 +20,7 @@ ifeq ($(DEBUG), 1)
   OPT_FAST=-Og
   MS_LDS=memtest_shared_debug.lds
 else
-  OPT_SMALL=-Os
+  OPT_SMALL=-Os -ffast-math
   OPT_FAST=-O3 -ffast-math -ftree-vectorize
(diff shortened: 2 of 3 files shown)


debrouxl:
Well... A crushing majority of the program's run time is spent in the memory testing loops, whose source files are already compiled with `-O3` (although not with `-ffast-math`, but they shouldn't be sensitive to it). Outside of that, it makes sense to use `-Os`, and to use other, code-based ways to shave seconds over the run time or bytes over the program's size. There are several of those on the `optimizations` branch.

V10lator:
> Well... A crushing majority of the program's run time is spent in the memory testing loops, whose source files are already compiled with `-O3`

but call code from other source files (like `do_tick(int my_cpu)` from `app/display.c`) in this hot loops which in return call other hot loops from other source files to draw characters to screen and stuff. I got measurable better performace by reducing calls to -Os optimized codes from withing the -O3 optimized hot loops (these where just PoC coded through, nothing you want to see in production).

So maybe selected source files, like rendering related, might benefit from `-O3` ?

Also not calling draw functions in each test loop iteration might be a good idea, will look into that some more.

x86fr:
It’s important to explain what you’re trying to achieve here.

Changing optimization flags like -Ox can seriously affect error detection. They alter timing, memory access order, loop unroll, cache behavior, etc... any of which can hide or expose errors. The current flags have been tested on a large scale and make the code behave as expected. We also had cases in the past where math-related flag changes that looked harmless led to unexpected issues.

Testing the behavior on many real hardware is very time-consuming, so such changes need a strong, measurable reason.

(This is more detailed here : https://github.com/memtest86plus/memtest86plus/discussions/492)

debrouxl:
> but call code from other source files (like `do_tick(int my_cpu)` from `app/display.c`) in this hot loops which in return call other hot loops from other source files to draw characters to screen and stuff. I got measurable better performace by reducing calls to -Os optimized codes from withing the -O3 optimized hot loops (these where just PoC coded through, nothing you want to see in production).

You'd need to provide detailed benchmarks on a range of computers, then :)

The `optimizations` branch contains a drawing optimization which we haven't integrated yet, but probably should be (or at least should be discussed again) after the upcoming release.

PR #554, opened 2025-10-23
Add Steam Deck reboot workaround and fix EFI warmboot

V10lator:
This correctly shows the APU name but somehow still does a warmboot. So creating as a draft as I'm sure I just undertsand the quirk system wrong. Any help appreciated.

### DIFF
--- a/system/hwquirks.c
+++ b/system/hwquirks.c
@@ -293,7 +293,7 @@ void quirks_init(void)
     //  -- Steam Deck coldboot workaround --
     //  ------------------------------------
     // The Steam Deck corrupts the screen while booting if warmbooted.
-    if (strncmp("AMD Vangogh (Steam Deck)", cpuid_info.brand_id.str, sizeof("AMD Vangogh (Steam Deck)")) == 0) {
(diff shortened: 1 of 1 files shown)


V10lator:
So I added this right after `quirk.type |= QUIRK_TYPE_COLDBOOT;`:
```
        prints(0, 0, "AMD VANGOGH COLDBOOT QUIRK!");
        usleep(10000000);
```
And this after `bool cold = quirk.type & QUIRK_TYPE_COLDBOOT;`:
```
    clear_message_area();
    display_notice(cold ? "Coldboot" : "Warmboot");
    usleep(10000000);
```
The result on a Steam Deck is that it shows `AMD VANGOGH COLDBOOT QUIRK!` while loading the .efi and `Warmboot` while rebooting. I really don't get this, what am I doing wrong?

V10lator:
Here a video showing both, the quirk not working and the garbled display while rebooting without the quirk applied:
https://public.v10lator.de/html/memtest.mp4

x86fr:
The Steam Deck relies on proprietary display methods that don’t follow standard protocols.
Since a proper display output can’t be achieved, it’s considered incompatible

PR #556, opened 2025-10-25
[RFC] Add screen rotation quirks

### LINKED ISSUE #553: Garbled display output if screen.mode != "bios" and screen.rhs-up
As the title says the output is garbled when the resolution is not the native EFI resolution and the screen is rotated. Same resolution works without rotation. Also rotated output works correctly if `screen.mode="bios"` (but ofc. produces a small image).

V10lator:
This fixes #553 and introduces a new bug where the display area isn't correctly centered anymore. At the time of writing I have no idea why it does either of both.

Here a picture of a Steam Deck with the quirk (QUIRK_TYPE_ROTATE_90) applied:

### DIFF
--- a/app/main.c
+++ b/app/main.c
@@ -214,6 +214,10 @@ static void global_init(void)
 
     hwctrl_init();
 
+    pci_init();
+
+    quirks_init();
+
     screen_init();
 
     cpuinfo_init();
@@ -221,10 +225,6 @@ static void global_init(void)
     pmem_init();
 
     heap_init();
-
-    pci_init();
-
-    quirks_init();
 
     acpi_init();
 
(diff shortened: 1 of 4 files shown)


x86fr:
The Steam Deck relies on proprietary display methods that don’t follow standard protocols. 
Since a proper display output can’t be achieved, it’s considered incompatible

debrouxl:
Linux is also forced to deal with the Steam Deck's buggy firmware: https://www.phoronix.com/news/Linux-7.0-EFI -> https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0c61526621ec1916527d6f6226d8a466340cca22 .

PR #561, opened 2025-11-16
Avoid glitch in runtime hours display when time is reset

### LINKED ISSUE #560: Time field not properly cleared before starting at 0:00:00 again after runs >=10:00:00
When memtest runs for >= 10h, the time field occupies 8 characters instead of 7 (12:45:78 vs 1:34:67).

So, if memtest already ran for >10h and then starts again at 0 (due to some menu action like selecting other tests), the last digit of the seconds of the previous run keeps being displayed:

Image

This is on v7.20.549cc 64bit iso from memtest.org

Image

tormodvolden:
Use a space-padded width of two for displaying the hours.

Otherwise, if the hours had been 2 digit, the last digit from the seconds would linger after the hours were going 1 digit again.

Fixes #560

### DIFF
--- a/app/display.h
+++ b/app/display.h
@@ -194,7 +194,7 @@ typedef enum {
     }
 
 #define display_run_time(hours, mins, secs) \
-    printf(7, 51, "%i:%02i:%02i", hours, mins, secs)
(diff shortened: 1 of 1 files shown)


debrouxl:
Sounds highly reasonable :)

PR #566, opened 2025-11-24
Add: ECC polling option

Berbe:
Fixes: #436 

Adds an option to toggle `enable_ecc_polling` to `true`

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,8 @@ recognised:
     * modifies the console to print a newline after every change to the frame buffer
       * useful in logging over serial where an escape or newline is needed
     * only used when using console/serial output
+  * ecc
+    * Enable ECC polling
 
 ## Keyboard Selection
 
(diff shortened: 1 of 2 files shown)


x86fr:
README.md file must be edited to add this. Also, a menu option would be nice to enable ECC polling.

Berbe:
> README.md file must be edited to add this.

It was the plan… but eventually forgot about it: fixed!
Thanks.

I used the current documentation format, which feels odd; I opened another PR aimed at addressing that: #567

> Also, a menu option would be nice to enable ECC polling.

This could be the subject of another PR

Berbe:
@x86fr I made an attempt at adding a menu option, bundling it with the trace one in & "Boot options" menu, only loadable during the `initial` phase.

This seems to compile & work, per my own test:
I have no way to directly check the result on the `enable_ecc_polling` variable, but the `enable_trace` option seems to correctly be toggled through the new layout.

I would be glad to have this work reviewed!

x86fr:
Can you please fix the conflict and allow write to the Berbe:add/ecc-option branch (”Allow edits from maintainers”)?

PR #344, opened 2023-09-23
Add "tests" parameter to specify the tests to enable

gnubie2:
Provides a mechanism through which the tests to be run can be specified on the command line.

PR #570, opened 2025-11-28
Specify tests to run in boot options

gmongaras:
Added a new boot option `testlist` to specify the initial tests to run as opposed to having to manually change the test configuration every time memtest is run.


debrouxl:
Good, that's one of the forms of the CLI I suggested in #58 :)
I prefer the comma-separated list to the bit field representation, which was contributed in #344 .

PR #576, opened 2025-12-11
hwquirks: Detect cache on VIA VP/VPX

### LINKED ISSUE #573: Cache & RAM speed is not shown on Pentium MMX 200 MHz
Memtest86+ 8.00 doesn't show Cache and RAM speeds when run on an old computer from 1997 with Pentium MMX 200 MHz.

Going through old Memtest86+ versions 5.31b shows memory speeds while 6.00 doesn't.

Image
Image

jonathan-teh:
Extend the existing quirk for VP3/MVP3 to VP/VPX.

Fixes #573.


debrouxl:
On principle, no objection from me about the contents of this PR in its current draft state, as long as you post pictures :)

jonathan-teh:
This is waiting on @malaire to post pictures.

malaire:
Building fails using given instructions (Change directory into the `build/i586` directory and type `make`.)

```
In file included from /usr/lib/gcc/x86_64-linux-gnu/14/include/limits.h:210,
                 from /usr/lib/gcc/x86_64-linux-gnu/14/include/syslimits.h:7,
                 from /usr/lib/gcc/x86_64-linux-gnu/14/include/limits.h:34,
                 from ../../app/error.c:14:
/usr/include/limits.h:26:10: fatal error: bits/libc-header-start.h: No such file or directory
   26 | #include <bits/libc-header-start.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:146: app/error.o] Error 1
```

Seems like it's using x86_64 libraries? I don't know much about cross-compiling C but I do have e.g. `i686-linux-gnu-gcc` installed. OS is Debian 13 on x86_64 CPU.

My i586 computer is ancient from 1997, not sure if that could be used for building ...

malaire:
ok, I just had to install `gcc-14-multilib`.

With this patch L2 cache is detected and memory speeds are shown:
memtest-patched

debrouxl:
Thanks for the test. LGTM, then.

jonathan-teh:
Thanks for testing. I was under the impression that the build artefacts could be downloaded but that doesn't seem to be the case.
The L1 cache and memory throughput appears to be identical to the results from 5.31b. It looks like requiring L2 cache to measure cache and memory throughput was implemented in memtest86plus@f7ae4dd3954b219f207b670d82f3aa019383512e but doesn't say why.

PR #578, opened 2025-12-21
system: loongarch: cpuinfo: Unroll memspeed copy loop

ziyao233:
Unlike x86_64 where rep movsl/movsq could be turned into efficient uOP to achieve a good performance, on LoongArch a simple loop that copies only 8 bytes per iteration couldn't fully feed the pipeline.

This makes calculated cache bandwidth suspcious: on my Loongson 3A5000 laptop, memtest86plus reports the same bandwidth for L1 and L2, which is obviously incorrect.

Let's unroll memspeed loop to copy 32-bytes per iteration. Increasing the size further doesn't make obvious effects on the calculated bandwidth. Considering even the latest generation of Loongson cores, LA664, implement only two load address generation units and two store ones, this should be a reasonable size.

### DIFF
--- a/system/loongarch/cpuinfo.c
+++ b/system/loongarch/cpuinfo.c
@@ -123,7 +123,7 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
 
     dst = src + len;
 
-    wlen = len / 8;
+    wlen = len / 64;
     // Get number of clock cycles due to overhead
     start_time = get_tsc();
     for (i = 0; i < iter; i++) {
@@ -133,8 +133,8 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
           "move   $t2, %2\n\t"        \
           "2:\n\t"                    \
           "beqz   $t2, 1f\n\t"        \
-          "addi.d $t0, $t0, 8\n\t"    \
-          "addi.d $t1, $t1, 8\n\t"    \
+          "addi.d $t0, $t0, 64\n\t"   \
+          "addi.d $t1, $t1, 64\n\t"   \
           "addi.d $t2, $t2, -1\n\t"   \
           "b      2b\n\t"             \
           "1:\n\t"                    \
@@ -153,10 +153,24 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
         "move   $t2, %2\n\t"        \
         "2:\n\t"                    \
         "beqz   $t2, 1f\n\t"        \
-        "ld.d   $t3, $t0, 0x0\n\t"       \
-        "st.d   $t3, $t1, 0x0\n\t"       \
-        "addi.d $t0, $t0, 8\n\t"    \
-        "addi.d $t1, $t1, 8\n\t"    \
+        "ld.d   $t3, $t0, 0x0\n\t"  \
+        "st.d   $t3, $t1, 0x0\n\t"  \
+        "ld.d   $t3, $t0, 0x8\n\t"  \
+        "st.d   $t3, $t1, 0x8\n\t"  \
+        "ld.d   $t3, $t0, 0x10\n\t" \
+        "st.d   $t3, $t1, 0x10\n\t" \
+        "ld.d   $t3, $t0, 0x18\n\t" \
+        "st.d   $t3, $t1, 0x18\n\t" \
+        "ld.d   $t3, $t0, 0x20\n\t" \
+        "st.d   $t3, $t1, 0x20\n\t" \
+        "ld.d   $t3, $t0, 0x28\n\t" \
+        "st.d   $t3, $t1, 0x28\n\t" \
+        "ld.d   $t3, $t0, 0x30\n\t" \
+        "st.d   $t3, $t1, 0x30\n\t" \
+        "ld.d   $t3, $t0, 0x38\n\t" \
+        "st.d   $t3, $t1, 0x38\n\t" \
+        "addi.d $t0, $t0, 64\n\t"   \
+        "addi.d $t1, $t1, 64\n\t"   \
         "addi.d $t2, $t2, -1\n\t"   \
         "b      2b\n\t"             \
         "1:\n\t"                    \
@@ -173,10 +187,24 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
           "move   $t2, %2\n\t"        \
           "2:\n\t"                    \
           "beqz   $t2, 1f\n\t"        \
-          "ld.d   $t3, $t0, 0x0\n\t"       \
-          "st.d   $t3, $t1, 0x0\n\t"       \
-          "addi.d $t0, $t0, 8\n\t"    \
-          "addi.d $t1, $t1, 8\n\t"    \
+          "ld.d   $t3, $t0, 0x0\n\t"  \
+          "st.d   $t3, $t1, 0x0\n\t"  \
+          "ld.d   $t3, $t0, 0x8\n\t"  \
+          "st.d   $t3, $t1, 0x8\n\t"  \
+          "ld.d   $t3, $t0, 0x10\n\t" \
+          "st.d   $t3, $t1, 0x10\n\t" \
+          "ld.d   $t3, $t0, 0x18\n\t" \
+          "st.d   $t3, $t1, 0x18\n\t" \
+          "ld.d   $t3, $t0, 0x20\n\t" \
+          "st.d   $t3, $t1, 0x20\n\t" \
+          "ld.d   $t3, $t0, 0x28\n\t" \
+          "st.d   $t3, $t1, 0x28\n\t" \
+          "ld.d   $t3, $t0, 0x30\n\t" \
+          "st.d   $t3, $t1, 0x30\n\t" \
+          "ld.d   $t3, $t0, 0x38\n\t" \
+          "st.d   $t3, $t1, 0x38\n\t" \
+          "addi.d $t0, $t0, 64\n\t"   \
+          "addi.d $t1, $t1, 64\n\t"   \
           "addi.d $t2, $t2, -1\n\t"   \
           "b      2b\n\t"             \
           "1:\n\t"                    \


ziyao233:
Tested on Loongson-3A5000M @ 2.0GHz, without the patch

image

L1 and L2 are reported to have the same bandwidth, 15.2GB/s. With the patch,

image

MarsDoge:
Since the LoongArch processor has a 64-byte cache line, performing only a single 8-byte load/store operation per iteration in the L1/2 cache speed test would result in insufficient pipeline utilization, which is indeed unreasonable.

@ziyao233 I think it would be more reasonable to change the cache line size from 32 bytes to 64 bytes, thus covering the access scale of one 64-byte cache line in a single iteration.

ziyao233:
> Since the LoongArch processor has a 64-byte cache line, performing only a single 8-byte load/store operation per iteration in the L1/2 cache speed test would result in insufficient pipeline utilization, which is indeed unreasonable.
> 
> @ziyao233 I think it would be more reasonable to change the cache line size from 32 bytes to 64 bytes, thus covering the access scale of one 64-byte cache line in a single iteration.

I've tested copying 64bytes per iteration, but there isn't obvious improvement, so I doubt whether it's worth.

I'm working on a LSX (SIMD) accelerated version, let's see whether it works better.

ziyao233:
> > Since the LoongArch processor has a 64-byte cache line, performing only a single 8-byte load/store operation per iteration in the L1/2 cache speed test would result in insufficient pipeline utilization, which is indeed unreasonable.
> > @ziyao233 I think it would be more reasonable to change the cache line size from 32 bytes to 64 bytes, thus covering the access scale of one 64-byte cache line in a single iteration.
> 
> I've tested copying 64bytes per iteration, but there isn't obvious improvement, so I doubt whether it's worth.
> 
> I'm working on a LSX (SIMD) accelerated version, let's see whether it works better.

image

So it does, testing is done on the same 3A5000M machine

ziyao233:
Converting the PR to draft for now, it still requires investigation to determine whether LSX or LASX should be used.

debrouxl:
Thanks for your report and PR :)

On the one hand, in general, memtest86+ does not try to do a great job at measuring memory bandwidth, by integrating every possible trick to reach and accurately measure the optimal memory bandwidth. There are better, dedicated tools for that purpose, and we leave the optimal bandwidth measurements to them.
On the other hand, your first screenshot indicates that the current code essentially reports L1 bandwidth = L2 bandwidth ~ L3 bandwidth on your platform. That would be unexpected on most platforms anyway, and your second screenshot shows this to be significantly wrong.

There's a difference between not trying to do a great job in order to avoid complexity, and doing a provably significantly inaccurate job which can be improved by adding several instructions as you did here. Therefore, I wouldn't be against integrating a patch aiming at using a SISD copy loop for cacheline-sized blocks on LA64; we'll see what the other maintainers think.

However, as for SIMD versions (you'd probably want both LSX and LASX to maximize hardware support ?)... I'd say that we should integrate SIMD bandwidth measurements for LA64 only if we do it as well for x86/x86-64, where that would require adding unrolled SISD loops (may be faster than `rep movs` on several platforms where the latter is slow) + MMX + SSE (my Athlon XP showed SSE usage benefiting bandwidth) + SSE2 + AVX/AVX2 (separate code if AVX2 brings new possibilities for data transfer) + AVX512 + whatever else makes sense. This can technically be done, but memtest86+ has historically avoided doing so...

MarsDoge:
I agree with @debrouxl ’s point.

Adding SIMD bandwidth tests for one architecture (e.g. LA64 with LSX/LASX) would likely imply providing comparable coverage on others as well (e.g. SSE/AVX on x86, NEON/SVE on ARM), which significantly increases scope and long-term maintenance cost.

Given that, this may go beyond memtest’s primary goal of providing simple and broadly comparable memory tests, and could introduce additional complexity without clear benefit for most users.

MarsDoge:
However, I also saw a PR https://github.com/memtest86plus/memtest86plus/pull/498 that says they are preparing to support SSE2, so I think it's possible that SIMD support will be added later.  @ziyao233 's idea is also very progressive for the development of the project.

ziyao233:
> Thanks for your report and PR :)
> 
> On the one hand, in general, memtest86+ does not try to do a great job at measuring memory bandwidth, by integrating every possible trick to reach and accurately measure the optimal memory bandwidth. There are better, dedicated tools for that purpose, and we leave the optimal bandwidth measurements to them. On the other hand, your first screenshot indicates that the current code essentially reports L1 bandwidth = L2 bandwidth ~ L3 bandwidth on your platform. That would be unexpected on most platforms anyway, and your second screenshot shows this to be significantly wrong.

Agree. I originally didn't plan to make the memspeed test very accurate, but only to fix the obviously wrong result (L1 bandwidth = L2 ~= L3).

> There's a difference between not trying to do a great job in order to avoid complexity, and doing a provably significantly inaccurate job which can be improved by adding several instructions as you did here. Therefore, I wouldn't be against integrating a patch aiming at using a SISD copy loop for cacheline-sized blocks on LA64; we'll see what the other maintainers think.

Thanks.

> However, as for SIMD versions (you'd probably want both LSX and LASX to maximize hardware support ?)... I'd say that we should integrate SIMD bandwidth measurements for LA64 only if we do it as well for x86/x86-64, where that would require adding unrolled SISD loops (may be faster than `rep movs` on several platforms where the latter is slow) + MMX + SSE (my Athlon XP showed SSE usage benefiting bandwidth) + SSE2 + AVX/AVX2 (separate code if AVX2 brings new possibilities for data transfer) + AVX512 + whatever else makes sense. This can technically be done, but memtest86+ has historically avoided doing so...

Okay, I understand the point and will keep the SISD version for now. For anyone interested in the LSX (128-bit SIMD) version, here's a patch,

```
diff --git a/system/loongarch/cpuinfo.c b/system/loongarch/cpuinfo.c
index 465a83bb14aa..521abbc9f52f 100644
--- a/system/loongarch/cpuinfo.c
+++ b/system/loongarch/cpuinfo.c
@@ -27,6 +27,9 @@
 
 #define BENCH_MIN_START_ADR 0x10000000   // 256MB
+#define CSR_EUEN 0x2
+#define  CSR_EUEN_LSXEN (1 << 1)
+#define  CSR_EUEN_FPEN (1 << 0)
 
 // Public Variables
@@ -119,11 +122,17 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
     uintptr_t dst;
     uintptr_t wlen;
     uint64_t start_time, end_time, run_time_clk, overhead;
+    uint32_t euen;
     int i;
 
+    euen = __csrrd_w(CSR_EUEN);
+    __csrxchg_w(CSR_EUEN_FPEN | CSR_EUEN_LSXEN,
+                CSR_EUEN_FPEN | CSR_EUEN_LSXEN,
+                CSR_EUEN);
+
     dst = src + len;
 
-    wlen = len / 8;
+    wlen = len / 64;
     // Get number of clock cycles due to overhead
     start_time = get_tsc();
     for (i = 0; i < iter; i++) {
@@ -133,8 +142,8 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
           "move   $t2, %2\n\t"        \
           "2:\n\t"                    \
           "beqz   $t2, 1f\n\t"        \
-          "addi.d $t0, $t0, 8\n\t"    \
-          "addi.d $t1, $t1, 8\n\t"    \
+          "addi.d $t0, $t0, 64\n\t"   \
+          "addi.d $t1, $t1, 64\n\t"   \
           "addi.d $t2, $t2, -1\n\t"   \
           "b      2b\n\t"             \
           "1:\n\t"                    \
@@ -153,15 +162,21 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
         "move   $t2, %2\n\t"        \
         "2:\n\t"                    \
         "beqz   $t2, 1f\n\t"        \
-        "ld.d   $t3, $t0, 0x0\n\t"       \
-        "st.d   $t3, $t1, 0x0\n\t"       \
-        "addi.d $t0, $t0, 8\n\t"    \
-        "addi.d $t1, $t1, 8\n\t"    \
+        "vld    $vr0, $t0, 0x0\n\t" \
+        "vst    $vr0, $t1, 0x0\n\t" \
+        "vld    $vr1, $t0, 0x10\n\t"\
+        "vst    $vr1, $t1, 0x10\n\t"\
+        "vld    $vr2, $t0, 0x20\n\t"\
+        "vst    $vr2, $t1, 0x20\n\t"\
+        "vld    $vr3, $t0, 0x30\n\t"\
+        "vst    $vr3, $t1, 0x30\n\t"\
+        "addi.d $t0, $t0, 64\n\t"   \
+        "addi.d $t1, $t1, 64\n\t"   \
         "addi.d $t2, $t2, -1\n\t"   \
         "b      2b\n\t"             \
         "1:\n\t"                    \
         :: "r" (src), "r" (dst), "r" (wlen)
-        : "$t0", "$t1", "$t2", "$t3"
+        : "$t0", "$t1", "$t2", "$t3", "$f0", "$f1", "$f2", "$f3"
     );
 
     // Write these bytes
@@ -173,15 +188,21 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
           "move   $t2, %2\n\t"        \
           "2:\n\t"                    \
           "beqz   $t2, 1f\n\t"        \
-          "ld.d   $t3, $t0, 0x0\n\t"       \
-          "st.d   $t3, $t1, 0x0\n\t"       \
-          "addi.d $t0, $t0, 8\n\t"    \
-          "addi.d $t1, $t1, 8\n\t"    \
+          "vld    $vr0, $t0, 0x0\n\t" \
+          "vst    $vr0, $t1, 0x0\n\t" \
+          "vld    $vr1, $t0, 0x10\n\t"\
+          "vst    $vr1, $t1, 0x10\n\t"\
+          "vld    $vr2, $t0, 0x20\n\t"\
+          "vst    $vr2, $t1, 0x20\n\t"\
+          "vld    $vr3, $t0, 0x30\n\t"\
+          "vst    $vr3, $t1, 0x30\n\t"\
+          "addi.d $t0, $t0, 64\n\t"   \
+          "addi.d $t1, $t1, 64\n\t"   \
           "addi.d $t2, $t2, -1\n\t"   \
           "b      2b\n\t"             \
           "1:\n\t"                    \
           :: "r" (src), "r" (dst), "r" (wlen)
-          : "$t0", "$t1", "$t2", "$t3"
+          : "$t0", "$t1", "$t2", "$t3", "$f0", "$f1", "$f2", "$f3"
       );
     }
     end_time = get_tsc();
@@ -193,6 +214,8 @@ static uint32_t memspeed(uintptr_t src, uint32_t len, int iter)
 
     run_time_clk = ((len * iter) / (double)run_time_clk) * clks_per_msec * 2;
 
+    __csrwr_w(euen, CSR_EUEN);
+
     return run_time_clk;
 }
 

```

To me it doesn't add significant code size, and LSX is supported on any current Loongson desktop platforms with ACPI support. If you (or any maintainers) change the mind, I'm willing to commit the LSX version as an alternative. Otherwise the current SISD but unrolled version is fine for me, too.

ziyao233:
> There's a difference between not trying to do a great job in order to avoid complexity, and doing a provably significantly inaccurate job which can be improved by adding several instructions as you did here. Therefore, I wouldn't be against integrating a patch aiming at using a SISD copy loop for cacheline-sized blocks on LA64; we'll see what the other maintainers think.

By the way, LoongArch hardware usually utilizes 64-byte cacheline, not 32-byte ones. This patch chooses to copy 32 bytes per iteration, since larger values don't increase the resulted bandwidth when testing on real hardware. And copying 32bytes means four pairs of load and store per loop, since no Loongson design has more than 8 load/store units, this (theoretically, if the core doesn't merge load/stores) should already fully feed the pipeline.

debrouxl:
Even if copying 64 bytes instead of 32 does not yield a speedup on _current_ hardware, for the reasons you've described, I'd argue that copying a full cacheline at a time is more future-proof, though ?

ziyao233:
> Even if copying 64 bytes instead of 32 does not yield a speedup on _current_ hardware, for the reasons you've described, I'd argue that copying a full cacheline at a time is more future-proof, though ?

Yeah. I've pushed a new commit that copies 64 bytes per iteration.

MarsDoge (approved):
LGTM

x86fr:
Also looks good to me. There is definitely something we need to do on x86 arch to improve BW measurement.

debrouxl:
Since this PR, Sam integrated his work for more accurate and SIMD bandwidth measurement on x86_64: #587 . We could now take a new PR for memory bandwidth measurements using instructions from the LSX & LASX sets, as this was shown to be beneficial here.

Issue #571, opened 2025-11-29
EFI boot fails with GPF when using latest Fedora implementation of the GRUB "linux" command


martinwhitaker:
Fedora replace the upstream GRUB "linux" command. For legacy boot they make it an alias of the upstream "linux16" command, which isn't an issue for us. For EFI boot they replace it with an implementation that boots via the Linux EFI stub bootloader. This wasn't an issue for us until they added a patch (commit e41655885 in the Fedora package repository) that attempts to make the loaded code section(s) read-only and the loaded data section(s) non-executable. This is done using the UEFI firmware EFI_MEMORY_ATTRIBUTE_PROTOCOL if it is available and skipped if it is not available.

As the PE+ code section in our mt86plus binary contains the combined .text and .data sections for our application, this would mean any writes to the .data section before we switch to using our own page descriptor tables would result in a memory access exception. I have a potential fix for this, but I can't test if it works because none of my machines implement the EFI_MEMORY_ATTRIBUTE_PROTOCOL.

There is another side effect of this patch. Instead of loading just the code section at the preferred load address, it loads the header at that address, followed by the code section. This results in our .text and .data sections no longer being 4KB aligned. As a result, when we load the new page descriptor table address into CR3, we get a GPF. I have two ways to work around this problem:
1. Change `SETUP_SECTORS` in `boot.h` to 7, which makes the total header size 4KB, so preserving the 4KB alignment for the following .text and .data sections. This makes the mt86plus binary a little larger, but doesn't change the runtime memory footprint.
2. Use the GRUB "chainloader" command to start Memtest86+ instead of using the "linux" command. The EFI version of this command allows you to pass command line parameters to the loaded program, so it's just as good.

N.B. I don't use Fedora myself, but discovered this because Mageia apply the Fedora patches in their GRUB package.

debrouxl:
The fact that we'd probably have to add _some_ compatibility with W^X at some point was already mentioned in #241 , for instance. Beyond that, I implemented partial NX support a while ago; the most recent version of the patch keeps living on the `simd` branch, as does a patch which traces CR0, CR3 and CR4 upon processor exception.

martinwhitaker:
Yes, I was expecting W^X to hit us at some point. I wasn't expecting unaligned code sections though :-(

debrouxl:
I'd go for 1., forcing the alignment on sections ourselves instead of relying on `chainloader`.

damaestro:
[...]

PR #583, opened 2026-01-18
Add support for x2APIC

### LINKED ISSUE #70: Support for 256+ threads (x2APIC & AMD Extended APIC ID)

x86fr:
xAPIC will soon be deprecated in favor of x2APIC (starting with Granite Rapids and Sierra Forest for Server CPUs, and Nova Lake for Desktop/Mobile CPUs). Some new CPU (from Meteor Lake to Panther Lake) are also configured in x2APIC Mode by UEFI by default. This PR adds support for x2APIC to keep overall SMP support on these new CPUs.

MAX_APS has not been increased yet (Max 256 threads) because an evaluation of stack usage is needed to push this higher.

This PR deserves a lot of testing to be sure everything is OK (both in x2APIC and the old xAPIC mode)

Closes #70

### DIFF
--- a/system/smp.c
+++ b/system/smp.c
@@ -322,8 +322,8 @@ static memory_affinity_t memory_affinity_ranges[MAX_APIC_IDS];
 
 static uint32_t          proximity_domains[MAX_PROXIMITY_DOMAINS];
 
-static uint8_t           cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
-uint8_t                  used_cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
+static uint16_t           cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
+uint16_t                  used_cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
 
 static uintptr_t         smp_heap_page = 0;
 
@@ -559,7 +559,7 @@ static bool find_cpus_in_madt(void)
                 return false;
             }
             madt_processor_x2apic_entry_t *entry = (madt_processor_x2apic_entry_t *)tab_entry_ptr;
-            if (entry->flags & (MADT_PF_ENABLED|MADT_PF_ONLINE_CAPABLE)) {
+            if (entry->flags & MADT_PF_ENABLED) {
                 if (num_available_cpus < MAX_CPUS) {
                     cpu_num_to_apic_id[found_cpus] = entry->apic_id;
                     // The first CPU is the BSP, don't increment.
--- a/app/main.c
+++ b/app/main.c
@@ -101,7 +101,7 @@ static int              test_stage = 0;
 
 // These are exposed in test.h.
 
-uint8_t     chunk_index[MAX_CPUS];
+uint16_t     chunk_index[MAX_CPUS];
 
 int         num_active_cpus = 0;
 int         num_enabled_cpus = 1;
--- a/app/test.h
+++ b/app/test.h
@@ -25,11 +25,11 @@
  * enabled cores (in the current proximity domain, when NUMA awareness is
  * enabled).
  */
-extern uint8_t chunk_index[MAX_CPUS];
+extern uint16_t chunk_index[MAX_CPUS];
 /**
  * An array where the count of used CPUs in the current proximity domain.
  */
-extern uint8_t used_cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
+extern uint16_t used_cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
 
  /*
   * The number of CPU cores being used for the current test. This is always
--- a/boot/boot.h
+++ b/boot/boot.h
@@ -16,7 +16,7 @@
  *  - relocating the stacks when the program is loaded in low memory
  *  - adjusting the display if more than 3 digits are needed for CPU IDs
  */
-#define	MAX_APS		255		/* Maximum number of active APs */
+#define	MAX_APS		511		/* Maximum number of active APs */
 
 #define BSP_STACK_SIZE	16384		/* Stack size for the BSP */
 #ifdef __loongarch_lp64
--- a/system/smp.h
+++ b/system/smp.h
@@ -26,7 +26,7 @@
 /**
  * The maximum number of APIC IDs.
  */
-#define MAX_APIC_IDS                256
+#define MAX_APIC_IDS                512
 
 /**
  * The maximum number of NUMA proximity domains.
@@ -86,10 +86,10 @@ uint32_t smp_get_proximity_domain_idx(int cpu_num);
  * "Allocates" a CPU ID in the given proximity domain, for filling in NUMA-aware chunk index.
  * Returns the nth CPU ID found so far in the proximity domain.
  */
-static inline uint8_t smp_alloc_cpu_in_proximity_domain(uint32_t proximity_domain_idx)
+static inline uint16_t smp_alloc_cpu_in_proximity_domain(uint32_t proximity_domain_idx)
 {
-    extern uint8_t used_cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
-    uint8_t chunk_index = used_cpus_in_proximity_domain[proximity_domain_idx];
+    extern uint16_t used_cpus_in_proximity_domain[MAX_PROXIMITY_DOMAINS];
+    uint16_t chunk_index = used_cpus_in_proximity_domain[proximity_domain_idx];
     used_cpus_in_proximity_domain[proximity_domain_idx]++;
     return chunk_index;
 }


system/smp.c
```diff
@@ -1089,12 +1143,20 @@ void smp_send_nmi(int cpu_num)
 
 int smp_my_cpu_num(void)
 {
-    return num_available_cpus > 1 ? apic_id_to_cpu_num[my_apic_id()] : 0;
+    if (num_available_cpus <= 1) return 0;
+
+    int apic_id = my_apic_id();
+    for (int i = 0; i < num_available_cpus; i++) {
```
debrouxl: In general, when reviewing a patch, a O(1) access replaced by a O(n) search, at a time we're precisely trying to support more cores, even outside the hot code paths, is eyebrow-rising :)
However, in this case, I acknowledge that it's very likely to be the only reasonable implementation. Binary search would require the APIC IDs to be sorted, a property which we certainly can't rely on - even if it were mandated by the ACPI spec (I haven't checked), some platform somewhere may already have, or will, be screwed up. Other sub-linear searches would bring complexity.
IOW: boo (not your fault), but yeah...

system/smp.c
```diff
@@ -515,6 +554,22 @@ static bool find_cpus_in_madt(void)
                 found_cpus++;
             }
         }
+        else if (entry_header->type == MADT_PROCESSOR_X2APIC) {
+            if (entry_header->length != sizeof(madt_processor_x2apic_entry_t)) {
+                return false;
+            }
+            madt_processor_x2apic_entry_t *entry = (madt_processor_x2apic_entry_t *)tab_entry_ptr;
+            if (entry->flags & (MADT_PF_ENABLED|MADT_PF_ONLINE_CAPABLE)) {
```
debrouxl: If we need both bits set at once, then `entry->flags & (MADT_PF_ENABLED|MADT_PF_ONLINE_CAPABLE) == (MADT_PF_ENABLED|MADT_PF_ONLINE_CAPABLE)`, right ?

system/smp.c
```diff
@@ -343,11 +362,27 @@ static int my_apic_id(void)
 #if defined(__i386__) || defined(__x86_64__)
 static void apic_write(int reg, uint32_t val)
 {
+    if (apic_x2apic) {
+        if (reg == APIC_REG_ICRHI) {
```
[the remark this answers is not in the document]
x86fr: I kept that only for readability (and as a easy-to-understand safeguard for future use). Eg: `if (reg == APIC_REG_ICRHI) { return; }` is dead code because the only call using APIC_REG_ICRHI is not launched in x2APIC mode.

x86fr:
Memtest86+ running on more than 256 threads for the first time ever!

That platform is painful to work with, but it's the only one I have with >256 cores.

I increased MAX_APS to 512, but a stack size evaluation is needed. Any test on a 512 threads platform would be great.

PS: I tested with NUMA enabled on both this one and EPYC without issues.

image

martinwhitaker:
IIRC, I calculated that 256 threads was the limit when using low memory. So we'll need to prohibit use of low memory for higher thread counts.

Do you actually see any run time improvements from enabling that many threads? I would have expected the memory bandwidth to be saturated with far fewer threads. And the barrier overhead has an O(n) component.

I don't have much free time at the moment, but will try to review this at the weekend.

debrouxl:
Do you remember whether you performed this memory consumption computation before, or after, the program low-load range was widened by 93c9c8ded587323ff842d9ea3da6de4983c2e3b2 ?

martinwhitaker:
> Do you remember whether you performed this memory consumption computation before, or after, the program low-load range was widened by [93c9c8d](https://github.com/memtest86plus/memtest86plus/commit/93c9c8ded587323ff842d9ea3da6de4983c2e3b2) ?

I may be misremembering, but I think the program still gets loaded into low memory (below 1MB) when booting in real mode.

x86fr:
> IIRC, I calculated that 256 threads was the limit when using low memory. So we'll need to prohibit use of low memory for higher thread counts.
> 
> Do you actually see any run time improvements from enabling that many threads? I would have expected the memory bandwidth to be saturated with far fewer threads. And the barrier overhead has an O(n) component.

Absolutely. I don’t expect any significant performance gain, except maybe on VERY high-end multi-socket servers. In such systems, all legacy APIC IDs (< 256) may end up mapped to the first CPU (for example on EPYC 97xx), while the second (and 3rd, 4th, etc.) CPUs expose only APIC IDs > 256.

From my previous tests, 4 to 8 threads (per CPU) is already enough to saturate the memory controller of a modern CPU, so increasing the thread count beyond that brings little to no benefit in practice.

Therefore, keeping MAX_APS at 256 is perfectly fine for now. The main benefit of x2APIC support is to keep SMP working on newer CPUs (no longer supporting legacy APIC), rather than enabling support for > 256 threads.

system/smp.c
x86fr: Well I can't find a proper way to do that reliably. I have a commit ready, but I find the code awful, very obfuscated and it have an AP stack overflow issue probably due to the fallback on smp_my_cpu_num with required a big uint32_t array (AP stack on x86 is limited to 1024).

Let's write some snippet here :

On smp_init() :

```
    for (int i = 0; i < num_available_cpus; i++) {
        apic_id_to_cpu_num[cpu_num_to_apic_id[i]] = (uint16_t)i;
    }

    stack_base = (uintptr_t)_stacks + BSP_STACK_SIZE - LOCALS_SIZE;
    stack_max_top = stack_base + (uintptr_t)(num_available_cpus - 1) * AP_STACK_SIZE;
```
Later
```
int smp_my_cpu_num(void)
{
    if (num_available_cpus <= 1) return 0;

#if defined(__i386__) || defined(__x86_64__)
    uintptr_t sp;
#if defined(__i386__)
    __asm__ __volatile__("movl %%esp, %0" : "=r"(sp));
#else
    __asm__ __volatile__("movq %%rsp, %0" : "=r"(sp));
#endif
    if (sp <= stack_max_top && sp + AP_STACK_SIZE > stack_base) {
        uintptr_t adjusted = (sp - stack_base) + (AP_STACK_SIZE - 1);
        int cpu_num = (int)(adjusted / AP_STACK_SIZE);
        if (cpu_num >= 0 && cpu_num < num_available_cpus) {
            return cpu_num;
        }
    }
#endif

    return apic_id_to_cpu_num[my_apic_id()];
}
```

I looked at the asm code generated for all solution :

**Original O(1) access:**

```
int smp_my_cpu_num(void)
{
    return num_available_cpus > 1 ? apic_id_to_cpu_num[my_apic_id()] : 0;
 9d4:	48 8b 15 00 00 00 00 	mov    0x0(%rip),%rdx        # 9db <smp_my_cpu_num+0x7>
 9db:	31 c0                	xor    %eax,%eax
 9dd:	83 3a 01             	cmpl   $0x1,(%rdx)
 9e0:	7e 18                	jle    9fa <smp_my_cpu_num+0x26>
    return read32(&apic[APIC_REG_ID][0]) >> 24;
 9e2:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # 9e9 <smp_my_cpu_num+0x15>
__MEMRW_READ_FUNC(32)
 9e9:	8b 40 20             	mov    0x20(%rax),%eax
    return num_available_cpus > 1 ? apic_id_to_cpu_num[my_apic_id()] : 0;
 9ec:	48 8d 15 00 00 00 00 	lea    0x0(%rip),%rdx        # 9f3 <smp_my_cpu_num+0x1f>
    return read32(&apic[APIC_REG_ID][0]) >> 24;
 9f3:	c1 e8 18             	shr    $0x18,%eax
    return num_available_cpus > 1 ? apic_id_to_cpu_num[my_apic_id()] : 0;
 9f6:	0f b6 04 02          	movzbl (%rdx,%rax,1),%eax
}
 9fa:	c3                   	retq 
```

**The current O(n) search :**
 
```
 int smp_my_cpu_num(void)
{
    if (num_available_cpus <= 1) return 0;
 ab9:	48 8b 35 00 00 00 00 	mov    0x0(%rip),%rsi        # ac0 <smp_my_cpu_num+0x7>
 ac0:	83 3e 01             	cmpl   $0x1,(%rsi)
 ac3:	7f 05                	jg     aca <smp_my_cpu_num+0x11>
 ac5:	45 31 c0             	xor    %r8d,%r8d
 ac8:	eb 3c                	jmp    b06 <smp_my_cpu_num+0x4d>
    if (apic_x2apic) {
 aca:	80 3d 00 00 00 00 00 	cmpb   $0x0,0x0(%rip)        # ad1 <smp_my_cpu_num+0x18>
 ad1:	74 09                	je     adc <smp_my_cpu_num+0x23>
        rdmsr(MSR_IA32_X2APIC_BASE + APIC_REG_ID, msrl, msrh);
 ad3:	b9 02 08 00 00       	mov    $0x802,%ecx
 ad8:	0f 32                	rdmsr  
        return (int)msrl;
 ada:	eb 0d                	jmp    ae9 <smp_my_cpu_num+0x30>
    return read32(&apic[APIC_REG_ID][0]) >> 24;
 adc:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # ae3 <smp_my_cpu_num+0x2a>
 ae3:	8b 40 20             	mov    0x20(%rax),%eax
 ae6:	c1 e8 18             	shr    $0x18,%eax

    int apic_id = my_apic_id();
    for (int i = 0; i < num_available_cpus; i++) {
 ae9:	8b 0e                	mov    (%rsi),%ecx
 aeb:	31 d2                	xor    %edx,%edx
        if ((int)cpu_num_to_apic_id[i] == apic_id) {
 aed:	48 8d 35 00 00 00 00 	lea    0x0(%rip),%rsi        # af4 <smp_my_cpu_num+0x3b>
 af4:	41 89 d0             	mov    %edx,%r8d
    for (int i = 0; i < num_available_cpus; i++) {
 af7:	39 d1                	cmp    %edx,%ecx
 af9:	7e ca                	jle    ac5 <smp_my_cpu_num+0xc>
        if ((int)cpu_num_to_apic_id[i] == apic_id) {
 afb:	0f b6 3c 16          	movzbl (%rsi,%rdx,1),%edi
 aff:	48 ff c2             	inc    %rdx
 b02:	39 c7                	cmp    %eax,%edi
 b04:	75 ee                	jne    af4 <smp_my_cpu_num+0x3b>
    }
#endif

    return apic_id_to_cpu_num[my_apic_id()];
}
 b06:	44 89 c0             	mov    %r8d,%eax
 b09:	c3                   	retq   
 
```

And this solution (div is correctly shifted 10 as AP stack size is 1024) :

```
int smp_my_cpu_num(void)
{
    if (num_available_cpus <= 1) {
 b2c:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # b33 <smp_my_cpu_num+0x7>
        return 0;
 b33:	45 31 c0             	xor    %r8d,%r8d
    if (num_available_cpus <= 1) {
 b36:	8b 10                	mov    (%rax),%edx
 b38:	83 fa 01             	cmp    $0x1,%edx
 b3b:	7e 64                	jle    ba1 <smp_my_cpu_num+0x75>
#if defined(__i386__) || defined(__x86_64__)
    uintptr_t sp;
#if defined(__i386__)
    __asm__ __volatile__("movl %%esp, %0" : "=r"(sp));
#else
    __asm__ __volatile__("movq %%rsp, %0" : "=r"(sp));
 b3d:	48 89 e0             	mov    %rsp,%rax
#endif
    if (sp <= stack_max_top && sp + AP_STACK_SIZE > stack_base) {
 b40:	48 39 05 00 00 00 00 	cmp    %rax,0x0(%rip)        # b47 <smp_my_cpu_num+0x1b>
 b47:	72 2b                	jb     b74 <smp_my_cpu_num+0x48>
 b49:	48 8b 0d 00 00 00 00 	mov    0x0(%rip),%rcx        # b50 <smp_my_cpu_num+0x24>
 b50:	48 8d b0 00 04 00 00 	lea    0x400(%rax),%rsi
 b57:	48 39 ce             	cmp    %rcx,%rsi
 b5a:	76 18                	jbe    b74 <smp_my_cpu_num+0x48>
        uintptr_t adjusted = (sp - stack_base) + (AP_STACK_SIZE - 1);
 b5c:	48 29 c8             	sub    %rcx,%rax
 b5f:	48 05 ff 03 00 00    	add    $0x3ff,%rax
        int cpu_num = (int)(adjusted / AP_STACK_SIZE);
 b65:	48 c1 e8 0a          	shr    $0xa,%rax
 b69:	41 89 c0             	mov    %eax,%r8d
        if (cpu_num >= 0 && cpu_num < num_available_cpus) {
 b6c:	39 c2                	cmp    %eax,%edx
 b6e:	7e 04                	jle    b74 <smp_my_cpu_num+0x48>
 b70:	85 c0                	test   %eax,%eax
 b72:	79 2d                	jns    ba1 <smp_my_cpu_num+0x75>
    if (apic_x2apic) {
 b74:	80 3d 00 00 00 00 00 	cmpb   $0x0,0x0(%rip)        # b7b <smp_my_cpu_num+0x4f>
 b7b:	74 09                	je     b86 <smp_my_cpu_num+0x5a>
        rdmsr(MSR_IA32_X2APIC_BASE + APIC_REG_ID, msrl, msrh);
 b7d:	b9 02 08 00 00       	mov    $0x802,%ecx
 b82:	0f 32                	rdmsr  
        return (int)msrl;
 b84:	eb 0d                	jmp    b93 <smp_my_cpu_num+0x67>
    return read32(&apic[APIC_REG_ID][0]) >> 24;
 b86:	48 8b 05 00 00 00 00 	mov    0x0(%rip),%rax        # b8d <smp_my_cpu_num+0x61>
 b8d:	8b 40 20             	mov    0x20(%rax),%eax
 b90:	c1 e8 18             	shr    $0x18,%eax
            return cpu_num;
        }
    }
#endif

    return apic_id_to_cpu_num[my_apic_id()];
 b93:	48 98                	cltq   
 b95:	48 8d 15 00 00 00 00 	lea    0x0(%rip),%rdx        # b9c <smp_my_cpu_num+0x70>
 b9c:	44 0f b6 04 02       	movzbl (%rdx,%rax,1),%r8d
}
 ba1:	44 89 c0             	mov    %r8d,%eax
 ba4:	c3                   	retq  
```
IMO, on older systems using APIC, most of the time spent running this code goes into the read caused by the MMIO access (even on a 128-core CPU as anythinglarger should really be using x2APIC). This read can cost 200 to 500+ cycles, which is probably more than the time needed to execute the loop over 128 entries.

I can try to measure the real-world impact this way :

1. For x2APIC, there is no doubt that the MSR read is faster than the MMIO, even with the O(n) search on 256 cores. I already measured it.
2. For APIC, I can evaluate the O(n) seach this way: I will use my Xeon Phi platform, disable x2APIC and limit core count to 256. That's the absolute worst case scenario : VERY slow individual x86 core (~Pentium 4 level) with VERY high core count.

If the difference is barely noticeable, I would recommend to keep the O(n) seach.

x86fr: I just ran the test with 192T (for some reasons, I can't disable x2APIC at 256T). Test done within the 16GB of LLC to maximize the impact of core waiting at each barrier. Removed bit fade tests.

Original code with O(1) access needs 02:38.09 to complete tests 1-9 at half iteration 
This current PR with O(n) search needs 02:37.94 to complete the same pass

That's within the margin of error.

x86fr:
If there are no further objections, I’ll merge this. Having it on nightly will make beta testing much easier. We can still add optimizations in a follow-up PR before the release if needed.

martinwhitaker:
Go for it.

One thing to consider as a future optimisation is to try to pick a small number of cores from each proximity group, rather than just using the first 256.

debrouxl:
Yeah, let's go for this one, unless you want to temporarily add a mode where x2APIC usage can be disabled without rebuilding.

I plan on testing on EMR-SP in DL320 & DL380 Gen11 this week.

x86fr:
> Go for it.
> 
> One thing to consider as a future optimisation is to try to pick a small number of cores from each proximity group, rather than just using the first 256.

Just for the record, here is how all EPYC x2APIC IDs are ordered. It's quite simple and consistent across gens.

IMG_6125

debrouxl:
On a HPE DL380 Gen11 (2U, 2S) with BIOS U54 v2.72 from 2025/11/27, equipped with a single Xeon Gold 6530 and 4 x 64 GB DDR5 RAM, which loads mt86plus from RHEL 9 GRUB:
* the memtest86+ 8.00 release binary reproducibly triggers an immediate spontaneous reboot, after which the computer reaches the second boot screen only after three successive reboots during the first boot screen, one of which performs a full memory training;
* my own build of 450961c81e29d78697068529530d3b9e7deb88d2 , last commit before this PR was merged, runs in UP mode reliably enough but obviously slowly, I rebooted the computer after 6h run time;
* my own build of 4fa58699a95dba3a8cc9e40ffa55fb4e42012b92 , current HEAD with x2APIC support, attempts to use SMP 64T mode as it should; however, it reproducibly hangs after a small, random amount of time after starting test 3, in the seconds to dozens of seconds range. ESC isn't taken into account, forced power down ensues.

The spontaneous reboot with memtest86+ 8.00 also occurs on the closely related DL320 Gen11 (1U, 1S) with a Xeon Gold 6548N 32C/64T and 256 GB total RAM (4 x 64 GB or 8 x 32 GB, I don't remember). Maybe the #572 fix for #571 helped on both computers.

x86fr:
Please open an issue as the PR is closed. Also, does it work in xAPIC mode? Or by running it directly without being chainloaded?

PR #587, opened 2026-01-24
Better Cache & DRAM Bandwidth measurement (+Enable SSE2 on x86_64)

x86fr:
- Moved memory speed measurement to a dedicated file
- Use SSE2 on x86_64 for better DRAM BW measurements
- Add a lot of safeguards to avoid issue with very large L3 cache
- Add a more precise way to target a specific cache
- Change the iteration number from fixed to time-constrained so overall test time is always same
- Added many configuration parameters for fine-tuning

-------------------

This ended up being much more painful than I expected. Finding the sweet spot to target the right cache level across CPU architectures spanning more than 25 years is tricky. Getting cache alignment (almost) right and preventing the compiler from "helpfully" changing the ASM inlining behavior is also challenging.

This code isn’t perfect (far from it), but I think it’s an improvement over the previous version. 

Any hints or ideas are VERY welcome.

### DIFF
--- a/app/x86/interrupt.c
+++ b/app/x86/interrupt.c
@@ -84,10 +84,9 @@ static const char codes[][13] = {
 //------------------------------------------------------------------------------
 
 #ifdef __x86_64__
+
 typedef uint64_t    reg_t;
-#else
-typedef uint32_t    reg_t;
-#endif
+typedef float __m128 __attribute__((__vector_size__ (16), __aligned__ (16)));
 
 struct trap_regs {
     reg_t   ds;
@@ -99,26 +98,47 @@ struct trap_regs {
     reg_t   dx;
     reg_t   di;
     reg_t   si;
-#ifdef __x86_64__
     reg_t   r8;
     reg_t   r9;
     reg_t   r10;
     reg_t   r11;
-#else
-    reg_t   reserved1;
-    reg_t   reserved2;
-    reg_t   sp;
-#endif
+    reg_t   r12;
+    __m128  xmm[16];
     reg_t   bp;
     reg_t   vect;
     reg_t   code;
     reg_t   ip;
     reg_t   cs;
     reg_t   flags;
-#ifdef __x86_64__
     reg_t   sp;
-#endif
 };
+
+#else
+
+typedef uint32_t    reg_t;
+
+struct trap_regs {
+    reg_t   ds;
+    reg_t   es;
+    reg_t   ss;
+    reg_t   ax;
+    reg_t   bx;
+    reg_t   cx;
+    reg_t   dx;
+    reg_t   di;
+    reg_t   si;
+    reg_t   reserved1;
+    reg_t   reserved2;
+    reg_t   sp;
+    reg_t   bp;
+    reg_t   vect;
+    reg_t   code;
+    reg_t   ip;
+    reg_t   cs;
+    reg_t   flags;
+};
+
+#endif
 
 //------------------------------------------------------------------------------
 // Public Functions
--- a/app/x86/interrupt.c
+++ b/app/x86/interrupt.c
@@ -25,6 +25,12 @@
 // Constants
 //------------------------------------------------------------------------------
 
+#define ISR_GP_REGS_ONLY __attribute__((target( \
+  "no-mmx," \
+  "no-sse,no-sse2,no-sse3,no-ssse3,no-sse4.1,no-sse4.2," \
+  "no-avx,no-avx2," \
+)))
+
 #define INT_DIVBY0     0
 #define INT_RSV        1
 #define INT_NMI        2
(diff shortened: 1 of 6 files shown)


x86fr:
Regarding SSE2, instead of pushing 256 useless bytes onto the stack to preserve the XMM registers in case of interrupt (which may cause overflow even if I I haven’t seen it in my tests), maybe we can just use somethings like `__attribute__((target("no-sse,no-sse2")))` around the function to prevent the compiler from messing with them?

debrouxl:
In d244cbaa50aeaa8f31a2e82ce5a9e779e49567db , which is the current incarnation of SIMD writes in `mov_inv_fixed`, I used `#pragma GCC target (...)` instead of `__attribute__((target(...)))`, but the goal and checked effect were indeed the same as yours: forcing the compiler to use the desired instruction sets.

debrouxl:
Unless you want the release where this is integrated to be the breaking point where we start giving up on 32-bit hardware, I'd argue that SIMD code and non-temporal writes should be used on 32-bit x86 as well when cpuid indicates that they're available, like I did on the `simd` branch.
That's trading compile-time `#ifdef __x86_64__` blocks with run-time `if` blocks based on actual CPU features.

x86fr:
> Unless you want the release where this is integrated to be the breaking point where we start giving up on 32-bit hardware, I'd argue that SIMD code and non-temporal writes should be used on 32-bit x86 as well when cpuid indicates that they're available, like I did on the `simd` branch. That's trading compile-time `#ifdef __x86_64__` blocks with run-time `if` blocks based on actual CPU features.

We’re not really giving up on 32-bit hardware now, but I don’t see a practical real-world case for supporting SSE2 specifically on 32-bit x86. Aside from the Pentium III and some early Atom CPUs (where a 128-bit move can take two cycles, so you don’t gain memory-throughput benefits anyway) pretty much every CPU that supports SSE2 also supports x86-64. At this point, I don’t think there’s much we can meaningfully add for 20+ years old platforms. I’d much rather spend the effort on a new AVX2 code path than on adding an SSE2 path for 32-bit.

(But ultimately, yes : there will probably come a point when we should drop 32-bit support entirely for newer version.)

debrouxl:
If SIMD memory bandwidth measurements, and eventually tests, are to remain limited to SSE2 onwards, then the usefulness of 32-bit x86 support for SIMD is reduced, indeed. We can do better at low implementation cost, though: the SSE support I implemented on the `simd` branch, using `movaps` R/W and `movntps` W, makes it possible to target the PIII and Athlon XP. In SSE mode, `mov_inv_fixed` is measurably faster on my Athlon XP, IIRC by one third.
The one family of SIMD instructions I implemented on the `simd` branch, but which certainly makes the least sense, is MMX. It didn't seem to be interesting on my computers. Also, it used to cause issues with Ubuntu 20.04's version of GCC; the ones in Ubuntu 22.04 or other modern distros are unaffected, though.

Which are the instructions AVX2 would bring over AVX ?

x86fr:
On older CPUs, the store port throughput of these SIMD units is limited and not really better than the int units for what we are trying to achieve (maximum stress on the memory controller for error detection). That said, if it can be shown that using SSE on a Pentium III leads to better or faster memory error detection, I am fine with adding it (but from a µarch point of view I don't see how that would be possible). Without a proven benefit, there is no point in adding code for CPUs that are more than 20 years old.

Even on more modern CPUs, I tried many SSE2 test implementations and still couldn't demonstrate any improvement in error-detection performance. In many cases, errors are detected at the same rate as with the current code. 😒 

About AVX, yeah, AVX1 is probably enough to get access to the 256-bit YMM registers, but AVX2 adds a few instructions that sound interesting (though they may finally not be) to achieve maximum throughput, like `vmovntdqa`.

(PS: If you need a platform with a partially failing DIMM for experiments, I can probably find one for you)

debrouxl:
I don't remember seeing a speedup for `movaps` over `rep movsl` on the PIII, but there was one on the Athlon XP. I can't get fresh measurements tonight. And since I don't have any platform with calibrated errors for experimentation purposes, I have no clue whether memory error detection is better or faster...

FWIW, non-temporal stores were only useful on tests 1 and 2 IME; all other tests were reproducibly slower on at least one of my computers when using NT stores, no matter the unrolling factor (well, up to 64 bytes of written data, which is one cacheline in most architectures, that is). IOW, I wonder how useful NT stores will prove for measuring maximum memory bandwidth.

AVX1 is definitely enough to access the YMM registers and memory using `vmovdq / vmovntdq`, that's what I did in the `simd` branch. Making `vmovntdqa` available for XMM registers but not YMM registers at AVX level was an, uh, _interesting_ move. Also, I'd find it _silly_ if Intel or AMD ever made a micro-architecture on which `vmovntdqa` is faster than `vmovntdq` used with an aligned operand, but who knows...

martinwhitaker:
I would guess that in most systems there are enough CPU cores to saturate the memory bandwidth with the existing test implementations, so it would only be rare cases that would benefit from using SIMD instructions in the tests. The cache and DRAM bandwidth tests are a special case, because the cache bandwidth is far greater and because they are run on a single core.

It would certainly be a good idea to prevent the interrupt handling code using the SIMD registers to save pushing them on the stack.

x86fr:
Done

x86fr:
I’m about to merge this PR so it can be tested alongside the other major changes. I’ve tested it on older architectures (Pentium 4, K8, Core 2, and early Core), and the bandwidth measurements appear consistent, often showing gains of around 50% versus the actual legacy code.

PR #591, opened 2026-02-03
boot/x86: Use 32-bit zero idiom for shorter encoding

Integral-Tech:
Replace `xorq %rax, %rax` with `xorl %eax, %eax` to achieve a more compact encoding. The 32-bit instruction is one byte smaller while yielding the same semantics due to automatic zero-extension in 64-bit mode.

### DIFF
--- a/boot/x86/startup64.S
+++ b/boot/x86/startup64.S
@@ -160,7 +160,7 @@ startup:
 
 	# Pick the correct stack.
 
-	xorq	%rax, %rax
+	xorl	%eax, %eax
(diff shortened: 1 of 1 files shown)


x86fr:
Using `xorl %eax,%eax `relies on implicit x86-64 zero-extension to clear rax. Sure it's encoded 1 byte shorter, but it reduces clarity on this critical-but-cold path, and doesn’t provide a demonstrated performance or size benefit at the function level. On a cold path, we prefer to encode the intent directly.

PR #595, opened 2026-03-02
Fix Secureboot to allow shim loading

x86fr:
This is a WIP to allow Memtest86+ to be loaded by shim, and ultimately allow running on Secure Boot enabled PC.

I fixed the sbat formatting and removed the alignment that prevented the section to be parsed correctly by shim. 

Now Shim correctly validate the file signature and is happy with the sbat format, but for some reasons, it can't load the image correctly. Just after shim ends its verification code and try to load the image, my test computer just ... shut down. This is highly unexpected. A freeze, error or crash would be more common, but it just shut down. The signed binary correctly runs on SB enabled PC when loaded directly without shim.

image

I've tried a lot of stuff, but I can't understand what's the issue. I use Shim 15.8 source from Debian, so something could be wrong or hardcoded. The fact that it tries to load a "0xB8000" image seems wrong as it's much bigger than the size of Memtest86+ (but very close to the size of the original hardcoded grubx64.efi).

Any help welcome as I'm really not familiar with shim.

### DIFF
--- a/boot/sbat.csv
+++ b/boot/sbat.csv
@@ -1,2 +1,2 @@
 sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
-memtest86+,1,Memtest86+,6.0,https://github.com/memtest86plus
+memtest86+,1,Memtest86+,mt86plus,8.0,https://github.com/memtest86plus
--- a/boot/x86/startup64.S
+++ b/boot/x86/startup64.S
@@ -132,7 +132,21 @@ startup64:
 	.globl	efi_handover
 efi_handover:
 	andq	$~0xf, %rsp
-	call	efi_setup
+#	movq	%rdi, %r8		# save EFI image handle
+
+	# Clear .bss tail when VirtualSize > RawSize.
+#	xorq	%rax, %rax
+#	leaq	_bss(%rip), %rdi
+#	leaq	_end(%rip), %rcx
+#	subq	%rdi, %rcx
+#0:	cmpq	$0, %rcx
+#	je	1f
+#	movq	%rax, (%rdi)
+#	addq	$8, %rdi
+#	subq	$8, %rcx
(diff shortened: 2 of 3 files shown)


x86fr:
@vathpela > If you stop by, let us know if you think of anything. Your help would really be great for us to move forward. Tks!

@martinwhitaker @debrouxl > I've attached below a signed memtest86+ binary, the shim binary I used for the tests (you must rename memtest.efi to grubx64.efi to use it) and the certificate to be enrolled to be able to boot with SB enabled. 

Also poking @tenzap 

[Sbtest.zip](https://github.com/user-attachments/files/25697694/Sbtest.zip)

x86fr:
Good news: after studying shim’s code for several hours, I finally figured out how to make it work. I compiled the latest shim from source, added the vendor certificate, and successfully signed Memtest86+. Shim is now chainloading `mt86plus.efi` as expected with Secure Boot enabled. Of course, you still need to enroll the (MOK.esl) key into `db`.

This is a major step forward. My next step is to look into the submission process at the [`rhboot/shim-review` ](https://github.com/rhboot/shim-review) repository to get a shim signed. There are still a few important points to verify, such as the NX requirements, and I also need to set up the cryptographic infrastructure required to sign Memtest86+ builds. It appears that a FIPS 140-2 compliant security key is required, and those are quite expensive; even a basic YubiKey seems to cost more than 1000€. I’ll handle that as well.

If anyone is able to test the attached build, that would be very helpful.

[mt86plus_shim_signed.zip](https://github.com/user-attachments/files/26194936/mt86plus_shim_signed.zip)

debrouxl:
Good to read about that :)
I had started brainstorming for a minor piece of the submission process in #370 .
A FIPS 140-2 (-3 ?) compliant HSM, really ? Understandable, but pretty expensive for a project with unpaid maintainers. Still, I can participate to that cost somewhat.

PR #607, opened 2026-04-09
Add support for USB Mass Storage Devices

x86fr:
- Add detection for USB Mass Storage Devices (Flash Drive)
- Supports EHCI (USB 2) & xHCI (USB 3) controllers
- Supports MBR & GPT partitions
- Supports FAT16 & FAT32 filesystems
- Add basic report generation and saving to USB (using F1 -> F6)

This is big one I’ve been working on for several weeks. 

It’s far from finished, still contains many bugs, and will require extensive testing. There are so many quirks needed for certain controllers, implementations, hubs, and specific devices that it is almost impossible to support the thousands of different USB MSD devices that exist, but it already supports many of them.

I’ll have much fewer resources available for beta testing, since our production floor is not using it yet.


app/reports.c
```diff
@@ -0,0 +1,403 @@
...
+    va_end(args);
+    return pos;
+}
+
+static int format_results(char *buf, int bufsize)
+{
+    buf_end = buf + bufsize - 1;
+    char *pos = buf;
+
+    pos = buf_printf(pos, "Memtest86+ v%s Report\r\n", MT_VERSION);
```
debrouxl, line 168: MT_VERSION is a constant string literal, so let's just use C preprocessor string literal concatenation here:
`pos = buf_printf(pos, "Memtest86+ v" MT_VERSION " Report\r\n");`

(and if both GCC and Clang from Ubuntu 22.04 LTS support raw string literals in C mode, IMO, the string in the next line should be concatenated to this one)

app/reports.c
```diff
@@ -0,0 +1,403 @@
...
+            pos = buf_append_char(pos, ' ');
+        }
+        pos = buf_printf(pos, "]  Errors: %i\r\n", test_list[i].errors);
+    }
+    pos = buf_printf(pos, "\r\n");
+
+    // Test summary.
+    pos = buf_printf(pos, "Passes completed: %i\r\n", pass_num);
+    pos = buf_printf(pos, "Errors: %u\r\n", (uintptr_t)error_count);
+    pos = buf_printf(pos, "ECC Errors: %u\r\n", (uintptr_t)error_count_cecc);
```
debrouxl, line 282: Just a double `\r\n` at the end of this string ?

app/reports.c
```diff
@@ -0,0 +1,403 @@
...
+        usleep(2000 * MILLISEC);
+        goto cleanup;
+    }
+
+    char *results_buf = (char *)res_buf_addr;
+    int results_len = format_results(results_buf, RESULTS_BUF_SIZE);
+
+    // Build display filename with dot: "MT86P_NN.TXT"
+    char display_name[16];
+    memcpy(display_name, filename, 6);
```
debrouxl, line 378: `memcpy(dest,src,len)`, so `memcpy(display_name, filename, 8);` and remove the next two lines ?

app/reports.c
```diff
@@ -0,0 +1,403 @@
...
+
+#define MILLISEC    1000
+
+//------------------------------------------------------------------------------
+// Private Functions
+//------------------------------------------------------------------------------
+
+// RTC (CMOS) register reading. Ports 0x70/0x71 BCD format.
+
+static uint8_t rtc_read(uint8_t reg)
```
debrouxl, line 47: This is ISA-conditional, right ?

system/ehci.c
```diff
@@ -634,58 +707,69 @@ bool ehci_probe(uintptr_t base_addr, usb_hcd_t *hcd)
...
+            disable_ehci_port(op_regs, port_idx);
+        }
     }
 
-    print_usb_info(" Found %i low/full speed device%s, %i high speed device%s, %i keyboard%s",
-                   num_ls_devices, num_ls_devices != 1 ? "s" : "",
-                   num_hs_devices, num_hs_devices != 1 ? "s" : "",
-                   num_keyboards,  num_keyboards  != 1 ? "s" : "");
+    if (usb_mass_storage_found) {
+        print_usb_info(" Found %i low/full speed device%s, %i high speed device%s, %i keyboard%s, 1 USB drive",
```
debrouxl, line 723: In the current state of the code, the following would work:
```c++
print_usb_info(" Found %i low/full speed device%s, %i high speed device%s, %i keyboard%s%s,
                       num_ls_devices, num_ls_devices != 1 ? "s" : "",
                       num_hs_devices, num_hs_devices != 1 ? "s" : "",
                       num_keyboards,  num_keyboards  != 1 ? "s" : "",
                       usb_mass_storage_found ? ", 1 USB drive" : "");```
Likewise in a hunk from system/xhci.c .

File: system/fat32.h [hunk 6]
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * FAT16/32 filesystem context.
+ */
+typedef struct {
+    usb_msd_t   *msd;
+    uint32_t    partition_lba;      // LBA offset of the partition (0 if unpartitioned)
+    uint32_t    bytes_per_sector;
+    uint8_t     sectors_per_cluster;
+    uint16_t    reserved_sectors;
```
debrouxl, line 28: Move one line up to achieve better struct packing.

system/usbhcd.c
```diff
@@ -88,12 +93,18 @@ static int num_hcd = 0;
...
+
 //------------------------------------------------------------------------------
 // Public Variables
 //------------------------------------------------------------------------------
 
 usb_init_options_t usb_init_options = USB_DEFAULT_INIT;
 
+bool usb_mass_storage_found = false;
+
+char usb_msd_name[64] = "";
```
debrouxl, line 106: `= { 0 };` might be clearer, but no biggie.

system/usbmsd.c
```diff
@@ -0,0 +1,174 @@
...
+#define CSW_SIGNATURE       0x53425355
+
+#define CBW_FLAG_DATA_IN    0x80
+#define CBW_FLAG_DATA_OUT   0x00
+
+// SCSI command opcodes.
+#define SCSI_TEST_UNIT_READY    0x00
+#define SCSI_REQUEST_SENSE      0x03
+#define SCSI_INQUIRY            0x12
+#define SCSI_READ_CAPACITY_10   0x25
```
debrouxl, line 26: We probably want to add support for `SCSI_*_16` from the get go, multiple of my (larger) USB MSDs want it.

system/usbmsd.c
```diff
@@ -0,0 +1,174 @@
...
+
+static bool msd_bot_command(usb_msd_t *msd, const uint8_t *cdb, int cdb_len,
+                            void *data, uint32_t data_len, bool data_in)
+{
+    const usb_hcd_t *hcd = msd->hcd;
+
+    // Build Command Block Wrapper.
+    usb_cbw_t cbw;
+    memset(&cbw, 0, sizeof(cbw));
+    cbw.signature            = CBW_SIGNATURE;
```
debrouxl, line 65: memset(..., 0, ...) followed immediately by initialization of the fields, in order at that, is wasteful. Better initialize the fields and zero-initialize the portion of `cbw.cb` beyond `cdb_len`.

system/usbmsd.c
```diff
@@ -0,0 +1,174 @@
...
+    if (msd->block_size == 0) return false;
+
+    return true;
+}
+
+bool msd_read_sectors(usb_msd_t *msd, uint32_t lba, uint32_t count, void *buffer)
+{
+    uint8_t cdb[10];
+    memset(cdb, 0, 10);
+    cdb[0] = SCSI_READ_10;
```
debrouxl, line 150: memset followed by initialization, here and in the next function.

system/xhci.c
```diff
@@ -787,7 +802,10 @@ static bool assign_address(const usb_hcd_t *hcd, const usb_hub_t *hub, int port_
                 return false;
             }
         }
-        ep_context->max_packet_size = device->max_packet_size;
+        // For SS, bMaxPacketSize0 is an exponent (2^N bytes); expand it here.
```
debrouxl, line 805: This hunk, and several others in this file, look like fixes we want to apply ahead of this patch.

debrouxl:
That would fulfill one of the major feature requests, great :)

We definitely want to fuzz the partition scanning and FAT handling code.

system/spd.c
```diff
@@ -21,6 +21,7 @@
 
 ram_info_t ram = { 0, 0, 0, 0, 0, 0, "N/A"};
 ram_slot_info_t ram_slot_info[MAX_SPD_SLOT];
+spd_info spd_slot_cache[MAX_SPD_SLOT];
```
debrouxl, line 24: At the moment, nothing ever writes to this variable.

x86fr:
PR is really not ready for review yet. Code is only done at the "feature" level. It will be refined a bit later

x86fr:
At this early stage, the most needed is as much tests as possible on various hardware. Please report USB Drive Model, Motherboard model and status (OK or KO). 👍

system/usbmsd.c
x86fr: I wouldn't expect this to be needed for any USB Flash Drive with <2TB capacity, but ... added!

system/fat32.c
```diff
@@ -0,0 +1,549 @@
...
+
+// MBR partition type IDs.
+#define MBR_TYPE_FAT16_SMALL    0x04
+#define MBR_TYPE_FAT16          0x06
+#define MBR_TYPE_NTFS_EXFAT     0x07
+#define MBR_TYPE_FAT32          0x0B
+#define MBR_TYPE_FAT32_LBA      0x0C
+#define MBR_TYPE_FAT16_LBA      0x0E
+#define MBR_TYPE_GPT_PROTECTIVE 0xEE
+
```
debrouxl, line 28: Do we want to store data on the FAT filesystems in ESPs (0xEF, C12A7328-F81F-11D2-BA4B-00A0C93EC93B ) ? While it's not their main intended purpose, they're (ab)usable for storing data.
Not all ESPs are supposed to be writable, that said: for instance, those from isohybrid images.

system/fat32.c
```diff
@@ -0,0 +1,549 @@
...
+// Try to parse the BPB from the sector already in buf.
+// Returns true if it looks like a valid FAT16/32 BPB.
+static bool parse_fat_bpb(fat32_fs_t *fs, const uint8_t *buf)
+{
+    // Validate boot signature.
+    if (buf[510] != 0x55 || buf[511] != 0xAA) return false;
+
+    // Parse common BPB fields.
+    uint16_t tmp16;
+    memcpy(&tmp16, buf + 11, 2);
```
debrouxl, line 235: For reading 2 or even 4 bytes, we should read individual bytes and shift the result, through a `read_le16()` / `read_le32()` type inline function.
`memcpy()` is an endianness liability in general, and often a size + speed efficiency wart for such small reads, especially if the compiler decides to just make an unaligned read on x86.

system/fat32.c
martinwhitaker: > Do we want to store data on the FAT filesystems in ESPs

I would say no - let's not risk a bug in our code making a user's machine unbootable.

debrouxl: Good point.

x86fr: Clearly no. The next addition could be exFAT support or reading a config file from the USB key, but it's significantly more complex and the current code really needs some testing before adding new features.

system/usbmsd.c
debrouxl: Good.
Right, USB MSDs smaller than 2 TB shouldn't _require_ it. I recently saw a >2 TB HDD used for saving data wipe reports, and memtest86+ would be in the same boat as data wipers.

app/reports.c
x86fr: I find it much easier to read on separate lines, but the report format will need to be polished later (maybe even generating html)

system/fat32.c
x86fr: We're reading 2–4 bytes from an aligned sector buffer, so the access never crosses a cache line and there's no unaligned penalty possible here. For such small (constant-size) operands, memcpy() is optimized out to a single mov by GCC (not an actual call).

x86fr:
Ready for review. Very few failures on actual hardware BUT testing still limited to 2/3 betatesters. We need a bigger pool.

app/display.c
```diff
@@ -68,7 +68,7 @@ static int test_ticks = 0;      // current value (ticks_per_test is final value)
 static int pass_bar_length = 0; // currently displayed length
 static int test_bar_length = 0; // currently displayed length
 
-static uint64_t run_start_time = 0; // TSC time stamp
+uint64_t run_start_time = 0; // TSC time stamp
```
martinwhitaker: Move this from Private Variables to Variables.

app/config.c
```diff
@@ -982,7 +983,23 @@ void config_menu(bool initial)
             prints(POP_R+11, POP_LI, "<F10> Exit menu");
         } else {
             prints(POP_R+7,  POP_LI, "<F5>  Skip current test");
-            prints(POP_R+8 , POP_LI, "<F10> Exit menu");
+            if (usb_mass_storage_found) {
+                if (usb_msd_name[0]) {
+                    // Truncate the drive name so the line stays inside the popup.
```
martinwhitaker, line 1028: You don't seem to use `usb_msd_name` anywhere else after the initial USB port scan, so you could simply force a maximum display length by `usb_msd_name[19] = '\0';`

system/ehci.c
```diff
@@ -634,58 +730,65 @@ bool ehci_probe(uintptr_t base_addr, usb_hcd_t *hcd)
 
         num_hs_devices++;
 
-        // Look for keyboards attached directly or indirectly to this port.
+        // Look for keyboards and USB drives attached directly or indirectly to this port.
         if (find_attached_usb_keyboards(hcd, &root_hub, 1 + port_idx, USB_SPEED_HIGH, num_hs_devices,
```
martinwhitaker, line 825: I would normally say you should rename the function to indicate it does more than just search for keyboards. But I think ultimately we should be scanning for USB drives at the time the user requests the report (so the user doesn't have to have plugged the drive in before starting testing), so let's leave it as it is for now.

system/usbhcd.c
```diff
@@ -750,6 +803,8 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
     }
     usb_device_desc_t *device = (usb_device_desc_t *)hcd->ws->data_buffer;
     bool is_hub = (device->class == USB_CLASS_HUB);
+    uint8_t product_str_index = device->product_str;
+    uint8_t num_configs = device->num_configs;
```
martinwhitaker, line 954: Why create copies here? `device` is not pointing to a volatile struct, so the compiler will optimise for repeated reads.

system/usbhcd.c
```diff
@@ -786,6 +841,80 @@ bool find_attached_usb_keyboards(const usb_hcd_t *hcd, const usb_hub_t *hub, int
         get_keyboard_info_from_descriptors(hcd->ws->data_buffer, hcd->ws->data_length,
                                            keyboards, max_keyboards, &new_num_keyboards);
         if (new_num_keyboards == old_num_keyboards) {
+            // No keyboard interfaces found, check for mass storage.
+            if (!usb_mass_storage_found && hcd->methods->configure_bulk_ep != NULL) {
```
martinwhitaker: This function is getting overly long. I suggest moving the body of this if statement into a separate function.

system/usbhcd.c
```diff
@@ -291,6 +303,10 @@ static bool configure_keyboard(const usb_hcd_t *hcd, const usb_ep_t *ep0, int in
     return true;
 }
 
+static bool get_msd_info_from_descriptors(const uint8_t *desc_buffer, int desc_length,
+                                          usb_ep_t *ep_in, usb_ep_t *ep_out,
+                                          uint8_t *alt_setting);
```
martinwhitaker: Move the actual function here, or better still, immediately after `get_keyboard_info_from_descriptors`.

system/usbhcd.c
```diff
@@ -736,6 +757,38 @@ bool assign_usb_address(const usb_hcd_t *hcd, const usb_hub_t *hub, int port_num
     return true;
 }
 
+static void fetch_usb_string(const usb_hcd_t *hcd, const usb_ep_t *ep0, uint8_t str_index, char *out, int out_size)
```
martinwhitaker: This should be in the Private Functions section.

martinwhitaker (commented):
Some initial comments. I've only begun to review the xhci.c changes; my initial feeling is that it would be cleaner to add a second event ring for the bulk transfers, but it's a while since I looked at that code.

app/display.c
x86fr: Yep

app/config.c
x86fr: Yep! Simple & efficient.

system/usbhcd.c
x86fr: If I understand correctly, `device` points to the shared data buffer and the call to `get_configuration_descriptors()` below loads the configuration descriptors into that same buffer, so device descriptor is destroyed at that point. Because we need `product_str_index` and `num_config`s later, we save them here (same as for `is_hub`).

system/usbhcd.c
x86fr: Yes, and that could also help for running the USB MSD detection at run time.

system/usbhcd.c
x86fr: Moved. I also moved check_for_usb_msd() in the Private Functions section for the same reason. Both are now placed after configure_keyboard(), so all the enumeration helpers stay grouped together.

system/ehci.c
x86fr: Implemented the USB scan when user want to save report with [2134cb1](https://github.com/memtest86plus/memtest86plus/pull/607/commits/2134cb13a465d60331361cbc2effc2e8190592ba)

x86fr:
No major changes or new features are planned at this stage, only bug fixes and optimizations.

I would really like to merge 4 four enhancement commits soon (ideally next saturday), so the beta testers can start working on them all at once as soon as possible. They have availability in July and August, so this seems like the right time to move forward. I can use a rebase instead of a squash, which should make it easier to complete the review even after the merge.

x86fr:
This PR is planned to be merged 4th on 2026-07-11

PR #616, opened 2026-05-22
Add temperature quirks for AMD EPYCs

### LINKED ISSUE #411: Wrong temperature on Epyc Milan (Zen3)
Hi,

Thanks for this tool, I have (unfortunately) to use it a lot for debugging some hardware issues and memtest86+ is of a great help in that. The version tested is the v6.10 (debian12)
I changed recently CPU going from Epyc 7262 to 7203p, everything remained the same and the temperature changed from ~50° to >100°. Since both processors have the same number of cores (8) and TDP (155W), I double checked the heatsink and thermal paste.

When running under linux, I realized the module `k10temp` was giving temperatures ~50° while the `zenpower` module (with some patches to support zen3) tells ~100° plus a warning in the dmesg that the formula for zen2 was used. I suspect that memtest86+ is also using the wrong formula to get the temperature. 

According to https://github.com/torvalds/linux/blob/master/drivers/hwmon/k10temp.c#L451
The offset for the different generation of processor is different.

This is a graph of the recording of the temperature of the system with one system monitor to the other and it confirms the difference is apparently in the  offset (not in the scale or anything else)
sensors_temp-week

I can try to make a patch but it is probably better done if you can handle this offset. I am ready to compile & test for you. Thanks

Jerome

x86fr:
Closes #411

### DIFF
--- a/system/hwquirks.c
+++ b/system/hwquirks.c
@@ -175,8 +175,10 @@ static const amd_tctl_offset_t amd_tctl_offset_table[] = {
     { 0x8, 0x0, -27.0f, "AMD Ryzen Threadripper 19" },  // Whitehaven (1900X/1920X/1950X)
     { 0x8, 0x0, -27.0f, "AMD Ryzen Threadripper 29" },  // Colfax (29x0X/29x0WX)
     { 0x8, 0x0, -27.0f, "AMD EPYC 7"                },  // Naples (Family 17h, Model 01h)
-    // Rome/Milan/Genoa/Turin EPYC parts report Tdie directly using the
-    // bit-19 (T_OFFSET_PRESENT) path; no additional table entry needed.
+    //{ 0xA, 0xA, -49.0f, "AMD EPYC 8"                },  // Siena (Maybe needed)
+    //{ 0xA, 0x1, -49.0f, "AMD EPYC 9"                },  // Genoa (Maybe needed)
+    { 0xB, 0x0, -49.0f, "AMD EPYC 9"                },  // Turin (Family 19h, Model 11h)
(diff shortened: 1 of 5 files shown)


system/hwquirks.c
```diff
@@ -158,6 +159,46 @@ static void amd_k8_revfg_temp(void)
     cpu_temp_offset = 21.0f;
 }
 
+// AMD Zen Tctl -> Tdie offset table.
+typedef struct {
+    uint8_t      ext_family;
+    uint8_t      ext_model;
+    const char  *brand_prefix;
+    float        tctl_offset;
```
debrouxl, line 166: Let's swap `tctl_offset` and `brand_prefix` to take advantage of what would otherwise be superfluous padding on x86-64.

system/hwquirks.c
```diff
@@ -158,6 +159,46 @@ static void amd_k8_revfg_temp(void)
     cpu_temp_offset = 21.0f;
 }
 
+// AMD Zen Tctl -> Tdie offset table.
+typedef struct {
+    uint8_t      ext_family;
+    uint8_t      ext_model;
+    const char  *brand_prefix;
```
debrouxl: Like previous occurrences of a `const char *` pointing to a string, which cause costly 24-byte relocations on x86-64 ELF, this one ought to be `const char [max len for CPUID brand + model name]`.

system/hwquirks.c
x86fr: Yep

ThomasWaldmann:
Thanks @x86fr!

Will test it next time I get an EPYC machine in my hands (might take a while though). :-)

PR #626, opened 2026-07-05
Add support for aarch64 (ARM64) architecture

x86fr:
Damn, that was quite a challenge, but here it is: Memtest86+ running on an ARM laptop powered by a Snapdragon X SoC!

image

Fully working on a 13" Surface, with UEFI, SMP, USB, serial, and more. Only temp is missing, but it's really painful to add.

Open for review.

The plan is to merge this together with USB MSD support and the upcoming SSE2/AVX2 tests asap, then run a single beta-testing session for all three.


system/smp.c
```diff
@@ -591,17 +645,40 @@ static bool find_cpus_in_madt(void)
                 found_cpus++;
             }
         }
+#elif defined(__aarch64__)
+        if (entry_header->type == MADT_GICC) {
+            // GICC entries are 76 or 80 bytes long, depending on the ACPI
+            // revision. Both variants have the MPIDR at the same offset.
+            if (entry_header->length < 76) {
```
debrouxl, line 658: `if (entry_header->length != 76 && entry_header->length != 80)` ?

boot/aarch64/startup.S
```diff
@@ -0,0 +1,371 @@
...
+	add	x10, x10, x11
+	adrp	x11, _stacks
+	add	x11, x11, :lo12:_stacks
+	add	x11, x11, x10
+	mov	sp, x11
+
+	bl	reloc
+
+	// If this copy of the program doesnt own the current page tables,
+	// rebuild them in this copys memory (safe: we are still running on
```
debrouxl: "copy's", here and in the next line.

boot/aarch64/startup.S
```diff
@@ -0,0 +1,371 @@
...
+	adrp	x0, _start
+	add	x0, x0, :lo12:_start
+	adrp	x9, tables_for_base
+	ldr	x1, [x9, :lo12:tables_for_base]
+	cmp	x0, x1
+	b.eq	1f
+	str	x0, [x9, :lo12:tables_for_base]
+	bl	paging_init
+1:
+	// Every CPU (re)loads TTBR0 with *this copy* page tables
```
debrouxl: Also "'s" at lines 230 and 239 ?

boot/aarch64/startup.S
```diff
@@ -0,0 +1,371 @@
...
+	add	x0, x0, :lo12:vectors
+	msr	vbar_el1, x0
+	isb
+
+	// Release the startup mutex
+	adrp	x9, startup_mutex
+	add	x9, x9, :lo12:startup_mutex
+	stlr	wzr, [x9]
+
+	// Run MEmtest86+.
```
debrouxl: Typo.

system/smp.c
x86fr: The comment is wrong here. GICC entries are USUALLY 76 or 80 bytes, but newer ACPI (6.x?) revisions also allow other size.

This is inspired by Linux ARM acpi code at https://github.com/torvalds/linux/blob/7404ce51637231382873d0b55edabc2f3b841a9d/arch/arm64/include/asm/acpi.h#L40

boot/boot.h
```diff
@@ -19,7 +19,7 @@
 #define	MAX_APS		511		/* Maximum number of active APs */
 
 #define BSP_STACK_SIZE	16384		/* Stack size for the BSP */
-#ifdef __loongarch_lp64
+#if defined(__loongarch_lp64) || defined(__aarch64__)
 #define AP_STACK_SIZE	2048		/* Stack size for each AP */
```
debrouxl: While we're at it, does APX bring anything in the way of mandatory changes, e.g. exception frame format, which would raise stack usage, even if we don't use APX features because they aren't useful to a code base like ours ? And would RV64 be different from LA64 and AArch64 ?

Just trying to assess whether we'll arguably end up simply using AP_STACK_SIZE = 2048 on all supported 64-bit architectures in the near future anyway, in which case the preprocessor condition could be simpler.

x86fr: No mandatory change for APX. Exception frame from hardware stay exactly same, new registers R16-R31 only used if compiled with `-mapxf`. We will never do that because of BC with old CPUs
About RV64, I don't know, but I think it will probably land on ARM64/LA64 territory. 

About x86, it much better to stay on the proven 1024 bytes to avoid wasting up to 511 KB (MAX_APS), especially on old platforms. But the condition could be inverted (x86 = 1024, all others = 2048)

x86fr: I was wrong here. We can't waste 511KB on old platform because old platform do not have 512 APs.
More importantly, I just spend hours to debug an issue that was finally an AP stack corruption on x86_64.
The call chain (common_err -> scroll -> printf -> screen rendering + plus USB keyboard polling) needs more than 1 KB. I'm working on a fix, and maybe add some safe safeguard in the process.

x86fr:
This PR is planned to be merged 5th on 2026-07-11

system/smp.c
debrouxl: Alright, but IMO, there needs to be an upper bound check, either a comparison of `entry_header->length` against a reasonable length value, or an equivalent of the check found in the Linux ARM ACPI code you linked, e.g. `(uintptr_t)entry_header + entry_header->length > (uintptr_t)mpc_table_end` :)

x86fr: Yes, that would requires a (very badly) corrupted table, but adding a safeguard can't hurt. Fixed.

x86fr:
How cool is that? :)

image

debrouxl:
That's cool indeed, good work :)

AFAIK, ACPI remains relatively unpopular on ARM platforms, but the number of AArch64 (and RV64) platfoms optionally or permanently equipped with ACPI-speaking firmware is growing.

x86fr:
Yes, and device tree is painful to implement, mainly because its lack of standardization across vendors

app/aarch64/interrupt.c
```diff
@@ -0,0 +1,132 @@
...
+    uint64_t    far;
+    uint64_t    vec;
+    uint64_t    pad;
+};
+
+//------------------------------------------------------------------------------
+// Constants
+//------------------------------------------------------------------------------
+
+static const char *vector_name[] = {
```
debrouxl: Well, not only `const char *` + string data + relocation data is costly space-wise, like on x86... but also, unlike x86, there's significant redundancy in the string data.

I'd suggest something like
```diff
diff a/app/aarch64/interrupt.c b/app/aarch64/interrupt.c
--- a/app/aarch64/interrupt.c
+++ b/app/aarch64/interrupt.c
@@ -39,23 +39,18 @@ struct trap_regs {
 // Constants
 //------------------------------------------------------------------------------
 
-static const char *vector_name[] = {
-    "Synchronous EL1t",
-    "IRQ EL1t",
-    "FIQ EL1t",
-    "SError EL1t",
-    "Synchronous EL1h",
-    "IRQ EL1h",
-    "FIQ EL1h",
-    "SError EL1h",
-    "Synchronous EL0/64",
-    "IRQ EL0/64",
-    "FIQ EL0/64",
-    "SError EL0/64",
-    "Synchronous EL0/32",
-    "IRQ EL0/32",
-    "FIQ EL0/32",
-    "SError EL0/32"
+static const char vector_name_1[][12] = {
+    "Synchronous",
+    "IRQ",
+    "FIQ",
+    "SError",
+};
+
+static const char vector_name_2[][7] = {
+    "EL1t",
+    "EL1h",
+    "EL0/64",
+    "EL0/32",
 };
 
 //------------------------------------------------------------------------------
@@ -98,8 +93,9 @@ void interrupt(struct trap_regs *trap_regs)
     clear_message_area();
 
     display_pinned_message(0, 0, "Unexpected exception on CPU %i", smp_my_cpu_num());
-    display_pinned_message(2, 0, "Type: %s (%s)", vector_name[trap_regs->vec & 0xF],
-                                                  exception_class_name(trap_regs->esr));
+    display_pinned_message(2, 0, "Type: %s %s (%s)", vector_name_1[trap_regs->vec & 0x3],
+                                                     vector_name_2[(trap_regs->vec >> 2) & 0x3],
+                                                     exception_class_name(trap_regs->esr));
     display_pinned_message(3, 0, "  PC: %016x", (uintptr_t)trap_regs->elr);
     display_pinned_message(4, 0, " FAR: %016x", (uintptr_t)trap_regs->far);
     display_pinned_message(5, 0, " ESR: %016x", (uintptr_t)trap_regs->esr);
```
which I have now compile-tested, and saves hundreds of bytes. However, even an array of 19-character-wide C strings would strongly beat `const char *` + string data + relocation data.

system/aarch64/cpuid.c
```diff
@@ -0,0 +1,241 @@
...
+
+#include "registers.h"
+
+//------------------------------------------------------------------------------
+// Types
+//------------------------------------------------------------------------------
+
+typedef struct {
+    uint16_t    part_num;
+    const char  *name;
```
debrouxl: Here as well, `const char *` + string data + relocation data is a significant pessimization.
```diff
diff --git a/system/aarch64/cpuid.c b/system/aarch64/cpuid.c
--- a/system/aarch64/cpuid.c
+++ b/system/aarch64/cpuid.c
@@ -18,7 +18,7 @@
 
 typedef struct {
     uint16_t    part_num;
-    const char  *name;
+    const char  name[14];
 } cpu_part_t;
 
 //------------------------------------------------------------------------------
@@ -61,7 +61,7 @@ static const cpu_part_t arm_parts[] = {
     { 0xD87, "Cortex-A725"  },
     { 0xD89, "Cortex-A720AE"},
     { 0xD8E, "Neoverse N3"  },
-    { 0, NULL }
+    { 0, "" }
 };
 
 static const cpu_part_t qcom_parts[] = {
@@ -73,7 +73,7 @@ static const cpu_part_t qcom_parts[] = {
     { 0x804, "Kryo 4xx Au"  },
     { 0x805, "Kryo 4xx Ag"  },
     { 0xC00, "Falkor"       },
-    { 0, NULL }
+    { 0, "" }
 };
 
 static const cpu_part_t apple_parts[] = {
@@ -89,12 +89,12 @@ static const cpu_part_t apple_parts[] = {
     { 0x035, "M2 Aval Pro"  },
     { 0x038, "M2 Bliz Max"  },
     { 0x039, "M2 Aval Max"  },
-    { 0, NULL }
+    { 0, "" }
 };
 
 static const cpu_part_t microsoft_parts[] = {
     { 0xD49, "Cobalt 100"   },
-    { 0, NULL }
+    { 0, "" }
 };
 
 static const cpu_part_t cavium_parts[] = {
@@ -108,7 +108,7 @@ static const cpu_part_t cavium_parts[] = {
     { 0x0B4, "OcteonTx2 95N"},
     { 0x0B5, "OcteonTx2 95M"},
     { 0x0B6, "OcteonTx2 95O"},
-    { 0, NULL }
+    { 0, "" }
 };
 
 //------------------------------------------------------------------------------
@@ -123,7 +123,7 @@ cpuid_info_t cpuid_info;
 
 static const char *lookup_part(const cpu_part_t *parts, uint16_t part_num)
 {
-    for (int i = 0; parts[i].name != NULL; i++) {
+    for (int i = 0; parts[i].name[0] != 0; i++) {
         if (parts[i].part_num == part_num) {
             return parts[i].name;
         }
```
saves another kilobyte or so. In the other patch I suggested tonight, my compiler produced warnings when the width was too small by one, so attempts to use too long string data should be caught at compiler time.

app/aarch64/interrupt.c
x86fr: A 512-byte saving in exchange for a less readable array. Well, it's a cold path but relocation cost is real, let's do it

system/aarch64/cpuid.c
x86fr: yes. it will be caught except with GCC < 15 and a 14-char string (in that case the null terminator will be silently dropped). Let's increase the array by a bit and add a comment

x86fr:
I’m exhausted, I’ll merge this tomorrow and let the betatesters start testing it on Monday. I still need to implement LASX tomorow, but this super-widely scoped PR really needs to be merged. All the latest large PRs interact with one another, which makes debugging quite painful.

PR #628, opened 2026-07-06
Expand and modernise the test suite for V9

x86fr:
-   Add a new SIMD moving-inversions test (# 4). Per-lane xorshift sequences with non-temporal stores, AVX2/SSE2/scalar tier picked at runtime. Replaces the old mov_inv_random test.
-   Add a bus stress test (# 9). DDR5+ oriented: per-CPU NT-store streams with fine-grained R/W turnarounds and duty-cycled idle gaps toprovoke link-SI / PMIC droop and similar  faults.
-   Rework the Bit-Fade test (# 10). Four solid/random fill–fade–check rounds with parallel fill/check and a fresh reproducible pattern each pass. The current 0s/1s only covers 15% of the faults according to the studies.
-   Add a Rowhammer test (# 11). Blacksmith-style many-sided REFRESH-synced hammering with DRAMA-lite row-stride discovery and a  blind-stride fallback. Default disabled.
-   Reorder/rebalance the test suite to cut time-to-first-fault. Mov_inv_walk was oversampled previously.

This is by far the biggest change for V9 and must be committed first
First tests on a limited number of PCs showed a strong reduction of TTFF. Much longer tests needed.

Ready to discuss any point.


tests/x86/vec_prsg_avx2.c
```diff
@@ -0,0 +1,137 @@
...
+        "incq     %[done]                  \n\t"
+        "cmpq     %[n], %[done]            \n\t"
+        "jb       0b                       \n\t"
+        "2:                                \n\t"
+        "vmovdqu  %%ymm0, (%[st])          \n\t"
+        "vzeroupper                        \n\t"
+        "sfence                            \n\t"
+        : [q] "+r" (q), [done] "+r" (done)
+        : [st] "r" (st), [n] "r" (nblocks), [splat] "r" ((int)splat)
+        : "rax", "xmm0", "xmm1", "xmm2", "xmm3", "memory", "cc"
```
debrouxl, line 142: In these AVX2-based functions, shouldn't the clobbers apply to ymm* instead ?

tests/mov_inv_rng.c
```diff
@@ -0,0 +1,479 @@
...
+
+    for (int k = shift; k < TESTWORD_WIDTH; k += shift) {
+        t >>= shift;
+        x ^= t;
+    }
+    return x;
+}
+
+// Returns the previous word in the pseudo-random sequence generated by prsg().
+static inline testword_t prsg_inverse(testword_t state)
```
debrouxl: With the loops inside the subroutines, even if the iteration step is shift instead of 1, that looks like a fair bit more computationally expensive than `prsg()`. Out of curiosity, have you benchmarked the forward and backwards versions separately ?

debrouxl:
Another interesting, large-scale PR :)

Reducing TTFF is a good thing.

tests/x86/vec_prsg_avx2.c
x86fr: They're the same hardware registers in GCC/Clang clobber lists, so they can be named xmm or ymm, only cosmetic. A clobber of xmmX invalidates the entire register (up to 256 bits). As long as it's not compiled with -mavx2, the compiler can't hold ymm uppers so that should be fine, BUT that deserve a comment and a safeguard.

tests/mov_inv_rng.c
x86fr: Backward is clearly slower (~4x), but it's still memory-bound (by a large margin) in SMT. Working on a faster loop, but it won't be MUCH faster.

x86fr:
Early tests show VERY interesting improvements in error detection for the new bus stress test. 
If this is confirmed later, a reordering of the tests may be a good idea (test 9 -> 3)

x86fr:
This PR is planned to be merged 1st on 2026-07-11

PR #629, opened 2026-07-06
Display memory info from DMI when SPD is unavailable

x86fr:
Add a fallback display for systems where SPD is unavailable (LPDDR5/soldered memory, unsupported I2C controllers, servers): populated  SMBIOS/DMI Type 17 records are collected, identical modules grouped, and size/type/configured speed/manufacturer/part number shown in the usual SPD area.

Manufacturer is resolved from the JEDEC module manufacturer ID (SMBIOS 3.2+) via the shared JEP-106 table, with a filtered fallback to  the free-form DMI string. That requires the SMBIOS v3 entry point support cherry-picked from x86fr/aarch64

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -303,7 +303,11 @@ void display_cpu_topology(void)
 void post_display_init(void)
 {
     print_smbios_startup_info();
-    print_spd_startup_info();
+
+    if (print_spd_startup_info() == 0) {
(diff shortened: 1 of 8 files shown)


x86fr:
This PR is planned to be merged 3rd on 2026-07-11

PR #630, opened 2026-07-07
Hardened AP Stack Management

x86fr:
Fix SMP freeze when a worker CPU reports memory errors (AP stack overflow)

On SMP runs with a defective DIMM, the test froze completely (spinner and  keyboard dead) after a variable number of reported errors, on any test. Issue is present in v8.10.

Each CPU's thread-local storage (the halt-barrier waiting flags) occupies the top LOCALS_SIZE bytes of its own stack slot, directly below the bottom of the next CPU's stack. The error reporting path (error mutex + scroll + printf + screen rendering + keyboard polling) needs more than the 1024 bytes an AP stack provided, so a reporting AP overran its slot and overwrote its neighbour's barrier flags with stack garbage. The releasing CPU then either skipped the victim's wakeup NMI, or the victim read a garbage flag and treated its wakeup as stale: one CPU slept forever in HLT and every other CPU deadlocked on the next barrier. Single-CPU runs were immune because the BSP has a 16KB stack.

- Increase AP_STACK_SIZE from 1024 to 2048 bytes (matching LoongArch/Aarch64)
- Add a canary at the bottom of each CPU's stack slot, checked once per second by the master, so a future overrun is reported on screen instead of deadlocking silently ***--- Performance penalty under evaluation ---***
-  Only poll the keyboard from the master CPU in the error reporting and scroll-lock paths; this also removes the deepest stack frames (USB keyboard polling) from the workers' error path
- Use spin_trylock() on the error mutex in the unexpected-interrupt handler, so a trap taken while the mutex is held reports instead of hanging

### DIFF
--- a/app/display.c
+++ b/app/display.c
@@ -6,6 +6,7 @@
 #include <stdint.h>
 
 #include "cpuid.h"
+#include "cpulocal.h"
 #include "cpuinfo.h"
 #include "hwctrl.h"
 #include "i2c_x86.h"
@@ -15,6 +16,7 @@
 #include "serial.h"
 #include "pmem.h"
 #include "smbios.h"
+#include "smp.h"
 #include "spd.h"
 #include "temperature.h"
 #include "tsc.h"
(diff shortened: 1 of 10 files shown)


x86fr:
Build that on top of the v9tests branch and messed up...

PR #631, opened 2026-07-07
Hardened AP Stack Management

x86fr:
Fix SMP freeze when a worker CPU reports memory errors (AP stack overflow)

On SMP runs with a defective DIMM, the test froze completely (spinner and  keyboard dead) after a variable number of reported errors, on any test. Issue is present in v8.10.

Each CPU's thread-local storage (the halt-barrier waiting flags) occupies the top LOCALS_SIZE bytes of its own stack slot, directly below the bottom of the next CPU's stack. The error reporting path (error mutex + scroll + printf + screen rendering + keyboard polling) needs more than the 1024 bytes an AP stack provided, so a reporting AP overran its slot and overwrote its neighbour's barrier flags with stack garbage. The releasing CPU then either skipped the victim's wakeup NMI, or the victim read a garbage flag and treated its wakeup as stale: one CPU slept forever in HLT and every other CPU deadlocked on the next barrier. Single-CPU runs were immune because the BSP has a 16KB stack.

- Increase AP_STACK_SIZE from 1024 to 2048 bytes (matching LoongArch/Aarch64)
- Add a canary at the bottom of each CPU's stack slot, checked once per second by the master, so a future overrun is reported on screen instead of deadlocking silently ***--- Performance penalty under evaluation ---*** **Done: about 300 ms for a 1 hour pass on 256 core**
-  Only poll the keyboard from the master CPU in the error reporting and scroll-lock paths; this also removes the deepest stack frames (USB keyboard polling) from the workers' error path
- Use spin_trylock() on the error mutex in the unexpected-interrupt handler, so a trap taken while the mutex is held reports instead of hanging

### DIFF
--- a/system/cpulocal.c
+++ b/system/cpulocal.c
@@ -2,16 +2,42 @@
 // Copyright (C) 2022 Martin Whitaker.
 
 #include <stdbool.h>
+#include <stdint.h>
 
 #include "boot.h"
+#include "smp.h"
 
 #include "cpulocal.h"
+
+//------------------------------------------------------------------------------
+// Constants
+//------------------------------------------------------------------------------
+
+// An arbitrary value that is unlikely to appear in a stack frame.
+#define STACK_CANARY    UINT32_C(0x446D6153)
 
 //------------------------------------------------------------------------------
 // Variables
 //------------------------------------------------------------------------------
 
 int local_bytes_used = 0;
+
+// Whether each stack canary is currently valid. The stack area is not
+// preserved across program relocations, hence the arm/disarm cycle.
+static bool canary_armed[MAX_CPUS];
+
+//------------------------------------------------------------------------------
+// Private Functions
+//------------------------------------------------------------------------------
+
+static uint32_t *stack_canary_addr(int cpu_num)
+{
+    uintptr_t slot_bottom = (uintptr_t)_stacks;
+    if (cpu_num > 0) {
+        slot_bottom += BSP_STACK_SIZE + (uintptr_t)(cpu_num - 1) * AP_STACK_SIZE;
+    }
+    return (uint32_t *)slot_bottom;
+}
 
 //------------------------------------------------------------------------------
 // Public Functions
@@ -23,4 +49,30 @@ int allocate_local_flag(void)
         return -1;
     }
     return local_bytes_used += sizeof(bool);
+}
+
+void stack_canary_arm(int cpu_num)
+{
+    if (cpu_num < 0 || cpu_num >= MAX_CPUS) {
+        return;
+    }
+    *stack_canary_addr(cpu_num) = STACK_CANARY;
+    canary_armed[cpu_num] = true;
+}
+
+void stack_canary_disarm_all(void)
+{
+    for (int i = 0; i < MAX_CPUS; i++) {
+        canary_armed[i] = false;
+    }
+}
+
+int stack_canary_check(void)
+{
+    for (int i = 0; i < num_available_cpus; i++) {
+        if (canary_armed[i] && *stack_canary_addr(i) != STACK_CANARY) {
+            return i;
+        }
+    }
+    return -1;
 }
--- a/app/display.c
+++ b/app/display.c
@@ -6,6 +6,7 @@
 #include <stdint.h>
 
 #include "cpuid.h"
+#include "cpulocal.h"
 #include "cpuinfo.h"
 #include "hwctrl.h"
 #include "i2c_x86.h"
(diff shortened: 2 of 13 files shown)


system/cpulocal.c
```diff
@@ -2,17 +2,43 @@
...
+
 //------------------------------------------------------------------------------
 // Variables
 //------------------------------------------------------------------------------
 
 int local_bytes_used = 0;
 
+// Whether each stack canary is currently valid. The stack area is not
+// preserved across program relocations, hence the arm/disarm cycle.
+static bool canary_armed[MAX_CPUS];
```
debrouxl, line 27: We ought to use a bit field here, so as to cut the size of that variable by a factor of 8.
We'll arguably raise MAX_CPUs in the relatively near future: by the end of the year, AMD is supposed to release 256-core, SMT=2 processors, which will yield 1024 threads in a dual-socket configuration; now that x2APIC is supported and is known to work on a number of real-world platforms, we can actually work with such platforms.

x86fr: There's two arguments against a bit array. 

First, we're speaking about 448 (512 - 64) bytes out of 1MB stacks (BSP is 16 KB + 511 * 2KB), that's 0.05%. Negligible gain. 

Then, it won't work as-is. This array is accessed simultaneously by all APs. With a bool array, each CPU writes its own byte on a distinct memory location. Easy, no race condition, no synchronization needed. A bit-map would be much more complex because setting a bit becomes a read-modify-write on a word shared by 32/64 APs. The bit set/cleared by one AP can be silently dropped if another AP write another bit from the same word at the same time. 

There is technical solution (like atomic OR, which would require per-arch code while this file is arch-neutral), but that's way too much code and complexity than 448 bytes justify.

x86fr:
An issue has been detected with BIOS/CSM, where the increased stack size seems to wipe various structures like RSDP and SMBIOS when the program is running in low mem. Some safeguards must be added to set_load_addr() and to force early relocation if not enough memory is available. Fix incoming.

x86fr:
Fixed

x86fr:
This PR is planned to be merged 2nd on 2026-07-11

debrouxl:
This remains useful to help detect AP stack overflows, even if they're much less likely with `AP_STACK_SIZE = 2048` than with `AP_STACK_SIZE = 1024`. 

TIL that there's a way to lower the large stack usage of varargs functions on x86-64: requesting that the compiler uses only GPRs in the calling convention of these functions, skipping the save/restore for XMM registers. This can be done:
* on a per-file basis, through compiling e.g. lib/print.c, app/display.c, app/reports.c with the `-mgeneral-regs-only` compiler flag;
* on a per-function basis, through the addition of `__attribute__((target("general-regs-only")))` onto the likes of `printf`, `vprintf` and `do_trace`.

x86fr:
Make sense to get more stack headroom, but that will prevent adding some float format (eg: `%f`) in `printf()` in the future. 
If we're doing that, I prefer the per-file way.

debrouxl:
On x86, we could force going back to x87 instructions (we underwent the x87 -> SSE2 change at eac97205ff3b42233013fd2149aa6b212d3f6069 shortly before 8.10), probably keep the current SSE2 computations for floats and doubles while passing them into x87 registers just for printing purposes, or using more questionable constructs (casts to and from uint32_t and uint64_t through a union come to my mind...), but point taken, this is something we need to keep in mind.

It's probably not 9.00 material if the beta-testing process' progress is significantly complete, but yeah, I think that we need to do it, in order to reduce the stack usage of every varargs function in a given call chain by 128 bytes... Indeed, 2048-byte AP stacks will cause (2 MB - 2 KB) of AP stacks and therefore a much greater consumption of memory within the first 4 MB (and potential compatibility issues) when we raise MAX_APS to 1023, MAX_CPUS and MAX_APIC_IDS to 1024 - which is something I'd argue we need to do do rather soon, considering that the availability of 2S-capable, 256C/512T Zen 6 EPYC processor models has been announced for Q4 2026.

The per-file way requires compiling the following files with `-mgeneral-regs-only`:
* `app/display.c`, because of `do_trace()`;
* `app/reports.c`, because of `buf_printf()`, `buf_vprintf()` and `slog()`;
* `lib/print.c`, because of `printf()` and `vprintf()`;
* `system/usbhcd.c`, because of `print_usb_info()`.

There seem to be sporadic occurrences of SSE2 instructions within those files, but it's not like they're hot paths, and besides, spilling and restoring the XMM registers has a (marginal) cost, too.

x86fr:
Agreed (and yes, after v9 as a cleanup commit). 

About HCC CPUs, I don't think there's much benefit in running an extremely (256+) large number of threads in parallel. A better approach would probably be topology-aware scheduling (using NUMA), to make sure each CCD (or core groups) is exercised without necessarily testing every logical CPU at once.

PR #634, opened 2026-07-12
la64: Add LASX Test (instead of the scalar fallback path)

x86fr:
Add a proper 256-bit LASX test

### DIFF
--- a/boot/loongarch/startup64.S
+++ b/boot/loongarch/startup64.S
@@ -336,8 +336,13 @@ Jmp:
     li.w       $t1, 0x1004
     iocsrwr.w  $t0, $t1
 
-    # Turn on FPE to enable FPU.
-    li.w    $t0, (0x1 << 0)
+    # Turn on the FPU, plus LSX/LASX when implemented
+    # (CPUCFG2 bits 6/7 map directly onto EUEN.SXE/ASXE at bits 1/2).
+    li.w    $t0, 0x2
+    cpucfg  $t0, $t0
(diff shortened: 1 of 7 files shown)


x86fr:
@kilaterlee / @MarsDoge Can you please check if the LASX implementation looks good to you?

MarsDoge (commented):
@x86fr Excellent, the 256-bit LASX look solid. Do you plan to also introduce 128-bit as a fallback for hardware that supports LSX but not LASX (e.g. 2K3000/3B6000M)? Or is the scalar fallback sufficient for those targets?

@ziyao233 You mentioned in https://github.com/memtest86plus/memtest86plus/pull/578 that you were working on an LSX-accelerated memspeed version and showed it measured better — curious to hear your thoughts on this vec_prsg LASX path as well.

kilaterlee:
> @x86fr Excellent, the 256-bit LASX look solid. Do you plan to also introduce 128-bit as a fallback for hardware that supports LSX but not LASX (e.g. 2K3000/3B6000M)? Or is the scalar fallback sufficient for those targets?
> 

I think just adding 256 in this PR is fine, 128-bit can be added later.

x86fr:
Since I don't have the hardware to test a LSX-only kernel on the real CPU (LSX without LASX), I’d rather merge it as is. As @kilaterlee said, the 128-bit LSX kernel (and membench implementation) can be added later with another PR.

kilaterlee (approved):
LGTM, thank you @x86fr !

PR #638, opened 2026-07-17
implement serial console logging

classabbyamp:
I use memtest86+ for some automated testing of RAM, and the best way I've found to exfiltrate the results has been via logging progress, errors, and results to a serial console. The existing `console` option was not easily machine-parsable, because it did lots of moving around the console and redrawing things.

To enable serial logging, I used a command-line option similar to `console`, and because `log` and `console` both use the serial console, they are made mutually exclusive.

image

(screenshot from an older version, but I've tested it on the latest HEAD of `main`)

example logs:
```
Memtest86+ v8.00.unknown.x64
start pass 0 test 0
pass 0 (0%) test 0 (1%)
pass 0 (0%) test 0 (3%)
start pass 0 test 2
pass 0 (0%) test 2 (15%)
pass 0 (0%) test 2 (30%)
pass 0 (0%) test 2 (40%)
pass 0 (0%) test 2 (50%)
[...]
pass 0 (99%) test 10 (83%)
pass 0 (99%) test 10 (84%)
pass 0 (99%) test 10 (85%)
pass 0 (99%) test 10 (85%)
pass 0 (99%) test 10 (86%)
pass 0 (99%) test 10 (87%)
pass 0 (99%) test 10 (88%)
pass 0 (99%) test 10 (88%)
pass 0 (99%) test 10 (89%)
pass 0 (99%) test 10 (94%)
pass 0 (99%) test 10 (97%)
Test Result: PASS
Done, rebooting
```

I've been testing this on Dell servers with iDRAC serial over LAN for several months without issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,11 @@ recognised:
       * mmio16 = 16-bit MMIO
       * mmio32 = 32-bit MMIO
     * and *y* is the MMIO address in hex. with `0x` prefix (eg: 0xFEDC9000)
+  * log=ttyS*x*,*y*
+  * log=*x*,*y*
+    * log status/progress to serial/tty/UART console.
+    * takes the same parameters as console.
+    * mutually exclusive with console.
   * newline
     * modifies the console to print a newline after every change to the frame buffer
       * useful in logging over serial where an escape or newline is needed
--- a/app/config.c
+++ b/app/config.c
@@ -111,6 +111,7 @@ bool            dark_mode          = false;
 power_save_t    power_save         = POWER_SAVE_HIGH;
 
 bool            enable_tty         = false;
+bool            enable_tty_log     = false;
 uintptr_t       tty_address        = 0x3F8;             // Legacy IO or MMIO Address accepted
 int             tty_baud_rate      = 115200;
 int             tty_update_period  = 2;                 // Update TTY every 2 seconds (default)
@@ -128,8 +129,6 @@ bool            err_banner_redraw  = false;             // Redraw banner on new 
 
 static void parse_serial_params(const char *params)
 {
-    enable_tty = true;
-
     // No parameters passed (only "console"), use default
      if (params == NULL) {
         return;
@@ -297,6 +296,16 @@ static void parse_option(const char *option, const char *params)
     if (params == NULL) params = "";
 
     if (strncmp(option, "console", 8) == 0) {
+        enable_tty = true;
+        if (enable_tty_log) {
+            enable_tty_log = false;
+        }
+        parse_serial_params(params);
+    } else if (strncmp(option, "log", 4) == 0) {
+        enable_tty_log = true;
+        if (enable_tty) {
+            enable_tty = false;
+        }
         parse_serial_params(params);
     } else if (strncmp(option, "newline", 7) == 0) {
         tty_new_line = true;
--- a/app/config.h
+++ b/app/config.h
@@ -61,6 +61,7 @@ extern bool         enable_temp_ram;
 
 extern bool         enable_sm;
 extern bool         enable_tty;
+extern bool         enable_tty_log;
 extern bool         enable_bench;
 extern bool         enable_mch_read;
 extern bool         enable_ecc_polling;
--- a/app/display.c
+++ b/app/display.c
@@ -22,6 +22,7 @@
 #include "tsc.h"
 
 #include "barrier.h"
+#include "log.h"
 #include "spinlock.h"
 
 #include "config.h"
@@ -603,25 +604,25 @@ void do_tick(int my_cpu)
 
     pass_type_t pass_type = (pass_num == 0) ? FAST_PASS : FULL_PASS;
 
-    int pct = 0;
+    int test_pct = 0;
     if (ticks_per_test[pass_type][test_num] > 0) {
-        pct = 100 * test_ticks / ticks_per_test[pass_type][test_num];
-        if (pct > 100) {
-            pct = 100;
+        test_pct = 100 * test_ticks / ticks_per_test[pass_type][test_num];
+        if (test_pct > 100) {
+            test_pct = 100;
         }
     }
-    display_test_percentage(pct);
-    display_test_bar((BAR_LENGTH * pct) / 100);
+    display_test_percentage(test_pct);
+    display_test_bar((BAR_LENGTH * test_pct) / 100);
 
-    pct = 0;
+    int pass_pct = 0;
     if (ticks_per_pass[pass_type] > 0) {
-        pct = 100 * pass_ticks / ticks_per_pass[pass_type];
-        if (pct > 100) {
-            pct = 100;
+        pass_pct = 100 * pass_ticks / ticks_per_pass[pass_type];
+        if (pass_pct > 100) {
+            pass_pct = 100;
         }
     }
-    display_pass_percentage(pct);
-    display_pass_bar((BAR_LENGTH * pct) / 100);
+    display_pass_percentage(pass_pct);
+    display_pass_bar((BAR_LENGTH * pass_pct) / 100);
 
     bool update_spinner = true;
     if (clks_per_msec > 0) {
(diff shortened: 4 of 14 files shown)


debrouxl:
Thanks for your contribution :)
Serial output which can be easily parsed was an existing feature request. We'll see what the other maintainers have to say about it.

Implementation-wise, I can't say I'm fond of the partial code duplication between `lib/print.c` and `lib/log.c` brought by this PR, especially in `(log_)printk`. I do understand some of it, but I'd expect a bit of refactor to reduce the need for it ?

classabbyamp:
> Implementation-wise, I can't say I'm fond of the partial code duplication between lib/print.c and lib/log.c brought by this PR, especially in (log_)printk. I do understand some of it, but I'd expect a bit of refactor to reduce the need for it ?

that's understandable. I can look into this. Would something like moving the common parts of `log.c` and `print.c` functions into `format.c` be preferable?

x86fr:
First, let me say that I fully understand the need for this feature. Our SI factory uses it internally quite a lot, even if they plan to switch to a more modern USB report soon. Splitting the serial output between a “console” mode and a “log” mode is clearly a nice feature to have.

From an implementation point of view, however, I don't think the current code on this PR is the correct approach. It spreads the changes in many places, while almost everything could now be added to the new `report.c` file. This file can be extended to contain nearly all the code needed for a serial log output mode.

That said, the code itself is quite trivial and really not a big deal. I already have a similar change ready, using the new `report.c` file. This is the version we currently use in the factory.

I did not push it yet because the main point that needs discussion is not the code, but the format of the report.

In this PR, the serial log output is formatted for your specific needs. We formatted ours differently for our own needs, and probably everyone would do the same, with different requirements. That's the main caveat. Before adding this feature to the main branch, we should find a consensus about an output format that is useful for the majority of users.

It would be very interesting to understand exactly which information you need, how often the output needs to be updated, and any other requirements related to your test case.

Our current report looks like this. A timestamp is included on every line. We try to avoid flooding the serial port with constant update (our Digi PortServer hates it).

First, it sends a start beacon:

```text
[yyyy-dd-mm hh:ii:ss] [START] Memtest86+ 8.10.abcdef.x64
```

Then it sends a list of information similar to what is included in the USB report:

```text
[yyyy-dd-mm hh:ii:ss] [INFO] [TYPE] XXXXXXXXXXXXXXXXXXXXXX
```

`TYPE `being, for example: `SPD`, `SERIAL`, `CPU`, `IMC`, `RMW`, etc.

Then all tests are reported as they run:

```text
[yyyy-dd-mm hh:ii:ss] [TEST #1] BEGIN
[yyyy-dd-mm hh:ii:ss] [TEST #1] END
[yyyy-dd-mm hh:ii:ss] [TEST #2] BEGIN
...
```

When an error is detected, it outputs:

```text
[yyyy-dd-mm hh:ii:ss] [TEST #X] ERROR FOUND!
```

No detailed error information is included at this point.

At the end of the pass, it writes either:

```text
[yyyy-dd-mm hh:ii:ss] [PASS] NO ERROR FOUND
[yyyy-dd-mm hh:ii:ss] [FAIL] 123 ERRORS FOUND
```

When loop mode is selected, the reporting starts again from the beginning.

Would this format work for you? Do you need different information, more detailed data, or more frequent output?

classabbyamp:
What you describe is mostly what I need from this feature, but the detailed error information is something I need. The exact formatting is entirely changeable on my end :)

I'm not 100% sure, but I think the Dell and HP virtual serial ports (SOL/VSP via BMC SSH) might need more frequent output to ensure the connection stays alive. Maybe reporting every 5 or 10% of a test could be a compromise between spamming the serial target and keeping the connection alive?

debrouxl:
Maybe a full trace for every step is overly verbose, but a single trace at the beginning and end of a test seems too infrequent to me. 2-6 traces per minute, i.e. a 30", 20", 15", 12" or 10" interval, indicating the current memory range being tested, provides some interesting heartbeating information IMO. Not that memtest86+ remotely has a habit of hanging or crashing, otherwise we ought to get more reports about it, but we seem to have found several computer models where that happens in x2APIC mode, per #605 .

Several ideas about the parsing-oriented output format:
* I'd minimize the superfluous characters in the output, so I'd remove those [];
* heartbeat outputs should have an extensible format. I haven't given up on the idea of eventually testing multiple memory ranges in parallel on NUMA hardware, it's just too time-consuming and probably too hard for me to implement and test (without a good coding-oriented LLM, I guess, but using a LLM to contribute to a GPL'ed code base is a known licensing minefield and some projects avoid it in the first place) - and having to modify the format later for that purpose would suck for users;
* taking one step back: what about adopting one of the simpler formats defined by **RFC 3164** (BSD syslog) or **RFC 5424** (syslog), typically timestamp which needn't be too fine-grained + log level + SSV message ? Such a format remains readable by humans, and it would make memtest86+ serial output (*) directly ingested by a syslog consumer, like well-behaved Linux and ESXi computers, and unlike e.g. many switches and firewalls which tend to spew out weird model-dependent proprietary formats.

*: or network output if it is ever implemented, but good luck achieving significant real-world coverage with the wide variety of NICs and their firmware requirements, and without bloating memtest86+ immensely...

x86fr:
What level of detail do you need for error reporting? The issue is that errors can occur by the thousands and easily flood the serial port. In our custom implementation, we report only a single “ERROR FOUND” message for each test. Agree that this could not be enough, but reporting every error with all details (adr, expexted, got) is probably too much.

About update rate, you're probably right. A forced update every x seconds with the test/pass % progress may be the sweet spot. What about 10 secs?

classabbyamp:
the detail I have in my change to `error.c` is what I'm aiming for.

10 seconds per heartbeat should be fine

classabbyamp:
Feel free to close this if you want to PR/merge your own implementation. I threw this together so I could monitor things, so I'm perfectly happy to adjust my automation to a more polished version of the concept.

x86fr:
> to implement and test (without a good coding-oriented LLM, I guess, but using a LLM to contribute to a GPL'ed code base is a known licensing minefield and some projects avoid it in the first place)

Clearly off-topic, but I strongly disagree. I share recent Linus Torvalds' view on using LLMs for open-source development. They trained their models on vast amounts of pillaged open-source code and now we shouldn't use their expensive tools because of "licensing concerns"? No way.

> * taking one step back: what about adopting one of the simpler formats defined by **RFC 3164** (BSD syslog) or **RFC 5424** (syslog), typically timestamp which needn't be too fine-grained + log level + SSV message ? Such a format remains readable by humans, and it would make memtest86+ serial output (*) directly ingested by a syslog consumer, like well-behaved Linux and ESXi computers, and unlike e.g. many switches and firewalls which tend to spew out weird model-dependent proprietary formats.

Syslog is quite complex and not really human readable for many techs with basic skills. Also it's a log format while the serial output is more a report / progress report. We must find the best of both worlds: keeping it really easy to read for a human and also easy to parse. I use [] because it's trivial to parse with a very simple regex.

> *: or network output if it is ever implemented, but good luck achieving significant real-world coverage with the wide variety of NICs and their firmware requirements, and without bloating memtest86+ immensely...

Yes. Network support is very painful and would require a painful additional work for shim review (really we must keep it simple on this point).

x86fr:
> Feel free to close this if you want to PR/merge your own implementation. I threw this together so I could monitor things, so I'm perfectly happy to adjust my automation to a more polished version of the concept.

The only real mismatch between our factories seems to be the level of detail needed in the error logs. May I ask how you use these details?

classabbyamp:
I want to be able to look at the log and see things like roughly where in the RAM the error is, so I can replace that one stick, basically.

x86fr:
But you need a platform with known address-to-slot mapping to achieve that, which is very uncommon these days (especially on server platform where nearly all IMCs now use multiple layers of interleaving).

This said, we stop the test at the first error on our side, so we don't really care about verbose error logs... 

We really need feedbacks from more users of a serial logging feature.

x86fr:
merged to #640

PR #639, opened 2026-07-17
add an option to limit the number of passes

classabbyamp:
can be set on the commandline (`maxpasses=n`, where `n = 0` means run forever and `n > 0` means stop after `n` passes) or at runtime

This is a more flexible alternative to #343

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,9 @@ recognised:
       For example: `testlist=0,1` would only run tests 0 and 1.
   * ecc
     * Enable ECC polling
+  * maxpasses=*n*
+    * performs *n* passes of selected tests then reboots rather than testing endlessly
+    * if not specified or set to `0`, run tests endlessly
 
 ## Keyboard Selection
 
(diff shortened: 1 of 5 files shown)


x86fr:
Just curious here, but how do you use this in your automation process? Eg: how can you tell whether your test target hasn’t already silently restarted x times?

(PS: Make sure your repo is writable by using)

classabbyamp:
a couple ways: the serial logging from #638 will also show some telltale messages from the BMC if a reboot happens, and the device under test will always boot to PXE (that I control and track by serial number), so I know exactly when it reboots, and have a state machine managing everything.

x86fr:
While reviewing this PR, I couldn't find a single use case for this feature outside of the serial log functionality. 

A `maxpasses` option only makes sense with an automatic reboot at the end (otherwise there's no harm in letting the tests continue) and an automatic reboot is only useful when the results have been captured through the serial log feature. Even the serial console mode wouldn’t make sense with `maxpasses`.

If you agree with that, it seems I need to push the serial log PR first, then this one (or maybe merge both).

classabbyamp:
that makes sense, I think

x86fr:
OK. Still thinking about the log format. I will share a first proposal tomorrow.

x86fr:
@classabbyamp Can you please check #640?

x86fr:
merged to #640

PR #641, opened 2026-07-20
Add SPD Checksum/CRC check at startup

### LINKED ISSUE #532: SPD content verification / CRC check
### Discussed in https://github.com/memtest86plus/memtest86plus/discussions/531

<div type='discussions-op-text'>

<sup>Originally posted by **Kuntsaito** July 24, 2025</sup>
Because weird memory behavior isn't always caused by bad DRAM chips. I learned that the hard way today.

In my experience, most consumer motherboards completely ignore checksums, allow some bad data in the SPD, and if they see memory installed in the slots, they will try their best to make it work. But sometimes that leads to unexpected consequences.

I recently received a motherboard with all components installed, in a "motherboard turns on and turns off instantly, you buy it as non-working, no complaints or returns" condition. This board came with a set of "roughly identical" AMD **Entertainment** 🤬 Series RAM sticks. I installed them one by one into my computer to test them, and they were in perfect working order.

Then I forgot about them, and now I've taken them out again to install them into another system. And surprise, that computer doesn't work and beeps about bad RAM. Wow! Furthermore, I tried to boot the system with only one stick inserted - neither worked. While diagnosing the problem, I found that if I put another, fully working stick of memory alongside the problematic one, everything worked fine, and of course the memory passed MemTest.

While investigating the cause, I thought that maybe these AMD sticks had incorrect timings. And here's where the worst part begins: most known programs that can provide information about RAM behave the same way as motherboards: they either guess or silently suppress errors caused by bad SPD data if they try to decode it. And none of them will tell you that there is something terribly wrong there.
disaster

Even worse: the program specifically designed to configure SPD data refused to read the SPD contents from these sticks and returned a "floating point divided by zero" error. But with its help I was finally able to see what was in the SPD and what the problem was - just one flipped bit. Great!
th

But it's not only programs that start to fail when they don't get what they expected. One of the motherboards I tested this memory on successfully passed POST and hung on the BIOS splash screen. When I corrected the incorrect data in the SPD, the memory in this board started working without any problems.

P.S. If anyone wants to play around, here is the spd contents of the problematic memory.
[AMD R538G1601U2SL DDR3-0-BAD1.txt](https://github.com/user-attachments/files/21397902/AMD.R538G1601U2SL.DDR3-0-BAD1.txt)
[AMD R538G1601U2SL DDR3-0-BAD2.txt](https://github.com/user-attachments/files/21397903/AMD.R538G1601U2SL.DDR3-0-BAD2.txt)

In both cases, byte 0x0b is corrupted. How and why this happened will most likely remain a mystery.</div>

x86fr:
Verify the JEDEC SPD checksum of each module after parsing, and flag modules with corrupted SPD content, which can hang POST on some boards while every existing tool silently ignores the invalid checksum 

Closes #532

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,8 @@ recognised:
     * disables the big PASS/FAIL pop-up status display
   * nosm
     * disables SMBUS/SPD parsing, DMI decoding and memory benchmark
+  * nospdcrc
+    * disables SPD checksum/CRC verification
   * nomch
     * disables memory controller configuration polling
   * nopause
--- a/app/config.c
+++ b/app/config.c
@@ -99,6 +99,7 @@ bool            enable_temp_cpu    = true;
 bool            enable_temp_ram    = true;              // DDR5+ temperature polling
 
 bool            enable_sm          = true;
+bool            enable_spd_crc     = true;
 bool            enable_bench       = true;
 bool            enable_mch_read    = true;
 bool            enable_numa        = false;
@@ -348,6 +349,8 @@ static void parse_option(const char *option, const char *params)
         enable_sm = false;
     } else if (strncmp(option, "nosmp", 6) == 0) {
         smp_enabled = false;
+    } else if (strncmp(option, "nospdcrc", 9) == 0) {
+        enable_spd_crc = false;
     } else if (strncmp(option, "numa", 5) == 0) {
         enable_numa = true;
     } else if (strncmp(option, "nonuma", 7) == 0) {
--- a/app/config.h
+++ b/app/config.h
@@ -60,6 +60,7 @@ extern bool         enable_temp_cpu;
 extern bool         enable_temp_ram;
 
 extern bool         enable_sm;
+extern bool         enable_spd_crc;
 extern bool         enable_tty;
 extern bool         enable_bench;
 extern bool         enable_mch_read;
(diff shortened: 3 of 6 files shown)


x86fr:
From @debrouxl 
> I wonder whether we can avoid checking the spd memory type twice, or rather, whether doing so improves or worsens the code.

Yes. At some point, I think we will need to preserve the complete SPD contents for every DIMM across relocations (8 KB total max.). The best approach would be to read the data once and then reuse the buffer for all subsequent processing.

Unrelated, but I measured the time required to read the entire SPD array:

- DDR2 and previous (64 bytes): ~30-50 ms per DIMM. 
- DDR3/DDR4 (~128 bytes) : ~60 ms 
- DDR5 (512 bytes) : ~250/300 ms typical (worst case 500 ms)

The verify-twice-on-mismatch retry 2x these numbers only for a genuinely corrupted module.

If we need to speed things up, I really don't want to mess with the delay across the I2C commands The better option is to switch the ICH5 path to SMBus read-word transactions (instead of bytes), BUT touching anything on that very sensitive path is risky and I don't think the added 250 ms/DIMM worth it.

x86fr:
Today, a RMA center found the first real-world SPD "Bad Checksum!" message on a DIMM that fails to boot on many motherboard. After investigation, the SPD was actually flawed.

